PR #9, opened 2017-10-29
Basic styling

VaclavSir:
In this pull request I was trying to:

- Fix navigation colors on Android
- Set buttons' color
- Add some margins and paddings
- Find some way to share colors and styles between different components

# Before / After
## Android
image
## iPhone
image

### DIFF
--- a/app/screens/homepage/Homepage.js
+++ b/app/screens/homepage/Homepage.js
@@ -1,7 +1,7 @@
 // @flow
 
 import * as React from 'react';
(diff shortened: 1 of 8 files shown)


mrtnzlml:
Good, job! Thank you very much. Merged by 6f5ae58f55f868b9694b1dfda10ed1d6c6502067...

In the very near future, I will try to reduce the amount of green color because it doesn't look very modern. I like Airbnb style (it's not red - it's white with red highlights and it looks very modern).

PR #28, opened 2017-11-20
WebView for search result

jakubzaba:


mrtnzlml:
Integrated. Thank you very much.

PR #30, opened 2017-11-24
Fix missing key in Login component

VaclavSir:
Fixing this yellow warning:

screen shot 2017-11-24 at 22 05 46
screen shot 2017-11-24 at 22 05 51


mrtnzlml:
Thanks for the fix. It's very weird because it seems like my Expo is **not** running in development mode and therefore I cannot reproduce it. I don't know why... :-(

PR #47, opened 2017-12-01
Enforce 'react/jsx-no-bind' eslint rule

mrtnzlml:
Even though I know about this rule I am not able to use it without
eslint rule. So let's enforce best practices (I believe this is
best practice)...

See: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md

### DIFF
--- a/app/common/src/badge/VerificationBadge.js
+++ b/app/common/src/badge/VerificationBadge.js
@@ -7,6 +7,17 @@ type Props = {|
   verified: boolean,
 |};
 
+type TextOnLayout = {|
+  nativeEvent: {|
+    layout: {|
+      x: number,
+      y: number,
+      width: number,
+      height: number,
+    |},
+  |},
+|};
+
 /**
  * This component creates badge for verified/unverified statuses. It
  * automatically calculates the container width based on width of text
@@ -22,36 +33,47 @@ type Props = {|
  * | UNVERIFIED with very long text to make it clear |
  * `-------------------------------------------------`
  */
-export default function VerificationBadge({ verified }: Props) {
-  let styleSet = unverifiedStyle;
-  let badgeText = 'UNVERIFIED';
+export default class VerificationBadge extends React.Component<Props> {
+  viewContainerReference: React.Ref<typeof View> | null;
 
-  if (verified === true) {
-    styleSet = verifiedStyle;
-    badgeText = 'VERIFIED';
-  }
+  /**
+   * React will call the 'ref' callback with the DOM element when the
+   * component mounts, and call it with null when it unmounts.
+   * 'ref' callbacks are invoked before componentDidMount or
+   * componentDidUpdate lifecycle hooks.
+   */
+  storeContainerReference = (reference: React.Ref<typeof View> | null) =>
+    (this.viewContainerReference = reference);
 
-  return (
-    <View
-      style={styleSet.button}
-      ref={viewContainer => {
-        this.viewContainer = viewContainer;
-      }}
-    >
-      <Text
-        style={styleSet.buttonText}
-        onLayout={textElement => {
-          this.viewContainer.setNativeProps({
-            style: {
-              width: textElement.nativeEvent.layout.width + 10, // extra padding to prevent text wrapping (text container must be always bigger)
-            },
-          });
-        }}
-      >
-        {badgeText}
-      </Text>
-    </View>
-  );
+  recalculateWidth = (textElement: TextOnLayout) => {
+    if (!this.viewContainerReference) {
+      return;
+    }
+    // $FlowFixMe: FLow is broken for Native, this is correct: http://facebook.github.io/react-native/releases/0.49/docs/direct-manipulation.html
+    this.viewContainerReference.setNativeProps({
+      style: {
+        width: textElement.nativeEvent.layout.width + 10, // extra padding to prevent text wrapping (text container must be always bigger)
+      },
+    });
+  };
+
+  render = () => {
+    let styleSet = unverifiedStyle;
+    let badgeText = 'UNVERIFIED';
+
+    if (this.props.verified === true) {
+      styleSet = verifiedStyle;
+      badgeText = 'VERIFIED';
+    }
+
+    return (
+      <View style={styleSet.button} ref={this.storeContainerReference}>
+        <Text style={styleSet.buttonText} onLayout={this.recalculateWidth}>
+          {badgeText}
+        </Text>
+      </View>
+    );
+  };
 }
 
 const basicButtonStyle = {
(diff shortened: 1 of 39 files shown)


app/common/src/badge/VerificationBadge.js
```diff
@@ -22,36 +33,47 @@ type Props = {|
...
+   */
+  storeContainerReference = (reference: React.Ref<typeof View> | null) =>
+    (this.viewContainerReference = reference);
+
+  recalculateWidth = (textElement: TextOnLayout) => {
+    if (!this.viewContainerReference) {
+      return;
+    }
+    // $FlowFixMe: FLow is broken for Native, this is correct: http://facebook.github.io/react-native/releases/0.49/docs/direct-manipulation.html
+    this.viewContainerReference.setNativeProps({
```
mrtnzlml: I really don't know how to solve this better without redefining React.Ref<>...

app/native/src/components/relay/PrivateApiRenderer.js
```diff
@@ -1,53 +1,31 @@
...
+import type { Props as PublicProps } from './PublicApiRenderer';
 
 type Props = {|
-  query: string,
-  render: ({ error: Object, props: Object }) => React.Node,
-  variables?: Object,
-  cacheConfig?: {|
-    force: boolean,
-  |},
+  ...PublicProps,
```
mrtnzlml: I am amazed this works. Usually, it's not possible to spread exact types. Maybe something else is wrong?

mrtnzlml:
Please test these changes in [Expo](https://docs.expo.io/versions/latest/introduction/installation.html#mobile-client-expo-for-ios-and-android):

  QR Code

app/native/src/components/relay/QueryRenderer.js
```diff
@@ -0,0 +1,45 @@
...
+
+export default class QueryRenderer extends React.Component<Props> {
+  renderRelayContainer = ({
+    error,
+    props,
+  }: {
+    error: Object,
+    props: Object,
+  }) => {
+    // FIXME: it Relay swallows all GraphQL errors?
```
jaroslav-kubicek: _Does_ Relay swallow 👮

app/native/src/screens/allBookings/AllBookingsListNode.js
```diff
@@ -0,0 +1,46 @@
...
+import type { AllBookingsListNode as AllBookingsListNodeType } from './__generated__/AllBookingsListNode.graphql';
+import type { Navigation } from '../../types/Navigation';
+
+type Props = {
+  data: AllBookingsListNodeType,
+  navigation: Navigation,
+  showSeparator: boolean,
+};
+
+class AllBookingsListNode extends React.Component<Props> {
```
jaroslav-kubicek: What do you think about recompose for future? there are many component class (instead functions) only because of some handlers...

jaroslav-kubicek:
seems like there are some updates related to graphql beyond the jsx-no-bind? ⚠️

mrtnzlml:
Yes, as I said it's a little bit tricky. It forces you to use better component structure and that means that you need to create new components. This leads to smaller Relay fragments => different generated GraphQL files structure.

app/common/src/badge/VerificationBadge.js
mrtnzlml: I think it's quite usual in React Native and the official type works with React but not with RN (React doesn't have `setNativeProps`).

app/common/src/cards/__tests__/SimpleCard.test.js
```diff
@@ -5,14 +5,16 @@ import renderer from 'react-test-renderer';
 
 import SimpleCard from '../SimpleCard';
 
+function onPressVoidCallback() {}
```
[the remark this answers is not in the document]
mrtnzlml: I will try it again. I pretty sure that in some cases the array function instead of normal function caused the same problem I am trying avoid here. Let me check later... :)

app/native/src/components/relay/QueryRenderer.js
mrtnzlml: 🤦‍♂️ This is what happens when you are rewriting just part of the message... 😄

app/native/src/screens/allBookings/AllBookingsListNode.js
mrtnzlml: I think there is nothing wrong with classes. We can discuss this later, not in this PR.

jaroslav-kubicek:
Cool, probably I missed that, thanks

app/native/src/components/authentication/EmailLoginForm.js
```diff
@@ -24,6 +24,29 @@ export default class SimpleLoginForm extends React.Component<Props, State> {
...
+    this.setState({
+      username: text,
+    });
+
+  _handlePasswordChange = text =>
+    this.setState({
+      password: text,
+    });
+
+  _handleFormSubmit = () =>
```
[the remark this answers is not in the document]
mrtnzlml: Out of scope now. Also, don't overuse Redux. It's not needed in this case because I am only changing state inside of this component. We can improve it later...

app/native/src/components/authentication/GoogleLoginForm.js
```diff
@@ -21,37 +21,39 @@ export default class GoogleLogin extends React.Component<Props, State> {
     loading: false,
   };
 
+  handleLogginAction = async () => {
+    this.setState({ loading: true });
+    const accessToken = await GoogleAuth.signIn();
+    if (accessToken) {
+      const jsonPayload = await (await fetch(
```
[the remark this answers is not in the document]
mrtnzlml: Same with the email login. Out of scope now but we can improve it later (if needed). I quite believe that this is the only place where you'll need to call this and you don't need to change global state (yet). I am very OK with mixing global (Redux) state and local state (as it should be). Let's modify it when the requirements change...

app/common/src/cards/__tests__/SimpleCard.test.js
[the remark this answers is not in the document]
mrtnzlml: Ok, in this case, it's not a problem and in other situations it's avoidable. But let's agree on one thing: we should not write these things in code reviews because it should be enforced by eslint rules. Instead, let's focus on real issues with code design and similar. If we don't like some code style (and I agree it's weird, similar to class methods `likeThis` or `_likeThis`) we should send PR with new eslint rule. Does it make sense for you? (to focus on important things in code reviews and enforcing good habits by eslint)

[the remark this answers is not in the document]
jaroslav-kubicek: of course, I agree, I expect the usual scenario: discussion in PR => (agreement on setting new rule => new PR) || not important 🙂

Issue #23, opened 2017-11-10
Investigate how to properly use ENV variables in RN


mrtnzlml:
https://medium.com/differential/managing-configuration-in-react-native-cd2dfb5e6f7b

(read the comments)

mrtnzlml:
We currently do not need ENVs since we are not creating standalone application yet. Will be solved later when needed. We can now integrate it easily with native code.

PR #49, opened 2017-12-04
Basic hotel map

### LINKED ISSUE #37: Map for all hotels
Please prepare a map for all hotels. It's probably easiest to use https://docs.expo.io/versions/latest/sdk/map-view.html

It may be tricky for 2 reasons:
- you need to take care of Google maps on Android (see "Deploying to a standalone app on Android")
- you'll need to adjust map center and zoom based on markers (maybe it's already a feature in the `airbnb/react-native-maps` package - I don't know)

To resolve this issue we should have 1) working map and 2) we should be able to see let's say 3 random markers (with proper map zoom and center).

zpl://screen?sid=5a17728101b10a0c81184686&pid=5a1308cdae8ae89cae5c15af

jaroslav-kubicek:
This PR closes #37 

Note: Currently, the **map is not working on Android**. See #51

image


mrtnzlml:
Please test these changes in [Expo](https://docs.expo.io/versions/latest/introduction/installation.html#mobile-client-expo-for-ios-and-android):

  QR Code

VaclavSir:
## Markers
- [ ] The marker should be positioned so that the arrow points to the coordinates.
Otherwise it points to different places depending on zoom and rotation:
screen shot 2017-12-06 at 14 05 41
screen shot 2017-12-06 at 14 05 55

Compare with Google Maps, where the marker always points to the same point:
screen shot 2017-12-06 at 14 17 12
screen shot 2017-12-06 at 14 17 28


## Navigation
The navigation should look like this:
screen shot 2017-12-06 at 14 08 01
Not like this:
screen shot 2017-12-06 at 14 08 19
I'm not sure if the navigation design belongs to this issue. 🤔

app/common/src/map/PriceMarker.js
```diff
@@ -0,0 +1,85 @@
...
+    flexDirection: 'row',
+    paddingVertical: 5,
+    paddingHorizontal: 8,
+    backgroundColor: '#fff',
+    borderRadius: 3,
+    borderColor: '#fff',
+    borderWidth: 2,
+  },
+  bubbleSelected: {
+    backgroundColor: '#4da2ab',
```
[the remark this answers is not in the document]
mrtnzlml:
```js
import { Color } from '@kiwicom/native-common';
// Color.brand
```

mrtnzlml:
> I'm not sure if the navigation design belongs to this issue. 🤔

@VaclavSir We are not taking care of navigation. We are doing only components inside of navigation screens.

mrtnzlml:
> Currently, the map is not working on Android.

@jaroslav-kubicek Please solve it without ENV variables now. The result should be "it works with Android but we need to solve #23 first and setup ENV variable"...

app/hotels/src/AllHotelsMap/index.js
```diff
@@ -0,0 +1,118 @@
...
+  };
+
+  fitToSuppliedMarkers(animate = true) {
+    const markerIds = this.props.markers.map(marker => marker.id);
+
+    if (!this.map) {
+      return;
+    }
+
+    this.map.fitToSuppliedMarkers(markerIds, animate);
```
mrtnzlml: Ok, this was really easy... 😄 But it may need additional fiddling:

simulator screen shot - iphone 8 - 2017-12-07 at 10 07 02

app/hotels/src/types/datastructures.js
```diff
@@ -0,0 +1,16 @@
...
+  id: string,
+  coordinate: LatLng,
+  data: {|
+    price: number,
+  |},
+|};
+
+export type CurrencyInfo = {|
+  currency: string,
+  position: 'before' | 'after',
```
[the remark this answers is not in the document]
VaclavSir: Now that I think about it, the design in Zeplin shows price with the ISO currency code. That would be always "after".

Airbnb uses local currency symbol placed "before", because they are Americans, but "Kč1,000" is IMHO just wrong.

If we decide to complicate it more in the future, note that on the Frontend we sometimes render something before as well as after the amount, typically for different dollars ("$100 CAD" vs "$100 AUD").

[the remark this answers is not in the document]
mrtnzlml: My main concern is: do we know what is before and what after or do we have to figure it on our own? Also very good point with the dollars...

mrtnzlml:
Sorry, I was curious what is the "Update" button for and it merged master to this branch... 🤦‍♂️ Try to squash it with your upcoming changes or I will deal with this during the merge if it's too complicated.

mrtnzlml:

  QR Code

app/hotels/src/types/datastructures.js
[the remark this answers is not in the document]
jaroslav-kubicek: Ohh, yes, I have difficulty how to grasp this formatting correctly, as I don't know how data will look, I will probably just remove all this formatting stuff for now and solve it later when it will be more clear.

jaroslav-kubicek:
Ad markers on map: this is the feature of the map itself and it works the same as in AirBnb app. Possible workaround would be expensive and I would rather avoid that this time 🙂  cc @VaclavSir

app/hotels/src/types/datastructures.js
```diff
@@ -0,0 +1,16 @@
+// @flow
```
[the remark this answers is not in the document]
jaroslav-kubicek: no, it's bad idea

app/common/src/map/PriceMarker.js
```diff
@@ -0,0 +1,85 @@
...
+    <View style={styles.container}>
+      <View style={[styles.bubble, isSelected && styles.bubbleSelected]}>
+        {props.currencyInfo.position === 'before' && renderCurrency(props)}
+        <Text style={[styles.price, isSelected && styles.priceSelected]}>
+          {props.price}
+        </Text>
+        {props.currencyInfo.position === 'after' && renderCurrency(props)}
+      </View>
+      <View style={[styles.arrowBorder]} />
+      <View style={[styles.arrow, isSelected && styles.arrowSelected]} />
```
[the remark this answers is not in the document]
jaroslav-kubicek: sounds good, but it smells by code duplication (see arrow & arrowSelected for example)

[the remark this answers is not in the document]
mrtnzlml: I am pretty sure you can figure out how to do it without duplication. Try it and if it's not going to be more readable, we can discard my comment... :)

VaclavSir:
I can't check on iPhone, but Aibnb on Android doesn't work like that, if I zoom out from a marker, it still points to the same point.

I don't see the expensiveness, all it needs is some negative margin.

app/common/src/map/PriceMarker.js
```diff
@@ -0,0 +1,85 @@
...
+    <View>
+      <View style={[styles.bubble, isSelected && selectedStyles.bubble]}>
+        <Text style={[styles.price, isSelected && selectedStyles.price]}>
+          {props.price}
+        </Text>
+        <Text
+          style={[styles.currency, props.isSelected && selectedStyles.currency]}
+        >
+          CZK
+          {/* FIXME probably some price & currency formatting is necessary */}
```
VaclavSir: You know what I think about todo comments. Plz delete it.

jaroslav-kubicek:
All should be fixed now, some additional notes:
- You need to provide api key for Android maps (readme info added & app.json updated)
- marker positions: I might be missing something, but the exact position is changing a little bit as you are zooming in/out, which would require us to calculate margin dynamically based on current region

mrtnzlml:
> it's fake news, second param can be only boolean, region positioning will change anyway soon in next PR

What are you talking about? I can quite clearly see `options` object... https://github.com/react-community/react-native-maps/blob/master/lib/components/MapView.js#L535 🙂

app/common/src/map/PriceMarker.js
```diff
@@ -0,0 +1,95 @@
...
+import * as React from 'react';
+import { StyleSheet, View, Text } from 'react-native';
+
+import Color from '../Color';
+
+const styles = StyleSheet.create({
+  container: {
+    // price marker is wider than default, so needs to be shifted more to the left
+    marginLeft: -30,
+    marginTop: -10,
```
mrtnzlml: I would not do that, it's very problematic (at least not that easy) and we are not able to do it properly now. It doesn't work if you are rotating the map, it doesn't work if you are changing angle (2D -> 3D map) so I cannot really see the improvement. Can we just keep it "not working" without margin? 🙂 Let's fix it properly later if needed.

app.json
```diff
@@ -10,7 +10,12 @@
       "bundleIdentifier": "com.kiwi.rn"
     },
     "android": {
-      "package": "com.kiwi.rn"
+      "package": "com.kiwi.rn",
+      "config": {
+        "googleMaps": {
+          "apiKey": "** GOOGLE MAPS API KEY HERE **"
```
mrtnzlml: I am curious - how does it work if you do not have Google Maps installed on your device?

app/hotels/README.md
```diff
@@ -21,6 +21,8 @@ import {
 <AllHotelsMap />
 ```
 
+Note: Currently you need to set Google Maps Api key manually to make it work on Android (see app.json), follow instructions in Expo [MapView Docs](https://docs.expo.io/versions/latest/sdk/map-view.html#deploying-to-a-standalone-app-on-android)
```
mrtnzlml: How does it work if I want to use this component outside of our project? Because native developer will not have `app.json`...

app/common/src/map/PriceMarker.js
```diff
@@ -0,0 +1,95 @@
...
+    backgroundColor: 'transparent',
+    borderColor: 'transparent',
+    borderWidth: 7,
+    borderTopColor: '#fff',
+    alignSelf: 'center',
+    marginTop: -18,
+  },
+});
+
+const selectedStyles = StyleSheet.create({
```
mrtnzlml: My idea was to use `StyleSheet.flatten` or spread operator or `_.defaultsDeep` before using it in the `render` method because then you can do just:

```js
<View style={styles.bubble}> // <<< Just "styles" because we already have final style...
  <Text> // <<< Uhm, why do we need this component here?
    <Text style={styles.price}>
      {props.price}
      // ...
```

app/common/src/map/PriceMarker.js
```diff
@@ -0,0 +1,95 @@
...
+          <Text style={[styles.price, isSelected && selectedStyles.price]}>
+            {props.price}
+          </Text>
+          <Text
+            style={[
+              styles.currency,
+              props.isSelected && selectedStyles.currency,
+            ]}
+          >
+            CZK
```
mrtnzlml: Should be probably `{props.currency}` or something like this...

app/common/src/map/PriceMarker.js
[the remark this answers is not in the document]
mrtnzlml: I read this again and let me respond differently: Why code duplication? This is not related to the `isSelected` condition. One component for the pure marker (`MapMarker`) with all the styles and second component with additional content (`PriceMapMarker`). So the second component will basically be only:

```js
render = () => <MapMarker>{props.price} {props.currency}</MapMarker>
```

The point is to separate the concerns because 1) we may easily reuse the marker itself and 2) we may have quite a lot of logic here (before, after currency, ...). But since we are no longer dealing with currency formatting, I am ok to leave it now "as is" until we decide to improve it. I just wanted to make my previous comment more clear... 🙂

jaroslav-kubicek:
> What are you talking about? I can quite clearly see options object... 

that is valid for `fitToCoordinates`, not for `fitToSuppliedMarkers`, although I could take marker coords and use this function instead probably 🤔

mrtnzlml:
Isn't this snippet exactly what you need? https://github.com/react-community/react-native-maps/issues/922#issuecomment-308670547

app/common/src/map/PriceMarker.js
jaroslav-kubicek: ahh, ok, thanks for idea

wrapping text is necessary for multiple text elements with various heighs, like in this case, so the smaller text is not flying in the air and it's aligned to the baseline

app.json
jaroslav-kubicek: it should use google maps tiles, not google maps app itself imho

jaroslav-kubicek:
round 3 - @mrtnzlml 
- marker styles updated
- I updated README with the link to installation instruction
- displayed region refactored to `fitToCoordinates`

thanks for cooperation 🙂

mrtnzlml:
Merged, thank you! 🙂

PR #50, opened 2017-12-05
Update comment that is automatically put into PR conversation

jaroslav-kubicek:
Reason: While PR is building, reviewer might be confused by old version of app.

It takes some time before build is finished, so I would like to avoid confusion by improving the comment (It already happened to me and you may forget this fact again)


mrtnzlml:

  QR Code

mrtnzlml:
Yes, makes sense... 🙂

PR #53, opened 2017-12-07
Search: add filters stripe

### LINKED ISSUE #36: Hotels "stripe filter"
Please prepare stripe with filter buttons (you'll find `Color.brand` in `@kiwicom/native-common` package). At this moment it's not very clear what are all possible values and equivalent API parameters so let's prepare it just like a presentational component without functionality (but with final design - should be really simple).

jakubzaba:
Closes #36


app/hotels/src/components/filters/FilterButton.js
```diff
@@ -0,0 +1,56 @@
...
+    dispatch({
+      type: 'toggleFilter',
+      filter,
+    });
+  },
+});
+
+export default connect(mapStateToProps, mapDispatchToProps)(
+  UnwrappedFilterButton,
+);
```
mrtnzlml: It seems really weird to me that simple on/off button needs Redux. Is this really problem of the button? Shouldn't this be problem of the filter? Why do you need Redux after all?

app/hotels/src/reducers/SearchReducer.js
```diff
@@ -0,0 +1,35 @@
+// @flow
+
+import type { ReduxState, SearchActions } from '../types/Redux';
+
+type SearchState = $PropertyType<ReduxState, 'search'>;
+
+const InitialFilter = {
+  isActive: false,
```
mrtnzlml: I am afraid that it's not only about `on/off` but it's more fuzzy. You can for example have `off/"1,2,4"` or `off/>123` and so on...

mrtnzlml:

  QR Code

mrtnzlml:
@jakubzaba What about this? It's sitting here for days and there are still several issues:

- the design is still not accurate - see Zeplin
- you cannot toggle two filters at the same time

jakubzaba:
I mentioned some reason why design is not the same. Can we finish the design later if it's really that important to use unusual buttons.
When you toggle filter popup with overlay will show up. When you apply the filter it becomes active so you can toggle other filter. I don't see the problem there.

mrtnzlml:
> I rather avoid using non-standard elements since it takes a lot of time to make it perfect in contrast with simple button with icon from some library.

> Can we finish the design later if it's really that important to use unusual buttons.

Hm, we have a design in Zeplin and we cannot make our own just because it's too complicated. Also, the important part of this issue was to create a final design. If you really think it's a stupid idea and it's better to create our own design, please talk with @MRTNKCR. I cannot accept this without approval.

> When you toggle filter popup with overlay will show up. When you apply the filter it becomes active so you can toggle other filter. I don't see the problem there.

I tried the filter and it seems that if I toggle first and then click the second, the first is deactivated. I would expect that you can toggle the buttons independently. Meaning you can turn on all the buttons (not only one at the time).

app/common/src/buttons/Button.js
```diff
@@ -1,7 +1,9 @@
 // @flow
 
 import * as React from 'react';
+import { Icon, type IconProps } from 'react-native-elements';
```
mrtnzlml: Is it possible to use this? https://docs.expo.io/versions/latest/guides/icons.html

app/hotels/src/filter/types.js
```diff
@@ -0,0 +1,11 @@
+// @flow
+
+type FilterConfiguration = Object;
+
+export type Filter = {
```
mrtnzlml: I am missing exact typing here (if possible).

app/hotels/src/filter/FilterStripe.js
```diff
@@ -0,0 +1,78 @@
...
+    this.setState({ filters });
+  };
+
+  _renderFilterButton = ({ title, icon, filter, isActive }: Filter) => (
+    <FilterButton
+      key={filter}
+      title={title}
+      icon={icon}
+      isActive={isActive}
+      onPress={() => this._handleButtonPress(filter)} // eslint-disable-line react/jsx-no-bind
```
mrtnzlml: This is "no go" for me. We have this rule for good reasons... :)

mrtnzlml:
Now it honestly looks and feels really good... 👍 🙂

mrtnzlml:
Integrated, thanks. I had to refactor it a little bit to fix the broken buttons...

mrtnzlml:
@jakubzaba Also, if you send something after [this comment](https://github.com/kiwicom/react-native-app/pull/53#issuecomment-352968762) then it's not integrated. Send another PR and for the future reference please make clear that you are still working on it...

jakubzaba:
There was just a conflict.

PR #54, opened 2017-12-07
Show region based on markers & navigate to selected marker

jaroslav-kubicek:
- Compute region manually from markers so the right place is displayed from the beginning (to avoid initial zooming animation, e.g. Europe => CZ => Prague)
- after map is ready (=whole region is loaded), animate to the currently selected marker

Must come after #49

### DIFF
--- a/app/hotels/src/AllHotelsMap.js
+++ b/app/hotels/src/AllHotelsMap.js
@@ -2,8 +2,9 @@
 
 import * as React from 'react';
 import { StyleSheet, View } from 'react-native';
-import { MapView, type LatLng } from 'expo';
-import { PriceMaker } from '@kiwicom/react-native-app-common';
+import { MapView, type LatLng, type Region } from 'expo';
+import { PriceMarker } from '@kiwicom/react-native-app-common';
+import { getCenter, getBounds } from 'geolib';
 
 export type HotelMarker = {|
   id: string,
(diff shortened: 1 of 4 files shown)


mrtnzlml:

  QR Code

app/hotels/src/AllHotelsMap/index.js
```diff
@@ -0,0 +1,186 @@
...
+      longitude: boundaries.minLng,
+    };
+    const bottomRight = {
+      latitude: boundaries.minLat,
+      longitude: boundaries.maxLng,
+    };
+    const center = getCenter([topLeft, bottomRight]);
+
+    // FIXME do it better
+    const BULGARIAN_CONSTANT = 1.3;
```
mrtnzlml: Yes, please do it better or at least explain what is the idea behind... 🙂

jaroslav-kubicek:
Should be ready for CR, cc @mrtnzlml

mrtnzlml:
Can you please also fix the overlapping of markers? The selected one should have bigger `zIndex` or something like this so it's always on top of the other markers.

app/hotels/src/AllHotelsMap.js
```diff
@@ -15,16 +16,27 @@ export type HotelMarker = {|
...
-  right: 50,
-  bottom: 50,
-  left: 50,
+type State = {|
+  region: Region,
+  mapIsReady: boolean,
+  selectedMarkerId: string,
+|};
+
+type MarkerPressEvent = {
```
mrtnzlml: Is it possible to make it exact? The same for `Props`.

jaroslav-kubicek:
Should be fixed. Also, I investigated the markers and the result is:
- order in which are rendered and it doesn't matter
- position: absolute with xIndex seems to be ignored as well
Nevertheless, I will continue in searching for solution

mrtnzlml:
> position: absolute with xIndex seems to be ignored as well
Nevertheless, I will continue in searching for solution

Don't waste your time on this now. Let's ship the product first and fix all the shit later. I tried, it should work and it doesn't - we can spend days on fixing it or finish the hotels. You did good job here, let's continue... 🙂

mrtnzlml:
Merged thank you: https://github.com/kiwicom/react-native-app/commit/b7c5b1eb921820014a069d412f78902555d05291

I just removed the exact type wizardry. Flow is sometimes broken and it doesn't worth to make it work... 🙂

PR #57, opened 2017-12-12
Refactor map component

jaroslav-kubicek:


mrtnzlml:

  QR Code

jaroslav-kubicek:
As I'm not able to achieve workable swipe cards, I'm giving up on this PR, removing previous attempt with `react-native-swiper` and leaving just refactoring of the component.

jaroslav-kubicek:
Just as an explanation,  `index` prop in `react-native-swiper` can't simply be updated - it's indented only as *initial* value

mrtnzlml:
> I'm giving up

I don't know what this means. This is not in Kiwi.com vocabulary... 🙂 

Also, please be more careful with the commit message. Message "Add swipe cards to the map" is not really accurate in this case.

PR #63, opened 2017-12-15
Introduce new component NetworkImage

mrtnzlml:
This component is taking care of loading indicator and weird edge
cases of the original Image component from React Native.


mrtnzlml:

  QR Code

jaroslav-kubicek:
good job 👍  I'll use it immediately

PR #64, opened 2017-12-18
Add less embarrassing loading indicator

mrtnzlml:
Main inspiration: https://www.kiwi.com/storybook/?selectedKind=Components%2FButton&selectedStory=loading-blue&full=0&down=1&left=1&panelRight=1&downPanel=storybook%2Factions%2Factions-panel (I basically copy pasted the behavior)

### DIFF
--- a/app/common/src/loaders/FullPageLoading.js
+++ b/app/common/src/loaders/FullPageLoading.js
@@ -3,12 +3,12 @@
 import * as React from 'react';
 
 import CenteredView from '../view/CenteredView';
-import SimpleLoading from './SimpleLoading';
+import IconLoading from './IconLoading';
 
 export default function FullPageLoading() {
   return (
     <CenteredView>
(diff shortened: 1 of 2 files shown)


mrtnzlml:

  QR Code

mrtnzlml:
It is not used. But I did not remove it in this PR. I can probably do that (tomorrow). Or you can during merge... :)

mrtnzlml:
The idea behind why I didn't do it was: It doesn't have to be necessary in one PR. I am adding a new component and they can live together (next to each other). I can keep the better one and remove the old one later.

jaroslav-kubicek:
ok, that's sufficient explanation, I was curious whether it was intentional or just forgotten

mrtnzlml:
> Note think about removing old simple loader

Ok, fuck it: https://github.com/kiwicom/react-native-app/pull/68 🙂

PR #65, opened 2017-12-18
Add snap cards on bottom of the map with details about displayed hotels

jaroslav-kubicek:
screen shot 2017-12-18 at 14 18 41

*Note*  - there is one todo related to flow types to be done

Any feedback from user testing on how it feels when you are using it, is welcomed 🙂

### DIFF
--- a/app/common/index.js
+++ b/app/common/index.js
@@ -20,16 +20,26 @@ export { default as NetworkImage } from './src/image/NetworkImage';
 export { default as FullPageLoading } from './src/loaders/FullPageLoading';
 
 export { default as PriceMarker } from './src/map/PriceMarker';
+export { default as HotelCard } from './src/map/HotelCard';
+export { default as Address } from './src/map/Address';
 
 export { default as Large } from './src/text/Large';
 
 export { default as CenteredView } from './src/view/CenteredView';
 export { default as Layout } from './src/view/Layout';
 export { default as LayoutWithoutHeader } from './src/view/LayoutWithoutHeader';
+export {
+  default as VerticalSwipeResponder,
+} from './src/view/VerticalSwipeResponder';
 
 export { default as Color } from './src/Color';
 export { default as Price } from './src/Price';
 
 // Flow types:
 
-export type { OnFocus, OnLayout } from './types/Events';
+export type {
+  OnFocus,
+  OnLayout,
+  OnDimensionsChange,
+  GestureState,
+} from './types/Events';
--- a/app/common/src/map/Address.js
+++ b/app/common/src/map/Address.js
@@ -1,0 +1,47 @@
+// @flow
+
+import * as React from 'react';
+import { StyleSheet, View, Text } from 'react-native';
+import { Ionicons } from '@expo/vector-icons';
+import Color from '../Color';
+
+type Props = {|
+  address: string,
+|};
+
+const styles = StyleSheet.create({
+  container: {
+    flexDirection: 'row',
+    borderTopColor: Color.grey._100,
+    borderTopWidth: 1,
+    paddingVertical: 10,
+    marginHorizontal: 10,
+  },
+  mapIcon: {
+    alignSelf: 'center',
+    marginRight: 10,
+  },
+  header: {
+    fontWeight: 'bold',
+  },
+});
+
+class Address extends React.Component<Props> {
+  render = () => {
+    const { address } = this.props;
+
+    return (
+      <View style={styles.container}>
+        <View style={styles.mapIcon}>
+          <Ionicons name="md-map" size={24} color={Color.brand} />
+        </View>
+        <View>
+          <Text style={styles.header}>Address</Text>
+          <Text>{address}</Text>
+        </View>
+      </View>
+    );
+  };
+}
+
+export default Address;
--- a/app/common/src/map/HotelCard.js
+++ b/app/common/src/map/HotelCard.js
@@ -1,0 +1,58 @@
+// @flow
+
+import * as React from 'react';
+import { StyleSheet, View, Text, Image } from 'react-native';
+import Color from './../Color';
+
+type Props = {|
+  width: number,
+  hotel: {|
+    name: string,
+  |},
+|};
+
+const styles = StyleSheet.create({
+  description: {
+    flexDirection: 'row',
+  },
+  image: {
+    width: 50,
+    height: 50,
+    borderRadius: 5,
+    marginRight: 10,
+  },
+  hotelName: {
+    fontSize: 18,
+  },
+  metainfo: {
+    color: Color.grey._700,
+    opacity: 0.8,
+  },
+  price: {
+    color: Color.brand,
+    fontWeight: 'bold',
+  },
+});
+
+const HotelCard = ({ width, hotel }: Props) => (
+  <View style={[{ width }]}>
+    <View style={styles.description}>
+      <View>
+        <Image
+          style={styles.image}
+          source={{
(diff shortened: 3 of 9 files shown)


mrtnzlml:

  QR Code

jaroslav-kubicek:
omg, FYI device orientation needs to be fixed...

mrtnzlml:
I don't know man. It feels so *not smooth*. I always drag like I would normally do but it waits for a while and that it finishes the movement with higher speed. Feels so weird.

Also, at some point, it scrolls to the very beginning (very fast). But if you continue long enough with scrolling it will never happen again (not nice).

Is the address not changing at all or is it still the same so I cannot see the effect?

> Device orientation needs to be fixed

Please do so.

app/common/src/map/Address.js
```diff
@@ -0,0 +1,40 @@
...
+class Address extends React.Component<Props> {
+  render = () => {
+    return (
+      <View style={styles.container}>
+        <View style={styles.mapIcon}>
+          <Ionicons name="md-map" size={24} color={Color.brand} />
+        </View>
+        <View>
+          <Text style={{ fontWeight: 'bold' }}>Address</Text>
+          <Text>950 Candice Wells Suite 375, Dublin</Text>
```
mrtnzlml: I would expect props here instead of text.

app/common/src/map/HotelSnap.js
```diff
@@ -0,0 +1,58 @@
...
+    color: Color.grey._700,
+    opacity: 0.8,
+  },
+  price: {
+    color: Color.brand,
+    fontWeight: 'bold',
+  },
+});
+
+const HotelSnap = ({ width, hotel }: Props) => (
```
mrtnzlml: Can we find a better name for this? Other than "snap".

app/hotels/src/map/HotelSnaps.js
```diff
@@ -0,0 +1,90 @@
...
+  componentDidMount = () => {
+    Dimensions.addEventListener('change', this.onDimensionsChanged);
+  };
+
+  componentWillUnmount = () => {
+    Dimensions.removeEventListener('change', this.onDimensionsChanged);
+  };
+
+  onDimensionsChanged = ({ screen: { width } }: OnDimensionsChange) => {
+    this.setState({ availableWidth: width });
```
mrtnzlml: Why is this needed? Cannot we just use "width" from `Dimensions`? The trick is that you have to use it inside of render function so it gets updated every time you rerender (this is what you want anyway).

app/hotels/src/map/HotelSnaps.js
```diff
@@ -0,0 +1,90 @@
...
+    const { availableWidth } = this.state;
+    const width = availableWidth * SNAP_WIDTH;
+
+    return <HotelSnap width={width} hotel={item.data} />;
+  };
+
+  render = () => {
+    const { items, selectedIndex, onSnapToItem } = this.props;
+    const { availableWidth } = this.state;
+    const itemWidth = availableWidth * SNAP_WIDTH;
```
mrtnzlml: Seems like this is calculated twice (render and renderItem). Try to be DRY... 🙂

app/hotels/src/map/HotelSnaps.js
jaroslav-kubicek: IMHO component is not re-rendered because the device has rotated, I even tried that, that's why I added lisneter, but I may miss something you already know...

mrtnzlml: I did it here: https://github.com/kiwicom/react-native-app/blob/master/app/hotels/src/gallery/PhotosStripe.js#L23 (just to fix exactly the same issue)

mrtnzlml: It's described here: https://facebook.github.io/react-native/docs/dimensions.html#get But now I really don't know why does it work... ❓

app/common/src/map/Address.js
jaroslav-kubicek: fixed

app/hotels/src/map/HotelSnaps.js
jaroslav-kubicek: fixed

app/hotels/src/map/HotelSnaps.js
mrtnzlml: Ok I bet it's about `onLayout` change in the `Swiper` so it doesn't work exactly how I thought... 🙂

app/common/src/map/HotelSnap.js
jaroslav-kubicek: fixed (hopefully)

app/common/src/map/HotelCard.js
```diff
@@ -0,0 +1,58 @@
...
+});
+
+const HotelCard = ({ width, hotel }: Props) => (
+  <View style={[{ width }]}>
+    <View style={styles.description}>
+      <View>
+        <Image
+          style={styles.image}
+          source={{
+            uri: 'http://q.bstatic.com/images/hotel/square128/117/11753593.jpg',
```
mrtnzlml: Should I care now or we will connect the real data later?

jaroslav-kubicek: I plan to do new PR for real data :)

jaroslav-kubicek:
> I don't know man. It doesn't feel very smooth. I always drag like I would normally do but it waits for a while and that it finishes the movement with higher speed. Feels so weird.

I removed momentum prop and added `decelerationRate`, which seems to be slightly smoother

> Also, at some point, it scrolls to the very beginning

valid point when `firstItem !== 0`, I expect `firstItem === 0` by default when you simply navigate to map by hitting "map" button in topbar, will investigate to make it work always.

Add resizing - I tried and it doesn't work without listeners :/

jaroslav-kubicek:
Related issue: https://github.com/archriss/react-native-snap-carousel/issues/232

mrtnzlml:
Let me know about the progress. It's not OK to merge obviously broken code...

jaroslav-kubicek:
workarounded :) cc @mrtnzlml

app/common/index.js
```diff
@@ -20,16 +20,26 @@ export { default as NetworkImage } from './src/image/NetworkImage';
 export { default as FullPageLoading } from './src/loaders/FullPageLoading';
 
 export { default as PriceMarker } from './src/map/PriceMarker';
+export { default as HotelCard } from './src/map/HotelCard';
+export { default as Address } from './src/map/Address';
```
mrtnzlml: I do not think this should be in this shared package "just in case". I do not think there is high chance that we'll reuse these components. It's absolutely fine to create local components and use this only for true "shared" things. I am actually considering rename `common` to `shared` (like in React). I'll leave it here for now. I just want to mention it... 🙂

mrtnzlml:
> workarounded

I can reproduce it again just with a different flow. Try to click few times on different map markers and then try to swipe. It's broken again. I'll merge it for now because everyone who is not actively creating this project is bitching around but we will need to spend few more hours on this swiper. It's not very good yet.

mrtnzlml:
There is also this warning:

> VirtualizedList: You have a large list that is slow to update - make sure your renderItem function renders components that follow React performance best practices like PureComponent, shouldComponentUpdate, etc.
Object {
   "contentLength": 6003,
   "dt": 2661,
   "prevDt": 1851,
 }

mrtnzlml:
Merged.

app/common/index.js
jaroslav-kubicek: good point, I'm actually going to change it today :)

PR #66, opened 2017-12-18
Search: add search form

### LINKED ISSUE #35: Hotels search form
Please create hotels search form. For this, you should modify and reuse `DatePicker` and `TextInput` from `@kiwicom/native-common` package. Do just simple form (location API is not ready, guest popup and the final design will be in next issue). Now it's important to have good component design and proper form behavior. Also please connect it to the component variables created in #34.

zpl://screen?sid=5a177284d9724be52c416a4d&pid=5a1308cdae8ae89cae5c15af

jakubzaba:
Closes #35

### DIFF
--- a/app/hotels/src/allHotels/AllHotels.js
+++ b/app/hotels/src/allHotels/AllHotels.js
@@ -2,17 +2,62 @@
 
 import * as React from 'react';
 import { graphql } from 'react-relay';
+import { View, StyleSheet } from 'react-native';
 import { PublicApiRenderer } from '@kiwicom/react-native-app-relay';
+import moment from 'moment';
 
+import SearchForm from './SearchForm';
 import AllHotelsSearchList from './AllHotelsSearchList';
 
 import type { AllHotelsSearchQueryResponse } from './__generated__/AllHotelsSearchQuery.graphql';
+import type { Search } from './types';
+
+const styles = StyleSheet.create({
+  wrapper: {
+    flex: 1,
+  },
+});
 
 type Props = {|
   onGoToSingleHotel: () => void,
 |};
 
-export default class AllHotelsSearch extends React.Component<Props> {
+type State = {|
+  search: Search,
+|};
+
+export default class AllHotelsSearch extends React.Component<Props, State> {
+  state = {
+    search: {
+      latitude: 50.08,
+      longitude: 14.44,
+      checkin: null,
+      checkout: null,
+      roomsConfiguration: {
+        adultsCount: 1,
+      },
+    },
+  };
+
+  _handleSearchChange = (search: Search) => {
+    this.setState({
+      search: {
+        latitude: 50.08,
+        longitude: 14.44,
+        checkin: search.checkin,
+        checkout: search.checkout,
+        roomsConfiguration: search.roomsConfiguration,
+      },
+    });
+  };
+
+  isReadyToSearch = () => {
+    const { search } = this.state;
+    return (
+      search.latitude && search.longitude && search.checkin && search.checkout
+    );
+  };
+
   renderInnerComponent = (propsFromRenderer: AllHotelsSearchQueryResponse) => (
     <AllHotelsSearchList
       data={propsFromRenderer.allHotels}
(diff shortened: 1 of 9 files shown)


mrtnzlml:

  QR Code

package.json
```diff
@@ -16,6 +16,8 @@
   },
   "dependencies": {
     "expo": "^23.0.0",
+    "lodash": "^4.17.4",
+    "moment": "^2.19.4",
```
mrtnzlml: This is not how monorepo works. You need to update `package.json` in the "hotels" package. Yarn even warn you about this...

app/core/src/navigation/ExploreStackNavigator.js
```diff
@@ -23,6 +23,7 @@ export default StackNavigator(
   },
   {
     ...NavigatorOptions,
-    initialRouteName: 'Home',
+    // initialRouteName: 'Home',
+    initialRouteName: 'AllHotels',
```
mrtnzlml: This is a mistake.

app/hotels/src/allHotels/AllHotels.js
```diff
@@ -2,49 +2,99 @@
...
+        checkin: search.checkin,
+        checkout: search.checkout,
+        roomsConfiguration: search.roomsConfiguration,
+      },
+    });
+  };
+
+  isReadyToSearch = () => {
+    const { search } = this.state;
+    return search.checkin && search.checkout;
```
mrtnzlml: I think the location is important as well.

app/hotels/src/allHotels/AllHotels.js
```diff
@@ -2,49 +2,99 @@
...
-        force: true,
-      }}
-    />
-  );
+  render = () => {
+    const { search } = this.state;
+    return (
+      <View style={styles.wrapper}>
+        <SearchForm onChange={this._handleSearchChange} />
+        {this.isReadyToSearch() && (
```
mrtnzlml: I think some kind of default search would be better than the empty page.

app/hotels/src/allHotels/SearchForm.js
```diff
@@ -0,0 +1,134 @@
...
+    flex: 1,
+    borderRadius: 0,
+  },
+  buttonText: {
+    color: Color.grey._900,
+  },
+});
+
+type Props = {
+  onChange: (search: Search) => void,
```
mrtnzlml: I think it should be possible to use exact types here (same for the state).

app/hotels/src/allHotels/SearchForm.js
```diff
@@ -0,0 +1,134 @@
...
+    this._handleOnChange();
+  };
+
+  _handleCheckinChange = date => this._handleDateChange(date, 'checkin');
+
+  _handleCheckoutChange = date => this._handleDateChange(date, 'checkout');
+
+  _handleDateChange = async (date, type) => {
+    await this.setState({ [type]: date });
+    this._handleOnChange();
```
mrtnzlml: I don't think the `await this.setState` works. I mean it works but not as you expect. You should use second parameter of the function in this case. Something like this:

```js
this.setState({ [type]: date }, () => this._handleOnChange());
```

Method `setState` [doesn't return `Promise`](https://github.com/facebook/react/blob/master/packages/react/src/ReactBaseClasses.js#L53). Correct me if I am wrong.

app/hotels/src/allHotels/AllHotels.js
```diff
@@ -2,49 +2,99 @@
...
+              }
+            `}
+            variables={{
+              search: {
+                ...search,
+                checkin: search.checkin
+                  ? moment(search.checkin, 'DD/MM/YYYY').format('YYYY-MM-DD')
+                  : null,
+                checkout: search.checkout
+                  ? moment(search.checkout, 'DD/MM/YYYY').format('YYYY-MM-DD')
```
mrtnzlml: Do we need the second parameter here? It seems that checkin and checkout dates are actually `Date` type so we are not parsing it here...

jakubzaba: Right, it is needed, but not here.

app/hotels/src/allHotels/AllHotels.js
jakubzaba: Added default search.

mrtnzlml:
Integrated into master. Thanks.

PR #72, opened 2017-12-19
Hotels map: create "drop" marker component

### LINKED ISSUE #58: "Drop" map marker
Please create map marker with drop shape. One picture instead of long description:

screen shot 2017-12-13 at 20 24 33

I am interested in your creative solution. Possible inspiration: http://browniefed.com/blog/the-shapes-of-react-native/ 🙂 

Zeplin: https://zpl.io/bo0Gqpg

mrtnzlml:
Note: I didn't do the white circle in the middle of marker and I hope
no-one will notice... :)) (I don't know how to do it easily)

Closes #58

Checklist:

### DIFF
--- a/app/common/src/map/DropMarker.js
+++ b/app/common/src/map/DropMarker.js
@@ -1,0 +1,35 @@
+// @flow
+
+import * as React from 'react';
+import { MaterialCommunityIcons } from '@expo/vector-icons';
+
+import Color from '../Color';
+
+type Props = {|
+  size?: number,
+|};
+
+/**
+ * This drop marker is always pointing to the (0,0) coordinate. It's because
+ * the marker itself is always absolutely shifted to the left-top corner as
+ * shown on the following picture:
+ *
+ *  .-.
+ *  \O/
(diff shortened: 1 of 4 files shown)


mrtnzlml:

  QR Code

app/common/src/map/DropMarker.js
```diff
@@ -0,0 +1,38 @@
...
+ *   v
+ *   .-------.
+ *   |       |
+ *   |   x   |
+ *   |       |
+ *   `-------`
+ */
+export default function DropMarker({ size = 50 }: Props) {
+  return (
+    <View style={{ position: 'relative' }}>
```
jaroslav-kubicek: isn't `position: 'relative'` set by default in RN?

mrtnzlml: Yes, yes it is. I did it for a reason that is funny and it will never happen so I will rather remove it...

mrtnzlml: Ok, should be fixed now.

PR #75, opened 2017-12-20
Eslint: enable 'no-underscore-dangle' to unify naming

mrtnzlml:
**Proposal**: I can see that underscore may be good visual clue providing information
about method being sort of private. But I have two arguments against:

1) Underscore method is not really private.
2) All React lifecycle methods are meant to be private but they are
not prefixed by the underscore which creates inconsistency and
additional confusion.

See: https://reactjs.org/docs/react-component.html#the-component-lifecycle
See: https://eslint.org/docs/rules/no-underscore-dangle
66646d5

### DIFF
--- a/app/core/src/components/authentication/EmailLoginForm.js
+++ b/app/core/src/components/authentication/EmailLoginForm.js
@@ -24,18 +24,18 @@ export default class SimpleLoginForm extends React.Component<Props, State> {
     loading: false,
   };
 
-  _handleEmailChange = text =>
+  handleEmailChange = (text: string) =>
     this.setState({
       username: text,
     });
 
-  _handlePasswordChange = text =>
+  handlePasswordChange = (text: string) =>
     this.setState({
       password: text,
     });
 
-  _handleFormSubmit = () =>
-    this._tryLogIn(
+  handleFormSubmit = () =>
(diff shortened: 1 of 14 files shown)


mrtnzlml:

  QR Code

app/core/src/components/authentication/EmailLoginForm.js
```diff
@@ -24,18 +24,18 @@ export default class SimpleLoginForm extends React.Component<Props, State> {
...
   };
 
-  _handleEmailChange = text =>
+  handleEmailChange = (text: string) =>
     this.setState({
       username: text,
     });
 
-  _handlePasswordChange = text =>
+  handlePasswordChange = (text: string) =>
```
mrtnzlml: I didn't know about this. Flow doesn't care about underscored methods. This is not what we want right? 💩 🙂

mrtnzlml:
Also, this helped to discover an error in Flow setting (underscored methods were ignored)... 💩

jaroslav-kubicek:
there are some conflicts, cc @mrtnzlml

mrtnzlml:
> there are some conflicts, cc @mrtnzlml

Resolved.

PR #76, opened 2017-12-20
README: add info about installing dependencies

mrtnzlml:
Checklist:


mrtnzlml:

  QR Code

PR #82, opened 2017-12-21
README: improve documentation around GraphQL

mrtnzlml:
Note: GraphQL Playground doesn't work now (and we cannot upgrade). I am trying to solve it (https://github.com/graphql-cli/graphql-cli/issues/76) but I already wrote the docs so... :)

We'll see what will be the result. We can easily remove this dependency in case of additional problems. It's really easy to download introspection on our own and we can use https://kiwi-graphiql.now.sh/ instead of the Playground.


mrtnzlml:

  QR Code

PR #84, opened 2017-12-21
Extract "Stars" component into reusable one

mrtnzlml:
Checklist:


mrtnzlml:

  QR Code

app/hotels/src/allHotels/AllHotelsSearchRow.js
```diff
@@ -18,6 +19,31 @@ type Props = {|
...
+  },
+  hotelWrapper: {
+    flexGrow: 1,
+  },
+  hotelTitle: {
+    fontWeight: 'bold',
+  },
+  hotelRating: {
+    fontSize: 12,
+    color: Color.grey._600,
```
jaroslav-kubicek: not valid anymore, it's necessary to change it to $600

PR #88, opened 2017-12-21
Connect MapView with real data

jaroslav-kubicek:
Checklist:


mrtnzlml:

  QR Code

app/hotels/src/map/MapView.js
```diff
@@ -0,0 +1,133 @@
...
+  }
+`;
+
+export default createFragmentContainer(
+  MapView,
+  graphql`
+    fragment MapView on HotelAvailabilityConnection {
+      edges {
+        node {
+          ...MapViewHotel @relay(mask: false)
```
mrtnzlml: This is not exactly what I would expect here. Do we really need it?

app/hotels/src/map/MapView.js
```diff
@@ -0,0 +1,133 @@
...
+const styles = StyleSheet.create({
+  container: {
+    ...StyleSheet.absoluteFillObject,
+    justifyContent: 'flex-end',
+    alignItems: 'center',
+  },
+});
+
+class MapView extends React.Component<Props, State> {
+  constructor(props: Props) {
```
VaclavSir: This constructor is not needed:

```
class MapView extends React.Component<Props, State> {
  state = {
    selectedHotelIndex: 0,
  };
```

app/hotels/src/map/MapView.js
```diff
@@ -0,0 +1,133 @@
...
+    this.setState({ selectedHotelIndex });
+  };
+
+  onSelectMarker = (hotelId: string) => {
+    const hotels = this.getHotels();
+    const index = hotels.findIndex(
+      hotel => idx(hotel, _ => _.node.id) === hotelId,
+    );
+
+    if (Number.isFinite(index)) {
```
VaclavSir: This condition will be always true, it should be `if (index !== -1)` (when `findIndex` doesn't find anything, it returns -1 and `Number.isFinite(-1) === true`).

app/hotels/src/map/MapView.js
jaroslav-kubicek: I'm not sure about this as well, I didn't find proper way how to create nested fragments like in AllHotels, because both swipe cards & markers require multiple parts of `HotelAvailability`, but this enabled me to create `MapScreenHotel` type I can still use among components

app/hotels/src/map/MapView.js
jaroslav-kubicek: good spot, thanks!

app/hotels/src/map/MapView.js
jaroslav-kubicek: in another words, Both `MapView` & `HotelSwipeList` (and some of their descendants) rely on `MapScreenHotel` fragment => there is no single FragmentContainer component => it's still useful to define it as fragment for flow checking => requires `@relay(mask: false)`

jaroslav-kubicek: ... if there is alternative way, I'm for it, but I found this so far the best

mrtnzlml: I am sorry, I still do not understand. It seems like you are not composing fragments correctly. Why it's not possible to define the fragments in subcomponents and just use them here? Every subcomponent has different requirements and they must define their requirements in the fragment. This is the whole point of Relay.

See: https://facebook.github.io/relay/docs/en/fragment-container.html#container-composition
See: https://facebook.github.io/relay/docs/en/thinking-in-relay.html#data-masking

jaroslav-kubicek: ok, might be my misunderstanding, I'm gonna revisit it tonight ;)

app/hotels/src/map/MapScreen.js
```diff
@@ -0,0 +1,129 @@
...
+  }
+`;
+
+export default createFragmentContainer(
+  MapView,
+  graphql`
+    fragment MapScreen on HotelAvailabilityConnection {
+      edges {
+        node {
+          ...MapScreenHotel @relay(mask: false)
```
mrtnzlml: I cannot accept PR with `mask: false`. It's a red alert and it indicates inappropriate usage of Relay. It's not used in this project and it's used only 2x in BLKN (not necessary though). I tried to figure out why you need it here but it's still unclear - should be quite straightforward to use fragments in this situation. Can you please help me understand why did you decide to ignore Relay data masking here? Thanks... 🙂

mrtnzlml: Or is it still under active development? I was under impression that you are almost finishing...

mrtnzlml:

  QR Code
  `exp://exp.host/@mrtnzlml/code-review-map-view-with-real-data`

app/hotels/src/map/MapScreen.js
jaroslav-kubicek: fixed now 🙂 , cc @mrtnzlml

PR #91, opened 2017-12-22
Move tests closer to the tested code

VaclavSir:
Implementation of this proposal: https://paper.dropbox.com/doc/RFC-Unit-Tests-Co-Location-JdfMUbEGp8GH4YffeNdwN?_tk=share_copylink

Checklist:

- [x] it doesn't change anything but the files structure


mrtnzlml:

  QR Code

mrtnzlml:
I am closing this issue for now because the arguments are not convincing and we didn't clearly agree on the result:

Václav and Jaroslav think this is a good change. Main arguments are: FE is using this system and it will be more clear that you have test file so you can check it immediately. Jakub and Michal think it's more confusing and doesn't see benefits of doing this. Moreover, they think it's a small step back because the folder is going to be bloated.

I would support this decision but I need to see first that we have good reason to do this and we will benefit from it. @VaclavSir don't feel bad about this. You can try it a few months later. Maybe you'll have better arguments and you'll convince more colleagues. Thanks for understanding.

mrtnzlml:
In case we'll decide to open this PR again in the future: Metro bundler is taking into account only `__tests__` folders (https://github.com/facebook/metro/blob/master/packages/metro/src/blacklist.js#L24). There should be quite a simple solution but it's important to know...

PR #81, opened 2017-12-20
Guests popup

jakubzaba:
Closes #73

Children are not defined yet. Want to review this PR and children finish in different PR?

PR #93, opened 2017-12-22
Search: add icon for location input

### LINKED ISSUE #74: Hotels: search form icons
Icons in the search form are completely missing now. Please try to add them. All icons are already in the Expo available...

screen shot 2017-12-20 at 11 30 41

https://zpl.io/am8WYJm

jakubzaba:
Closes #74 

* Location icon added.
* There is no space for icon for datepickers, should be done with range datepicker.
* Guests icon is part of #81.

### DIFF
--- a/app/common/src/Color.js
+++ b/app/common/src/Color.js
@@ -1,10 +1,16 @@
 // @flow
 
-export default {
+const Color = {
   // brand colors
   // https://images.kiwi.com/content-media/kiwicom_brand_colours.pdf
   brand: '#0097A9',
   brandSecondary: '#0CB3C7',
+
+  icon: {
+    get grey() {
+      return Color.grey.$600;
+    },
+  },
 
   // complimentary brand colors
   buttercup: '#eb9d08',
@@ -24,3 +30,5 @@ export default {
     $900: '#212121',
   },
 };
+
+export default Color;
--- a/app/common/src/forms/TextInput.js
+++ b/app/common/src/forms/TextInput.js
@@ -1,7 +1,10 @@
 // @flow
 
 import * as React from 'react';
+import { MaterialIcons } from '@expo/vector-icons';
 import { TextInput as OriginalTextInput, View } from 'react-native';
+import defaultsDeep from 'lodash/defaultsDeep';
+import Color from '../Color';
 
 export const styles = {
   input: {
@@ -15,12 +18,16 @@ export const styles = {
     height: 40,
     backgroundColor: '#fff',
     marginBottom: 5,
+    flexDirection: 'row',
   },
 };
 
 export default function TextInput(props: Object) {
   return (
     <View style={styles.wrapper}>
+      {props.icon && (
+        <MaterialIcons {...defaultsDeep(props.icon, defaultIconProps)} />
+      )}
       <OriginalTextInput
         underlineColorAndroid="transparent"
         {...props}
@@ -31,3 +38,12 @@ export default function TextInput(props: Object) {
     </View>
   );
 }
+
+const defaultIconProps = {
+  size: 20,
+  color: Color.icon.grey,
+  style: {
+    marginLeft: 10,
+    alignSelf: 'center',
+  },
+};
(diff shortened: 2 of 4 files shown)


mrtnzlml:

  QR Code

app/common/src/Color.js
```diff
@@ -1,11 +1,17 @@
...
   // brand colors
   // https://images.kiwi.com/content-media/kiwicom_brand_colours.pdf
   brand: '#0097A9',
   brandSecondary: '#0CB3C7',
 
+  icon: {
+    get grey() {
+      return Color.grey.$600;
+    },
+  },
```
mrtnzlml: What is the idea behind? Why do you redefine grey color for the icon? Every grey icon is going to be this grey? This restriction seems a little bit harsh to me. What if I need little bit darker one? Or lighter?

I think it's quite alright just to define the grey color for `TextInput` icon (and even this may be very limiting - see below).

app/common/src/forms/TextInput.js
```diff
@@ -15,12 +18,16 @@ export const styles = {
...
     marginBottom: 5,
+    flexDirection: 'row',
   },
 };
 
 export default function TextInput(props: Object) {
   return (
     <View style={styles.wrapper}>
+      {props.icon && (
+        <MaterialIcons {...defaultsDeep(props.icon, defaultIconProps)} />
```
mrtnzlml: Now the question is: how to use different icon than from `MaterialIcons` package? 🙂 Because don't forget that you are modifying general purpose `TextInput`. There are few possibilities:

1. you cannot - meh 😖 
2. you are benevolent (`<TextInput icon={<Icon ... />}>`)
3. you introduce unified icon component and you'll detect the appropriate class by icon name

VaclavSir: With option 2 or 3 - where would you put `defaultIconProps`? These dimensions are specific for the use inside the TextInput, I don't think they belong to a general icon component.

I am OK with the option 1 for now, I would even simplify it to pass only the icon name (`<TextInput icon="location-city" />`), until we have the need for something else than grey Material icon. To switch from this simple approach to a more complicated approach will be super easy (and maybe YAGNI).

If we try to prepare for the unknown future, there is a risk that we will either overcomplicate it, or we won't address the future needs, or both.

mrtnzlml: Example of a specialized component from the shared package: https://github.com/kiwicom/react-native-app/blob/master/app/core/src/components/relay/PrivateApiRenderer.js (you don't want Redux in common but you want to reuse the logic - split it).

`<TextInput icon="location-city" />` is OK for me. Better than pure config object with unclear structure.

app/common/src/Color.js
mrtnzlml: I honestly don't think that `icon` should be here because it's not color - should be in the `Icon` component. I removed `TextInput` from this file before - for the same reason. It's not very good design (better to reuse one component and not copy-paste code snippet with a default setting every time). So my idea is that theoretically, you need to use this color only once. But since I have a hard time to convince others, here is an idea how to avoid getters and related boilerplate:

```
const ColorPalette = {
  grey: {
    $100: '#f5f5f5',
    // ...
    $900: '#212121',
  },
};

export default {
  brand: '#0097A9',
  brandSecondary: '#0CB3C7',

  icon: {
    grey: ColorPalette.grey.$600,
  },

  ...ColorPalette,
};
```

cc @VaclavSir 
Related comments: https://github.com/kiwicom/react-native-app/pull/81#discussion_r158594068

VaclavSir: I would use `Color.grey.$600` directly in the TextInput style, I don't see the benefit of this `Color.icon.grey`. But I agree at least this is better than the getter approach.

mrtnzlml: I think that the idea was really to have unified grey level. Because without `Color.icon.grey` everyone must use exactly `$600` to keep it unified and future changes to all icons are going to be pain (like we need darker grey 🙄 ). So to prevent these stupid mistakes you need to have `Color.grey` or wrap it (or be really careful).

PR #98, opened 2017-12-27
Show an error when all hotels query fails

VaclavSir:
Checklist:

### DIFF
--- a/app/hotels/src/allHotels/AllHotelsSearchList.js
+++ b/app/hotels/src/allHotels/AllHotelsSearchList.js
@@ -3,7 +3,8 @@
 import * as React from 'react';
 import idx from 'idx';
 import { createFragmentContainer, graphql } from 'react-relay';
-import { ScrollView } from 'react-native';
+import { ScrollView, Text } from 'react-native';
+import { CenteredView } from '@kiwicom/react-native-app-common';
 
 import AllHotelsSearchRow from './AllHotelsSearchRow';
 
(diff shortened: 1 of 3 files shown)


mrtnzlml:

  QR Code
  `exp://exp.host/@mrtnzlml/code-review-allhotels-error`

app/hotels/src/allHotels/AllHotelsSearchList.js
```diff
@@ -15,8 +16,8 @@ type Props = {|
 |};
 
 function AllHotelsSearchList(combinedProps: Props) {
-  const hotels = idx(combinedProps, _ => _.data.edges) || []; // TODO: render 404 page instead of nothing
```
mrtnzlml: Haha. Fuck.

VaclavSir: 😄  Maybe it's about time to add this ESLint rule: https://eslint.org/docs/2.0.0/rules/no-warning-comments

app/hotels/src/allHotels/AllHotelsSearchList.js
```diff
@@ -31,6 +32,8 @@ function AllHotelsSearchList(combinedProps: Props) {
         }
       })}
     </ScrollView>
+  ) : (
+    <GeneralError errorMessage="Server error, please try again later" />
```
mrtnzlml: You sure? Is it server error? What if there are no edges? This is the most problematic part and one of the reasons why I don't know how to fix it properly...

What about instead of writing "server error" something like "it was not possible to load/find" or "no hotels found".

app/hotels/src/allHotels/AllHotelsSearchList.js
mrtnzlml: TODOs are OK, it's not finished yet. I just could do it properly here. But for example, Jakub is still working on search and instead of empty callback it's better to add TODO: https://github.com/kiwicom/react-native-app/blob/7fe19451ae28b4578ab3f98c6dd9da1ee9ff56eb/app/hotels/src/allHotels/searchForm/Guests.js#L52

We just need to fix them or delete at the end...

app/hotels/src/allHotels/AllHotelsSearchList.js
VaclavSir: I would expect that the API returns an empty array if no hotel is found and `!![] === true`.

mrtnzlml: Ok, this is even worse then. If it returns the empty array the result will be empty screen without the error message, right?

VaclavSir: You're right, we need to handle both.

mrtnzlml: Try to think about reusability please, because there are so many places where we need it. Maybe something like "error boundaries" component? (inspired by React itself but with completely different functionality)...

mrtnzlml:
Please deal only with 404 situations. I will send PR with partial errors handling soon (it's not possible to handle it during render phase - it's too late).

PR #99, opened 2017-12-28
Handle partial GraphQL errors

mrtnzlml:
Simple workaround for Relay bug (see: https://github.com/facebook/relay/issues/1913).
I have no idea how to write tests for this part of application.
I would prefer to merge this and think about it later since we will
need to refactor it a little bit because of other tests...

Related issue: #98

img_1398

Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,8 @@ Try it in [Expo](https://expo.io/):
 * [Project structure](#project-structure)
 * [Working with GraphQL API](#working-with-graphql-api)
 * [Offline first](#offline-first)
-* [Best Practices](#best-practices)
+* [Error handling](#error-handling)
+* [Best practices](#best-practices)
   * [Accessing arbitrarily nested, possibly nullable properties on a JavaScript object](#accessing-arbitrarily-nested-possibly-nullable-properties-on-a-javascript-object)
 * [Known issues](#known-issues)
 * [Troubleshooting](#troubleshooting)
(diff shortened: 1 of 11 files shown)


mrtnzlml:

  QR Code
  `exp://exp.host/@mrtnzlml/code-review-handle-partial-errors`

README.md
```diff
@@ -11,7 +11,8 @@ Try it in [Expo](https://expo.io/):
 * [Project structure](#project-structure)
 * [Working with GraphQL API](#working-with-graphql-api)
 * [Offline first](#offline-first)
-* [Best Practices](#best-practices)
+* [Error handling](#error-handling)
+* [Best practices](#best-practices)
```
VaclavSir: Capital H, capital P.

README.md
```diff
@@ -114,7 +115,25 @@ this.setState({ refreshing: true }, () => {
...
-## Best Practices
+## Error handling
+
+Error handling is complicated in general - especially in GraphQL environment. There are several scenarios that may occur:
+
+1. GraphQL API returns `data` field and no `errors`
+
+This should be considered as a valid full response and there should not be any errors. There may be nullable fields, however.
+
+2. GraphQL API returns `data = null` and `errors` field
```
VaclavSir: I'm trying to test this scenario by modifying the fetchQuery:

```
  const fetchQuery = async (operation, variables, cacheConfig) => {
    return {
      data: null,
      errors: [
        {
          message: 'Some message',
        },
      ],
    };
```

Then I get the following message, is this the expected behavior? Or am I testing it wrong (and how can I test it correctly)?

simulator screen shot - iphone 5s - 2017-12-28 at 15 21 40

README.md
mrtnzlml: Would you prefer first capitals everywhere? I did this change intentionally but I also prefer first capitals everywhere in English headings...

README.md
mrtnzlml: Yes, this is correct behavior. Data cannot be `null` because this indicates a fatal GraphQL error and not partial error. See documentation. This is the only situation when Relay actually works correctly and handles errors...

mrtnzlml:
> I also noticed that we create a new relay environment for every call of commitMutation and in the render method of QueryRenderer. Isn't the environment supposed to be a single instance for the whole application?

Sounds like a mistake but that's not relevant here. I should fix it though. Probably during refactoring of this section, not now. You can create a new issue...

README.md
mrtnzlml: Ah, you are pointing to the docs. So yes, this is `GeneralError` component in action.

README.md
VaclavSir: Now that I tried to find some arguments, I'm not so sure anymore. There are two main standards - *title case* and *sentence case*. Until now I thought the title case is more "correct", but apparently the sentence case is also a valid option. Maybe we can skip this decision for now and continue using the sentence case.

https://en.wikipedia.org/wiki/Letter_case#Headings_and_publication_titles

VaclavSir:
Issue created #101

PR #102, opened 2017-12-29
Use 'eslint-plugin-import' rules to stop the import mess

mrtnzlml:
Note: we do not have a rule for importing Flow types (yet).

Checklist:


mrtnzlml:

  QR Code
  `exp://exp.host/@mrtnzlml/code-review-eslint-plugin-import`

PR #114, opened 2018-01-02
Do not animate when map is rendered & show better region area

jaroslav-kubicek:
---


mrtnzlml:

  QR Code
  `exp://exp.host/@mrtnzlml/code-review-improve-initial-region`

jaroslav-kubicek:
I should ship some tests through...

mrtnzlml:
Now it behaves really well. 👍 We can do small tweaks later if needed... 🙂 

> I should ship some tests through...

Please do. I am interested basically only in `getDelta` method. I would also expect a small JSDoc comment about how does it actually work.

PR #117, opened 2018-01-03
Remove 'graphql-cli' dependency

mrtnzlml:
We need to keep one version of GraphQL but this dependency strictly
requires version '0.12.3'. As a result it downloads GraphQL snapshot
that is incompatible with Relay.

See: https://github.com/graphql-cli/graphql-cli/issues/76 (original
issue is resolved but it's still broken)

---


mrtnzlml:

  QR Code
  `exp://exp.host/@mrtnzlml/code-review-delete-graphql-cli`

PR #121, opened 2018-01-03
Common: add Slider component

jakubzaba:
**You can try a slider in #122**

Known issues
* ~~it should detect device orientation changed and recalculate slider length (use `react-native-orientation` or store screen dimensions in redux and check on `mapStateToProps`)~~
* snapped (jump by steps) is broken in lib, but it's not needed right now (see: https://github.com/ptomasroos/react-native-multi-slider/issues/49)

---

### DIFF
--- a/app/common/index.js
+++ b/app/common/index.js
@@ -20,6 +20,7 @@ export { default as PartialFailure } from './src/errors/PartialFailure';
 export { default as DatePicker } from './src/forms/DatePicker';
 export { default as TextInput } from './src/forms/TextInput';
 export { default as NumberControl } from './src/forms/NumberControl';
+export { default as Slider } from './src/forms/Slider';
 
 export { default as NetworkImage } from './src/image/NetworkImage';
 
--- a/app/common/package.json
+++ b/app/common/package.json
@@ -4,6 +4,7 @@
   "private": false,
   "main": "index.js",
   "dependencies": {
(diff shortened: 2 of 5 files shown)


mrtnzlml:

  QR Code
  `exp://exp.host/@mrtnzlml/code-review-slider`

mrtnzlml:
> it should detect device orientation changed and recalculate slider length (use react-native-orientation or store screen dimensions in redux and check on mapStateToProps)

Yes, please fix it. `Dimensions` module is good enough. This should help you quite significantly: https://github.com/kiwicom/react-native-app/blob/285b6533de8318a50336cbf6342e764207c7c2f8/app/hotels/src/map/HotelSwipeList.js#L51-L57

> snapped (jump by steps) is broken in lib, but it's not needed right now

Can you please link the issue number here?

jakubzaba:
Thanks a lot, `Dimensions` works great.
Here is a issue https://github.com/ptomasroos/react-native-multi-slider/issues/49. I'll fix it, but it's not necessary for us right now.

mrtnzlml:
Unfortunately, `Dimensions` works badly because it calculates width based on the window. That's wrong. I am trying to figure out a better solution. Btw - if it doesn't work please remove it from props so no one is tempted to use it (+ add a comment).

mrtnzlml:
@jakubzaba Thank you very much - especially for opening the issue and sending PR... 👍

PR #122, opened 2018-01-03
Price filter design

jakubzaba:
---

### DIFF
--- a/app/hotels/src/filter/FilterStripe.js
+++ b/app/hotels/src/filter/FilterStripe.js
@@ -4,6 +4,7 @@ import * as React from 'react';
 import { ScrollView, View } from 'react-native';
 
 import StarsFilter from './stars/StarsFilter';
+import PriceFilter from './price/PriceFilter';
 
 const styles = {
   view: {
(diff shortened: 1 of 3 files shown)


mrtnzlml:

  QR Code
  `exp://exp.host/@mrtnzlml/code-review-price-filter-design`

app/hotels/src/filter/FilterStripe.js
```diff
@@ -30,6 +31,14 @@ export default class FilterStripe extends React.Component<{||}> {
           showsHorizontalScrollIndicator={false}
         >
           <StarsFilter onChange={this.handleChange} />
+          <PriceFilter
+            onChange={this.handleChange}
+            min={18}
+            max={365}
+            start={18}
+            end={365}
```
mrtnzlml: What is the story behind these limits? 🙂 Why is for example minimum 18 euro? I expect that you want to replace it during connection phase to the API. But the limit values are still unclear because we don't know what are min and max price of all hotels in the region. They did not think this through. Can you please ask someone via Slack? Or do you know the answer already?

jakubzaba: It's just dummy data, no story behind it :) This PR is just about design / functional components.

PR #123, opened 2018-01-04
Unit test for MapView region calculation added

jaroslav-kubicek:
---


mrtnzlml:

  QR Code
  `exp://exp.host/@mrtnzlml/code-review-unit-test-map-view-region`

app/hotels/src/map/MapView.js
```diff
@@ -50,7 +50,7 @@ class Map extends React.Component<Props, State> {
     }
   };
 
-  getCoordinate = (hotel: Object) => {
+  getCoordinate = (hotel: Object): LatLng | null => {
```
mrtnzlml: Nice, but completely pointless. It's actually way worse than before. You basically wrote here this:

```js
(hotel: Object): any | null
```

So I can easily write this and it's OK:

```js
return {
  foo: coordinate.lat,
  bar: coordinate.lng,
};
```

I am sorry, I did not notice before. The mistake is here:

```js
import { MapView, type Region, type LatLng } from 'expo';
```

Type `LatLng` does not exist (probably the same with `Region`). Did you also make this up? This is a new trend in RN repo - you are not the first one... 😄

Honestly, I don't know how to prevent this. 😢

app/hotels/src/map/__tests__/MapView.test.js
```diff
@@ -0,0 +1,97 @@
+// @flow
+
+import { Map } from './../MapView';
+
+const oldTown = {
+  node: {
+    id: '1',
+    price: {},
```
mrtnzlml: You could be a little bit more DRY here... 🙂

mrtnzlml:
Integrated, thanks.

PR #125, opened 2018-01-04
Fix build on .circleci

mrtnzlml:
CircleCI is more powerful for our needs but the scripts are not ready yet. So, for now, it will run in parallel with Travis-CI until we fix all related issues.

---


mrtnzlml:

  QR Code
  `exp://exp.host/@mrtnzlml/code-review-circleci`

PR #132, opened 2018-01-05
Search by cityId based on user provided location name

### LINKED ISSUE #119: Connect hotels search form to the API
We need to make it work. Should be quite self-explanatory.

jaroslav-kubicek:
This closes #119.

---

### DIFF
--- a/app/hotels/src/HotelsReducer.js
+++ b/app/hotels/src/HotelsReducer.js
@@ -1,8 +1,6 @@
 // @flow
 
-export type HotelsSearchParametersType = {|
-  latitude: number | null,
-  longitude: number | null,
+export type HotelsSearchParametersType = {
   checkin: Date | null,
   checkout: Date | null,
   roomsConfiguration: {|
@@ -11,21 +9,23 @@ export type HotelsSearchParametersType = {|
       age: number,
     |}>,
   |},
-|};
+};
 
 export type HotelsReducerState = {
+  cityId: string | null,
+  location: string,
   searchParams: HotelsSearchParametersType,
 };
 
-export type HotelsReducerActions = {|
-  type: 'setSearchFilters',
-  filter: HotelsSearchParametersType,
-|};
+export type HotelsReducerActions =
+  | {| type: 'setSearchFilters', filter: HotelsSearchParametersType |}
+  | {| type: 'setLocation', location: string |}
+  | {| type: 'setCityId', cityId: string | null |};
 
 const InitialHotelsState: HotelsReducerState = {
+  cityId: null,
+  location: 'Prague',
   searchParams: {
-    latitude: 50.08,
-    longitude: 14.44,
     checkin: null,
     checkout: null,
     roomsConfiguration: {
@@ -43,7 +43,20 @@ export default (
     case 'setSearchFilters':
       return {
         ...state,
-        searchParams: action.filter,
+        searchParams: {
+          ...state.searchParams,
+          ...action.filter,
+        },
+      };
+    case 'setLocation':
+      return {
+        ...state,
+        location: action.location,
+      };
+    case 'setCityId':
+      return {
+        ...state,
+        cityId: action.cityId,
       };
     default:
       return state;
--- a/app/hotels/src/allHotels/AllHotels.js
+++ b/app/hotels/src/allHotels/AllHotels.js
@@ -2,35 +2,41 @@
 
 import * as React from 'react';
 import { graphql } from 'react-relay';
-import { PublicApiRenderer } from '@kiwicom/react-native-app-relay';
-import { Layout } from '@kiwicom/react-native-app-common';
+import { SimpleQueryRenderer } from '@kiwicom/react-native-app-relay';
+import moment from 'moment/moment';
+import idx from 'idx';
 
-import SearchForm from './searchForm/SearchForm';
-import FilterStripe from '../filter/FilterStripe';
-import AllHotelsSearchList from './AllHotelsSearchList';
-import type { AllHotelsSearchQueryResponse } from './__generated__/AllHotelsSearchQuery.graphql';
-import type { SearchParametersType } from './searchForm/SearchParametersType';
+import AllHotelsSearch from './AllHotelsSearch';
+import type {
+  SearchParametersType,
+  OnChangeSearchParams,
+} from './searchForm/SearchParametersType';
 import type { AvailableHotelSearchInput } from '../singleHotel';
 
 type Props = {|
+  location: string,
   search: SearchParametersType,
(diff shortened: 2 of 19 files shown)


mrtnzlml:

  QR Code
  `exp://exp.host/@mrtnzlml/code-review-119-connect-to-real-data`

app/core/src/services/redux/reducers/HotelsReducer.js
```diff
@@ -5,9 +5,10 @@ import type { ReduxState, ReduxActions } from '../../../types/Redux';
 type HotelsState = $PropertyType<ReduxState, 'hotels'>;
 
 const InitialHotelsState: HotelsState = {
+  location: '',
   searchParams: {
-    latitude: 50.08,
-    longitude: 14.44,
+    latitude: null,
+    longitude: null,
```
mrtnzlml: We can now change it to the city ID.

app/hotels/src/allHotels/AllHotels.js
```diff
@@ -17,53 +15,69 @@ type Props = {|
...
-  isReadyToSearch = (): boolean => {
-    const { search: s } = this.props;
+  renderInnerComponent = rendererProps => {
+    const {
+      onFilterChange,
+      openSingleHotel,
+      search,
+      location,
+      onLocationChange,
+    } = this.props;
```
mrtnzlml: Note: I am not exactly super big fan of destructions like this because even simple things like `const p = this.props` will save so many characters... 😊

app/hotels/src/allHotels/AllHotels.js
```diff
@@ -17,53 +15,69 @@ type Props = {|
...
-              search: search,
-            }}
-            render={this.renderInnerComponent}
-            cacheConfig={{
-              force: true,
-            }}
-          />
-        )}
-      </Layout>
+      <SimpleQueryRenderer
```
mrtnzlml: Oh, now I understand. I was confused by the `simple` name. So it's like a "background" query renderer? It should be used only when you need to fetch something at the background (location in this case) and you don't want to show all the unnecessary spinners and errors, right?

So I would add it to the `AllHotelsSearch` or maybe directly to the `SearchForm`. It seems weird that you need to introduce two additional props just to use this query renderer in the parent component and send the data back down. It's very confusing.

Also, does it make sense to introduce new query like this?

```graphql
{
  allHotels: allAvailableHotelsInCity(city, checkin, checkout, roomsConfiguration) {
    ...AllHotelsSearchList
  }
}
```

It would be even nicer this way because you don't need to send the city ID...

app/hotels/src/allHotels/AllHotels.js
```diff
@@ -17,53 +15,69 @@ type Props = {|
...
+              ...AllHotelsSearch
+            }
+          }
+        `}
+        render={this.renderInnerComponent}
+        variables={{
+          prefix: location,
+        }}
+        cacheConfig={{
+          force: true,
```
mrtnzlml: We can probably save the result of this query into the offline cache. Does it make sense?

app/hotels/src/allHotels/AllHotelsSearch.js
```diff
@@ -0,0 +1,92 @@
...
+      <Layout>
+        <SearchForm
+          onChange={this.handleSearchChange}
+          onLocationChange={onLocationChange}
+          search={search}
+          location={location}
+          data={null}
+        />
+        <FilterStripe />
+        {this.isReadyToSearch() && (
```
mrtnzlml: I know that you didn't introduce it here - but this is stupid check. The search results may disappear only because you set `checkin > checkout` for a while. This implementation is 💩 ...

app/hotels/src/allHotels/searchForm/SearchForm.js
```diff
@@ -108,24 +75,26 @@ export default class SearchForm extends React.Component<Props, State> {
         <View style={styles.row}>
           <DatePicker
             placeholder="Start date"
-            format={DATE_FORMAT}
-            date={this.state.checkin}
+            format={DISPLAY_DATE_FORMAT}
+            date={moment(search.checkin, DATA_DATE_FORMAT).toDate()}
```
mrtnzlml: I think `DatePicker` should accept `Date` and do necessary formatting on its own. This interface is hard to (re)use. Also - the underlying `react-native-datepicker` accepts `Date` objects so we can improve it this way...

app/hotels/src/allHotels/AllHotels.js
jaroslav-kubicek: I can imagine that

The only possible drawback  scenario I can imagine is to show something like "you entered silly city name, try again something different"

app/hotels/src/HotelsReducer.js
```diff
@@ -1,8 +1,6 @@
 // @flow
 
-export type HotelsSearchParametersType = {|
-  latitude: number | null,
-  longitude: number | null,
```
mrtnzlml: You were supposed to replace just these coordinates with city ID. Why is it so different now? Parameters to search hotels (`HotelsSearchParameters(Type)`) are: city ID, checkin, checkout, room configuration. This change makes the whole picture even more complicated. Can you please elaborate?

app/hotels/src/allHotels/AllHotels.js
mrtnzlml: Now it behaves totally weird, no loading indicator (for quite a while to be honest), blank page (I tried "Pilsen" and it's blank forever - is it an error, still loading?), to short city = no feedback. Make it work, please.

jaroslav-kubicek: This issue should be solved now

app/hotels/src/HotelsReducer.js
jaroslav-kubicek: Probably my fault by introducing QueryRenderer. I also removed completely local state away from SearchForm as it was redundant and now all settings are here.

jaroslav-kubicek:
Look on it now if you have second, @mrtnzlml , and help me decide whether it's feasible or 💩

app/hotels/src/map/AllHotelsMap.js
```diff
@@ -20,17 +21,22 @@ class AllHotelsMap extends React.Component<Props> {
...
 
   render = () => {
+    const { cityId, search } = this.props;
+
     return (
       <PublicApiRenderer
         query={graphql`
           query AllHotelsMapQuery($search: HotelsSearchInput!) {
-            allAvailableHotels(search: $search) {
+            allAvailableHotels: allAvailableHotels(search: $search) {
```
mrtnzlml: Nice alias.

mrtnzlml:
Merged.

PR #141, opened 2018-01-09
Introduce new Icon component

mrtnzlml:
This wraps `@expo/vector-icons` package and it will help us to do future
 changes easily. We currently support only icons from material design
 because these are the icons used in the design (only).

 Closes: https://github.com/kiwicom/react-native-app/issues/94

---


mrtnzlml:
Also, since we are detaching - we will update only this component. Easy... :)

PR #142, opened 2018-01-09
Fix script for sending link to expo app to discussion related to PR

jaroslav-kubicek:
---

### DIFF
--- a/scripts/expo/commands/publishCodeReview.js
+++ b/scripts/expo/commands/publishCodeReview.js
@@ -48,7 +48,7 @@ module.exports = () => {
     }),
   });
 
-  const githubPullRequestId = env.CI_PULL_REQUEST_ID;
+  const githubPullRequestId = env.CIRCLE_PR_NUMBER;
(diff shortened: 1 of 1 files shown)


scripts/expo/commands/publishCodeReview.js
```diff
@@ -48,14 +48,16 @@ module.exports = () => {
     }),
   });
 
-  const githubPullRequestId = env.CI_PULL_REQUEST_ID;
```
mrtnzlml: This is not OK - you need to use values from this file: https://github.com/kiwicom/react-native-app/blob/master/scripts/expo/environment.js (`env.*` is not `process.env` here).

mrtnzlml:
@jaroslav-kubicek Seems like it's really broken, but I would recommend to ignore it because we are removing Expo...

scripts/expo/commands/publishCodeReview.js
jaroslav-kubicek: I know, just trying... 🙂

jaroslav-kubicek:
💔

PR #146, opened 2018-01-09
Copy app folder to rnApp and make it run

tbergq:
---


mrtnzlml:
A little bit of misunderstanding here. I expected this:

> Move `app` into this new folder and move everything two levels up (to get rid of `rnApp/reactNativeApp`).

So at the end, there is going to be this structure:

```
.
├── app/
├── ios/
└── android/
```

Also please make sure that we are actually using `react-native` from `node_modules` so the correct `ios` command should be:

```diff
- "ios": "react-native run-ios",
+ "ios": "yarn react-native run-ios",
```

mrtnzlml:
My proposed migration is here: #150 
Can you please check it?

Note - icons doesn't work yet. We will do it in different PR so we can see the diff.

tbergq:
Yes, I think we misunderstood, I will close this MR

PR #151, opened 2018-01-10
Add react-native-vector-icons to icon component

tbergq:
---


mrtnzlml:
Shipped, thanks.

Issue #154, opened 2018-01-10
Improvements in Map View


jaroslav-kubicek:
- add missing navigation to a single hotel
- add "handle line" above hotel card to make it clear it can be swiped up
- fix overlapping markers

PR #164, opened 2018-01-11
Enable navigating to single hotel detail from map view

jaroslav-kubicek:
References #154

---


mrtnzlml:
Can you please rebase it? It's probably because we moved the hotel screens to the new location.

jaroslav-kubicek:
rebased ✅  cc @mrtnzlml

PR #166, opened 2018-01-11
Fix issue with overlapping map markers

jaroslav-kubicek:
🎉 🍾 🎆 
---

### DIFF
--- a/app/hotels/src/map/MapView.js
+++ b/app/hotels/src/map/MapView.js
@@ -45,18 +45,27 @@ const styles = StyleSheet.create({
 
 export class Map extends React.Component<Props, State> {
   map: typeof MapView | null;
+  markers: {| number: React.Ref<MapView.Marker> |};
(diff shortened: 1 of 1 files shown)


app/hotels/src/map/MapView.js
```diff
@@ -45,18 +45,27 @@ const styles = StyleSheet.create({
 
 export class Map extends React.Component<Props, State> {
   map: typeof MapView | null;
+  markers: {| number: React.Ref<MapView.Marker> |};
```
mrtnzlml: It's amazing how big is the difference between Flow enabled and Flow disabled. Flow was unfortunately disabled in this case. Should've been `{ [number]: type }`... :)

BTW - I have no idea how to type it correctly. It's happy all the time. Not sure about `typeof` (probably not needed) and it doesn't care about the `| null` even though it can be null...

PR #167, opened 2018-01-11
Hotels: expose only one standalone package

mrtnzlml:
Hopefully, there is no longer direct usage of core in hotels or hotels
in the core. It should be therefore standalone (not completely verified
yet).

---


app/hotels/src/navigation/NavigationStack.js
```diff
@@ -0,0 +1,139 @@
+// @flow
```
mrtnzlml: This file would like to be refactored as well (not now).

app/redux/src/ReduxContext.js
```diff
@@ -0,0 +1,62 @@
...
+  reducers: { [string]: Reducer },
+|};
+
+/**
+ * Creates Redux context for underlying components. It makes only sense if
+ * you pass additional reducers to be registered.
+ */
+export default class ReduxContext extends React.Component<Props> {
+  componentWillMount = () => {
+    // it's too late to register new reducers in componentDidMount
```
jaroslav-kubicek: I'm not sure if I comprehend this comment, why it's then done in this place? :)

mrtnzlml: Because `componentWillMount` is called before `componentDidMount` (too late). I could add it also to the constructor...

You can maybe delete this comment if it's not necessary.

jaroslav-kubicek: dammit, I saw `componentDidMount` instead `componentWillMount`, that's why I was confused 🤕

PR #168, opened 2018-01-11
Prepare example component with native implementation

tbergq:
---


app/examples/native-module/README.md
```diff
@@ -0,0 +1,16 @@
+## Native modules example
```
mrtnzlml: Please add a brief summary of what is this example all about and how is it useful.

mrtnzlml:
I have no idea what is happening here. 🙂 Integrated into master...

PR #170, opened 2018-01-11
Fix date range

### LINKED ISSUE #109: Adjust "to" date when "From" is changed to later
When user sets "From" date to later than "To", a new request for list of hotels is immediately dispatched, which fails because interval "from-to" doesn't make any sense, e.g. "From Feb 2 to January 9".

jakubzaba:
Closes #109 
- [x] disable past dates
- [x] show message when you pick check-out before check-in
- [x] show message when you pick check-out more than 30 days after check-in

---

### DIFF
--- a/app/common/index.js
+++ b/app/common/index.js
@@ -15,6 +15,7 @@ export { default as Date } from './src/datetime/Date';
 export { default as DateTime } from './src/datetime/DateTime';
 
 export { default as GeneralError } from './src/errors/GeneralError';
+export { default as ErrorMessage } from './src/errors/ErrorMessage';
 export { default as PartialFailure } from './src/errors/PartialFailure';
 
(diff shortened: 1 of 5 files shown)


mrtnzlml:
@jaroslav-kubicek Are you satisfied with this PR? 🙂

jaroslav-kubicek:
Sorry, I put a comment to issue. My concern is behaviour:

>What about simply adjusting the other date, just like in kiwi.com native app or Booking.com, e.g. If I set "To" to earlier date than "From", let's say January 25, "From" is automatically updated as well to "January 24"? UX should be much better with this.

mrtnzlml:
Ah, I would prefer to fix this bug first and improve the behavior later. But if it's your concern - can you please ask Martin K. if this is the correct behavior? I don't know how this should behave...

PR #176, opened 2018-01-12
Print a warning when I run yarn start

VaclavSir:
Took me some time to remember that we don't `yarn start` anymore.

---


mrtnzlml:
No, this starts packager. The behavior is exactly the same and correct. It worked the same even before except you got QR code (this is the only change in fact). Command `yarn ios` worked also the same way and I am honestly surprised you didn't use it... 🙂

PR #182, opened 2018-01-15
Set the maximum limit of markers displayed in initial region

### LINKED ISSUE #177: Hotels map: limit search results
Show only first 15 hotels and zoom closer to the price markers so they are not so crowded.

jaroslav-kubicek:
This closes #177 

---

### DIFF
--- a/app/hotels/src/map/MapView.js
+++ b/app/hotels/src/map/MapView.js
@@ -35,9 +35,11 @@ type MarkerPressEvent = {|
 |};
 
 type LatLng = {|
-  latitude: number,
   longitude: number,
+  latitude: number,
 |};
+
+const NO_OF_MARKERS_IN_REGION = 15;
 
 const styles = StyleSheet.create({
   map: StyleSheet.absoluteFillObject,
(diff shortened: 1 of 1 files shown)


app/hotels/src/map/MapView.js
```diff
@@ -4,7 +4,7 @@ import idx from 'idx';
 import * as React from 'react';
 import { createFragmentContainer, graphql } from 'react-relay';
 import { StyleSheet } from 'react-native';
-import MapView from 'react-native-maps';
+import MapView, { type LatLng } from 'react-native-maps';
```
mrtnzlml: I already fixed this once. This type does **not** exist. And it will be cast as an `any` type when you try to import non-existent types. And there is probably nothing worse than `any` in Flow... 🙂

app/hotels/src/map/MapView.js
```diff
@@ -34,10 +34,7 @@ type MarkerPressEvent = {|
   },
 |};
 
-type LatLng = {|
-  latitude: number,
-  longitude: number,
-|};
+const NO_OF_MARKERS_IN_REGION = 15;
```
mrtnzlml: It still looks very crowded. What about fetching just the first 15? Airbnb has a similar approach (+lazy fetching of additional hotels - we don't need this now).

app/hotels/src/map/MapView.js
jaroslav-kubicek: dammit 💩

PR #183, opened 2018-01-15
Add hotels package to native app

tbergq:
---


mrtnzlml:
I don't like this. Can you just create a new temporary repo? It would be much cleaner. Also, you cannot use `http://localhost:8081/index.bundle` URL. It must be already bundled. This will help us to create and distribute the package properly (via NPM)...

tbergq:
Sure, I thought you wanted it the same repo, I will create a separate repo.

mrtnzlml:
Well, it doesn't really work if you need to create `ios2/` and `android2/`. In this case, it doesn't make sense anymore.

PR #184, opened 2018-01-16
Limit the width of the swipe list

### LINKED ISSUE #174: Optimize map view for tablets
It's basically only about the swiper at the bottom.

See: https://zpl.io/2vnRjdj

VaclavSir:
Close #174 
screen shot 2018-01-16 at 11 38 36

---

### DIFF
--- a/app/hotels/src/map/HotelSwipeList.js
+++ b/app/hotels/src/map/HotelSwipeList.js
@@ -27,9 +27,16 @@ type State = {|
   open: boolean,
 |};
 
+const SNAP_WIDTH = 0.8;
+const OPEN_HEIGHT = 150;
+const CLOSED_HEIGHT = 80;
+const MAX_WIDTH = 668;
+
 const styles = StyleSheet.create({
   container: {
     backgroundColor: '#fff',
+    maxWidth: MAX_WIDTH,
+    overflow: 'hidden',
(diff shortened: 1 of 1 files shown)


VaclavSir:
I fixed the visibility of the next hotel. I don't quite understand the note about `getLandscapeThreshold`, do you mean the maxWidth should be applied only in landscape mode?

screen shot 2018-01-16 at 16 56 49

mrtnzlml:
Landscape threshold essentially solves the question "Why did you choose _668_? Why not _686_? Is it OK for every tablet?". Everything below this threshold is OK for portrait mode and everything above is OK for landscape mode. The name is stupid - feel free to make it less stupid.

This is basically how `Layout` works - it's wide open in portrait mode and it keeps its width for large devices. In the future, we can improve this method significantly by adding devices resolutions map or something like that - not important now.

I just don't like this next magic number constant (even the magic constant in `Layout` component is wrong).

mrtnzlml:
Please fix the unhappy CI... 🙂

VaclavSir:
I didn't chose 668, I copied it from Zeplin... I comforted the CI, now it's happy. 😸

mrtnzlml:
Zeplin is not really designed for every device... :)

Issue #186, opened 2018-01-16
Add animation when showing hotel address


jaroslav-kubicek:
When user swipes up in map view, an address is shown immediately. Use animation to give user more appropriate feedback.

PR #188, opened 2018-01-16
Use animation when showing/hiding address in map view

jaroslav-kubicek:
Closes: #186

---

### DIFF
--- a/app/common/src/view/VerticalSwipeResponder.js
+++ b/app/common/src/view/VerticalSwipeResponder.js
@@ -1,20 +1,28 @@
 // @flow
 
 import * as React from 'react';
-import { View, PanResponder } from 'react-native';
+import { Animated, PanResponder } from 'react-native';
 
 import type { GestureState, PanResponderEvent } from '../../types/Events';
 
 type SwipeConfig = {|
   velocityThreshold: number,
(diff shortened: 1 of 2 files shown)


mrtnzlml:
I don't think this is how it should behave. It's weird that I can stop it in the middle of the transition:

screen shot 2018-01-17 at 10 16 57

But I don't know. This is what I would expect (see for example Google Maps).

jaroslav-kubicek:
ok, good point, I will make it animate to the end

jaroslav-kubicek:
fixed

app/common/src/view/VerticalSwipeResponder.js
```diff
@@ -70,6 +79,7 @@ class VerticalSwipeResponder extends React.Component<Props> {
       //stop JS beautify collapse
       onStartShouldSetPanResponder: shouldSetResponder,
       onMoveShouldSetPanResponder: shouldSetResponder,
+      onPanResponderMove: this.handlePanResponderMove.bind(this),
```
mrtnzlml: Hm 🙈, I am leaving the old-way binding here only because of consistency in this file but it would be nice to fulfil our todos:

> FIXME: **refactor** by throwing away all unnecessary code

🙂

PR #191, opened 2018-01-17
ReduxContext: remove 'componentWillMount' method

mrtnzlml:
Methods 'componentWill*' are deprecated in next React version.
See: https://github.com/bvaughn/rfcs/blob/static-lifecycle-methods/text/0000-static-lifecycle-methods.md

---


mrtnzlml:
@jaroslav-kubicek CI is happy now - CircleCI had an outage.

PR #194, opened 2018-01-17
Optimize search results for tablets

### LINKED ISSUE #175: Optimize hotel search results for tablets
See: https://zpl.io/bWQMGvp

jakubzaba:
Closes #175 

There is one issue. When device is rotated and `onLayout` is called, it re-render both `AllHotels` and `AllHotelsMap` no matter search attributes are the same.

---

### DIFF
--- a/app/common/src/Device.js
+++ b/app/common/src/Device.js
@@ -3,8 +3,17 @@
 import { Dimensions } from 'react-native';
 
 function renewDimensions() {
-  return Dimensions.get('screen');
+  // Tests would throw error for not set 'screen'. Catch the error and return default values.
+  // See https://github.com/facebook/react-native/blob/0cd69e8a02d254577fac9ccd3ff8a1727a024cc8/Libraries/Utilities/Dimensions.js#L93
+  try {
+    return Dimensions.get('screen');
+  } catch (e) {
+    return { width: 0, height: 0 };
+  }
 }
+
+// Tablets have often dimensions ratio 3:2-4:3, phones have often ratio 16:9
+const WIDE_DEVICE_THRESHOLD = 1.6; // 16:10 is max acceptable for screen rotation
 
 export default {
   isPortrait() {
@@ -18,5 +27,11 @@ export default {
   getLandscapeThreshold() {
     const { height, width } = renewDimensions();
     return Math.min(height, width);
+  },
+  isTablet() {
+    const { width, height } = renewDimensions();
+    const min = Math.min(width, height);
+    const max = Math.max(width, height);
+    return max / min <= WIDE_DEVICE_THRESHOLD;
   },
 };
--- a/app/common/src/popup/Popup.js
+++ b/app/common/src/popup/Popup.js
@@ -4,6 +4,7 @@ import * as React from 'react';
 import { View, ScrollView, StyleSheet } from 'react-native';
 
 import Modal from '../Modal';
+import Device from '../Device';
 
 type Props = {|
   children: React.Node,
@@ -22,7 +23,9 @@ export default class Popup extends React.Component<Props> {
       onBackdropPress={this.onClose}
     >
       <View style={styles.body}>
-        <ScrollView>{this.props.children}</ScrollView>
+        <ScrollView contentContainerStyle={styles.content}>
+          {this.props.children}
+        </ScrollView>
       </View>
     </Modal>
   );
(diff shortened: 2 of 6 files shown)


app/common/src/Device.js
```diff
@@ -3,9 +3,16 @@
...
-  return Dimensions.get('screen');
+  try {
+    return Dimensions.get('screen');
+  } catch (e) {
+    return { width: 0, height: 0 };
+  }
 }
 
+// Tablets have often dimensions ratio 3:2-4:3, phones have often ratio 16:9
+const WIDE_DEVICE_THRESHOLD = 0.625; // 16:10 is max acceptable for screen rotation
```
mrtnzlml: What devices did you try with this magic constant? I tried iPad 10.5 inch and it doesn't look good (and also far away from the design again).

app/common/src/Device.js
```diff
@@ -3,9 +3,16 @@
 import { Dimensions } from 'react-native';
 
 function renewDimensions() {
-  return Dimensions.get('screen');
+  try {
+    return Dimensions.get('screen');
+  } catch (e) {
+    return { width: 0, height: 0 };
+  }
```
mrtnzlml: You are probably fixing something. So I miss tests...

app/hotels/src/allHotels/searchForm/SearchForm.js
```diff
@@ -40,6 +40,8 @@ type Props = {
   onLocationChange: (location: string) => void,
 };
 
+export const SEARCH_MIN_WIDTH = 360;
```
mrtnzlml: What is this magic number? It works nicely for every device? Why exactly 360?

app/common/src/Device.js
```diff
@@ -19,4 +26,13 @@ export default {
     const { height, width } = renewDimensions();
     return Math.min(height, width);
   },
+  getDimensions() {
+    return renewDimensions();
```
mrtnzlml: Do not expose this internal function. It renews dimensions every time you call it and it leads to inefficient usage like this (which is wrong):

```js
width: Device.getDimensions().width - SEARCH_MIN_WIDTH,
height: Device.getDimensions().height,
```

app/common/src/Device.js
jakubzaba: This actually fixes tests :) There is no `screen` when tests are run so it throws [this error](https://github.com/facebook/react-native/blob/0cd69e8a02d254577fac9ccd3ff8a1727a024cc8/Libraries/Utilities/Dimensions.js#L93).
I can put comment there or somehow mock globally.

mrtnzlml: Ok, please add the comment explaining what is going on here.

app/common/src/Device.js
jakubzaba: It's explained in comments. It's dimensions ratio that "empirically" make sense to display search results and map on one screen.
What else doesn't look good? Pls next time make a screenshot. It looks good to me.

Now I see map and results are actually connected. Well it's not optimization at all, but pretty much a feature.
We can considered this PR as a one step to final design...

app/hotels/src/allHotels/searchForm/SearchForm.js
jakubzaba: I put comment there.
It's about ~100 width for each input (to not wrap date into 2 rows) + padding (10x4). Should work for each device.

app/common/src/Device.js
mrtnzlml: My first confusion was: _"Why is he talking about threshold 0.625 if 16:10 is clearly 1.6... Aha, it's actually 10:16 - that is confusing."_

Also, it doesn't look OK for me: 
simulator screen shot - ipad pro 10 5-inch - 2018-01-23 at 15 23 02 and since you are using fixed width the landscape is also far away from what is designed here: https://zpl.io/aML67Y7 (search is about 45% of the screen and not just 360 points). It looks weird basically only because of the "360".

app/hotels/src/allHotels/searchForm/SearchForm.js
mrtnzlml: Yes, see my comment above about Zeplin design (https://zpl.io/aML67Y7). You adjusted the design again so if you agreed on this with someone - I am not aware of that.

jakubzaba:
- used 45%.for search results width
- 16:10 comment now corresponds with ratio and `isTablet` calculation
- removed `getDimensions`

_I asked about that 360pts https://zpl.io/aML67Y7. We don't need to strictly stick with design if it has a good reason. I don't like any list of results to be too much wide, that's why I used min width (360pts), nevertheless, 45% makes it simpler and looks still good._

jakubzaba:
@mrtnzlml I often assign you as reviewer but feel free to assign sb else.

mrtnzlml:
> There is one issue. When device is rotated and onLayout is called, it re-render both AllHotels and AllHotelsMap no matter search attributes are the same.

Seems like you solved even this issue. Good job... 👍 Simple solutions are the best! 😊

mrtnzlml:
Integrated. Thanks.

PR #196, opened 2018-01-18
Use currency prop in all relevant queries

jaroslav-kubicek:
Uses `currency` prop passed to `HotelsStandalonePackage` to show prices in that specific currency.

---


app/hotels/src/navigation/AllHotelsNavigationScreen.js
```diff
@@ -75,18 +76,32 @@ class AllHotelsNavigationScreen extends React.Component<Props> {
...
+    const {
+      search,
+      location,
+      filter,
+      onSearchChange,
+      onFilterChange,
+      onLocationChange,
+      onCityIdChange,
+      currency,
+    } = this.props;
```
mrtnzlml: These mega long destructors are ridiculous - I don't see the point.

PR #209, opened 2018-01-23
Use ScrollView only when it's needed

jakubzaba:
---

### DIFF
--- a/app/common/src/popup/Popup.js
+++ b/app/common/src/popup/Popup.js
@@ -9,6 +9,7 @@ type Props = {|
   children: React.Node,
   isVisible: boolean,
   onClose: () => void,
+  doScroll?: boolean,
 |};
 
 export default class Popup extends React.Component<Props> {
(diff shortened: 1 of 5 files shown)


jakubzaba:
Now it uses `doScroll` to disable scrolling.
I wanted to avoid to add another prop, but it's still ok for me.

mrtnzlml:
> I wanted to avoid to add another prop, but it's still ok for me.

I think I understand your concerns but on the other hand look at this PR before and after. This is much much much better (nice to read, easy to understand, simple)... 😊 

And don't worry about too many props if it makes sense (and is useful).

app/common/src/popup/Popup.js
```diff
@@ -22,7 +23,9 @@ export default class Popup extends React.Component<Props> {
       onBackdropPress={this.onClose}
     >
       <View style={styles.body}>
-        <ScrollView>{this.props.children}</ScrollView>
+        <ScrollView scrollEnabled={this.props.doScroll !== false}>
```
mrtnzlml: See? Even `ScrollView` has the ability to disable scroll. This is even more strange since scrolling is what is it all about but it still makes sense. You could've used `scrollEnabled` property as well because it's more descriptive than `doScroll`. Property `somethingEnabled` implicates boolean value but `doSomething` implicates callback or command to do something...

PR #210, opened 2018-01-24
Add simple Playbook for better components development

### LINKED ISSUE #135: Playground PoC
It should be possible to write different variants of the test (executable in Jest) with a possibility to see the results on the mobile device (needed for development - improved Storybook).

See: http://makeitopen.com/docs/en/1-2-design.html#the-design-iteration-cycle

mrtnzlml:
@jaroslav-kubicek Please check it. This is our first idea (with @tbergq help).

Closes #135

---


app/PlaygroundRenderer.js
```diff
@@ -0,0 +1,31 @@
+// @flow
+
+import * as React from 'react';
+import ShallowTestRenderer from 'react-test-renderer/shallow';
+import { it } from 'global';
+
+if (it === undefined && process.env.NODE_ENV !== 'test') {
+  global.it = (testName, testCallback) => testCallback();
```
jaroslav-kubicek: da fak? 😄

mrtnzlml:
> I'm starting thinking about some possible enhacements: like switch between normal app & playbook with ENV variable or to have something like "category" for each component, consisting of multiple sections with custom props.

We definitely need the first part (no matter whether it's going to be ENV or something else). About the second part - let's keep it simple now. I don't want to reinvent Storybook here and everything must have a good reason... 😊 

You can display one component at the time in the Playground right now and I would like to stick with it for a while. Mainly because we are not doing a showcase of components but we want to make the development of the components easier.

app/PlaygroundRenderer.js
mrtnzlml: 🔮

PR #211, opened 2018-01-24
Log real events in hotels package

mrtnzlml:
Note - we do not have information about `hasActiveBooking`.

See: https://confluence.kiwi.com/display/MOB/Semi+native+hotels#Seminativehotels-Logging

---

### DIFF
--- a/app/core/config/jestSetup.js
+++ b/app/core/config/jestSetup.js
@@ -1,0 +1,24 @@
+// @flow
+
+beforeEach(() => {
+  /**
+   * Jest will swallow all `console.log` methods used in our code
+   * and it will throw error for `console.error` methods. You have to
+   * overwrite this mock if you want to test these logs (in the test):
+   *
+   * ```
+   * let logs;
+   * beforeEach(() => {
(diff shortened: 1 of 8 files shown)


tbergq:
if the information about has active booking exists in the native code, it can be passed as a prop to the react native application from native code through: `initialProperties`
https://facebook.github.io/react-native/docs/integration-with-existing-apps.html

mrtnzlml:
@tbergq I decided not to log it now until we know where to get this information. It basically means I want to be sure about this in order to implement it correctly (not blindly):

> if the information about has active booking exists in the native code

app/core/config/jestSetup.js
```diff
@@ -0,0 +1,24 @@
+// @flow
+
+beforeEach(() => {
+  /**
+   * Jest will by swallow all `console.log` methods used in our code
```
mrtnzlml: Fuck, edit typo. Should be `Jest will swallow...`.

tbergq:
Makes sense :)

PR #212, opened 2018-01-24
Use CurrencyFormatter in Price component

mrtnzlml:
The only problem is that I don't really know how the formatter works
because there are not enough details in the specification.

See: https://confluence.kiwi.com/display/MOB/Semi+native+hotels#Seminativehotels-Currencyratesandformatters

---

### DIFF
--- a/app/common/src/Price.js
+++ b/app/common/src/Price.js
@@ -3,15 +3,15 @@
 import * as React from 'react';
 import { Text } from 'react-native';
 
+import CurrencyFormatter from './services/CurrencyFormatter';
+import type { StylePropType } from '../types/Styles';
+
 type Props = {|
   amount: number | null, // may be null (compatible with GraphQL query failure)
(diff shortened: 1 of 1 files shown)


app/common/src/Price.js
```diff
@@ -2,16 +2,16 @@
 
 import * as React from 'react';
 import { Text } from 'react-native';
+import type { StylePropType } from '@kiwicom/react-native-app-common';
```
jaroslav-kubicek: Why is there import from package, since `Price` is part of the `common` as well? :)

mrtnzlml: It's wrong. I was actually thinking about using public interface even inside of this package (and I will probably do it at least in tests like React itself did), but this is a mistake. I will fix it...

mrtnzlml: Fixed.

PR #213, opened 2018-01-24
Automatically adjust date ranges in search form

jaroslav-kubicek:
Relates to #109 

---


app/hotels/src/allHotels/searchForm/SearchForm.js
```diff
@@ -50,16 +52,57 @@ class SearchForm extends React.Component<Props> {
...
+  getValidDates = (
+    checkin: moment,
+    checkout: moment,
+    whatToAdjust: DateInputType,
+  ) => {
+    if (checkin === null || checkout === null) {
+      return null;
+    }
+    const diff = moment(checkout).diff(checkin, 'days');
+    if (diff < 1 || diff > 10) {
```
mrtnzlml: I don't understand this condition. Can you make it more clear?

mrtnzlml:
Also, this behavior is really weird and I was not able to figure out what is going on and why does it work like this. Where did the Jan 26 come from? 🙂 

jan-25-2018 12-19-15

app/hotels/src/allHotels/searchForm/SearchForm.js
```diff
@@ -50,16 +52,57 @@ class SearchForm extends React.Component<Props> {
...
-  handleDateChange = (datestring: string, type: 'checkin' | 'checkout') => {
-    const date = moment(datestring, DISPLAY_DATE_FORMAT).toDate();
+  handleDateChange = (datestring: string, type: DateInputType) => {
+    const date = moment(datestring, DISPLAY_DATE_FORMAT);
+    const checkin =
+      type === 'checkin' ? date : moment(this.props.search.checkin);
+    const checkout =
+      type === 'checkout' ? date : moment(this.props.search.checkout);
+    const whatToAdjust = type === 'checkin' ? 'checkout' : 'checkin';
+    const dates = this.getValidDates(checkin, checkout, whatToAdjust);
```
mrtnzlml: This part of the code is also not super readable. Especially because of ternary conditions like `'checkin' ? 'checkout' : 'checkin'`... 😊

app/hotels/src/allHotels/AllHotelsSearch.js
```diff
@@ -7,10 +7,8 @@ import {
   Layout,
   FullPageLoading,
   GeneralError,
-  ErrorMessage,
```
mrtnzlml: We can remove this component.

jakubzaba:
That weird date changing to Jan 26th is functionality from datepicker with `minDate`. `Start date has minDate tdy and End date is set to tmw, therefore it scrolls to 26. (Booking conditions, and common sense, says you have to set End date as 1 day after Start date).

mrtnzlml:
Ok, I think it should not modify date you are currently editing. It's very unexpected. I actually wanted to set 2018-01-01 but this happened. @jaroslav-kubicek can you please check it?

jaroslav-kubicek:
This is not a new behaviour you are describing ^^. As @jakubzaba  noted, it's caused by `minDate`

mrtnzlml:
@jaroslav-kubicek You do not consider it a bug?

jaroslav-kubicek:
you are probably right, I was just trying to point out it's not related to this PR...

app/hotels/src/allHotels/AllHotelsSearch.js
jaroslav-kubicek: removed

app/hotels/src/allHotels/searchForm/SearchForm.js
jaroslav-kubicek: fixed

app/hotels/src/allHotels/searchForm/SearchForm.js
jaroslav-kubicek: fixed

jaroslav-kubicek:
btw I would prefer to do refactoring in separate PR, to keep it short 🙂

mrtnzlml:
> btw I would prefer to do refactoring in separate PR, to keep it short

I appreciate it. But please do so. It needs refactoring quite urgently because it's absolutely unusable. I never noticed this. You cannot change date if you set January, you cannot change year at all (see: https://github.com/kiwicom/react-native-app/issues/214 - tests expected).

jaroslav-kubicek:
good spot, my idea is to probably remove min limitation and just adjust date after it's set by user in method added in this PR

mrtnzlml:
@jaroslav-kubicek Not necessary. I just did not understand this sentence:

> That weird date changing to Jan 26th is functionality from datepicker with minDate. `Start date has minDate tdy and End date is set to tmw, therefore it scrolls to 26. (Booking conditions, and common sense, says you have to set End date as 1 day after Start date).

Should be "you cannot set past dates"... :roll_eyes:

PR #215, opened 2018-01-25
Fix missing dependency on querystring

VaclavSir:
---


mrtnzlml:
I do not understand JS environment in RN... 😓

PR #216, opened 2018-01-26
Use max price from all available hotels query

### LINKED ISSUE #197: Show price range according to real data
Currently, the price range in the filter has fixed values (0-300 EUR) - let's fetch real max/min values from api and adjust these values accordingly.

tbergq:
Closes #197 

---

### DIFF
--- a/app/hotels/src/HotelsReducer.js
+++ b/app/hotels/src/HotelsReducer.js
@@ -8,19 +8,22 @@ import type {
   FilterParams,
   OnChangeFilterParams,
 } from './filter/FilterParametersType';
+import type { CurrentSearchStats } from './filter/CurrentSearchStatsType';
 
 export type HotelsReducerState = {
   cityId: string | null,
   location: string,
   searchParams: SearchParams,
   filterParams: FilterParams,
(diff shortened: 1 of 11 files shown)


mrtnzlml:
There is still one issue - application freeze when you try to click before it's loaded. It happened to me in the first 2 seconds of testing so it should be resolved somehow... :)

tbergq:
Ok, I will look into it

tbergq:
Seems it was the inital reducer value of Number.MAX_SAFE_INTEGER that made the ui freeze.

mrtnzlml:
What do you mean? It doesn't work in RN? How did you discover it so fast?

tbergq:
I had a hunch that it could be this, since it opens when the value has been changed. 

Probably the slider component that does not like a number that big (just guessing)

mrtnzlml:
@tbergq Yould you please rebase it? There are changes in `PriceFilter` and I am not sure how to fix it properly.

tbergq:
Rebaes

mrtnzlml:
Good, integrated. Thanks.

PR #219, opened 2018-01-29
Show active filters in stripe first before others

### LINKED ISSUE #201: Sort active filters
Put active filters at the beginning of stripe.

jaroslav-kubicek:
Closes #201 

---

### DIFF
--- a/app/hotels/src/filter/stars/StarsFilter.js
+++ b/app/hotels/src/filter/stars/StarsFilter.js
@@ -7,24 +7,38 @@ import StarsPopup from './StarsPopup';
 import FilterButton from '../FilterButton';
 import type { OnChangeFilterParams } from '../FilterParametersType';
 
-type Props = {|
-  stars: number[],
+type Stars = number[];
+
+type Props = {
+  stars: Stars,
   onChange: OnChangeFilterParams => void,
-|};
+};
 
 type State = {|
   isPopupOpen: boolean,
 |};
 
 export default class StarsFilter extends React.Component<Props, State> {
+  static isActive = (stars: Stars): boolean => stars.length > 0;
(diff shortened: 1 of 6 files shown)


app/hotels/src/filter/stars/StarsFilter.js
```diff
@@ -58,3 +62,5 @@ export default class StarsFilter extends React.Component<Props, State> {
     );
   }
 }
+
+StarsFilter.isActive = (stars: Stars) => stars.length > 0;
```
mrtnzlml: Is it a good idea to have half of the methods in the class and half of the methods outside of the class? Is it well readable for others? Will everyone expect to have additional methods somewhere in this file?

jaroslav-kubicek: thx for input, that's exactly one of the reasons what I couldn't decide and created PR - it's logic related to that specific component vs. code that is split across files

mrtnzlml: I mean the design is probably alright:

- stripe is responsible for you know striping: being scrollable, having all the filters and sort them
- the filter itself is responsible for being active or inactive (no reason to have it in the stripe)

So my point is only the code readability - I would prefer `static isActive` method instead of attaching it like this. Otherwise, I think it's alright (read I don't know how to do it better)... 🙂

mrtnzlml:
@jaroslav-kubicek So is this still WIP (see description) or can I try to integrate it?

jaroslav-kubicek:
oops,description updated, ready for integration @mrtnzlml

mrtnzlml:
Integrated without the "mounted" hack.

PR #222, opened 2018-01-31
Setup fastlane beta for ios

tbergq:
---


ios/reactNativeApp-tvOS/Info.plist
```diff
@@ -15,7 +15,7 @@
 	<key>CFBundlePackageType</key>
```
mrtnzlml: Can we remove this directory completely? We don't need `reactNativeApp-tvOS` right?

tbergq: I am 99% sure that we can remove this, I will test and remove it with the final PR

mrtnzlml:
Tested and integrated into master. Thanks.

PR #224, opened 2018-02-01
Show distance from city center in hotel results

jakubzaba:
**Merge after [Gitlab 591]**

---


mrtnzlml:
Looks good - Martin is happy.

Not sure though how is "15.8km from the center" useful... :)

PR #228, opened 2018-02-04
added RNLogging as subproject into current project, removed dependenc…

pistelak:
I've added `RNLogging` directly as a subproject until I (or we) will be able to resolve all issues with the build. 

I'm able to build project after replacing two import statements (https://github.com/facebook/react-native/issues/16039, https://github.com/facebook/react-native/issues/13198) with my current configuration *Xcode Version 9.2 (9C40b)* and *Cocoapods: 1.3.1*, but I have a newer version of the cocoapods on my second device and there were many more issues... 

Anyway, I tried to call native code with the following and it works...: 
```
import { NativeModules, StyleSheet } from 'react-native';

NativeModules.RNLoggingModule.ancillaryDisplayed("Hello from React Native!", false);
```

---

Please don't merge this PR.

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "node-fetch": "^1.7.3",
     "prettier": "^1.7.4",
     "react-native-dotenv": "^0.1.0",
-    "react-test-renderer": "16.2.0"
+    "react-test-renderer": "16.2.0",
+    "react-native-native-modules": "^0.1.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -40,6 +40,6 @@
     "prettier": "^1.7.4",
     "react-native-dotenv": "^0.1.0",
     "react-test-renderer": "16.2.0",
-    "react-native-native-modules": "^0.1.0"
+    "@kiwicom/react-native-native-modules": "^0.1.0"
   }
 }
--- a/ios/Podfile
+++ b/ios/Podfile
@@ -1,15 +1,14 @@
 # Uncomment the next line to define a global platform for your project
-# platform :ios, '9.0'
+platform :ios, '9.0'
 
-target 'reactNativeApp' do
-  # Uncomment the next line if you're using Swift or would like to use dynamic frameworks
-  # use_frameworks!
+workspace 'reactNativeApp.xcworkspace'
 
-  # Pods for reactNativeApp
-  rn_path = '../node_modules/react-native'
-  rn_maps_path = '../node_modules/react-native-maps'
+# Pods for reactNativeApp
+rn_path = '../node_modules/react-native'
+rn_maps_path = '../node_modules/react-native-maps'
 
-  pod 'yoga', path: "#{rn_path}/ReactCommon/yoga/yoga.podspec"
+def react_common(rn_path) 
+
   pod 'React', path: rn_path, subspecs: [
     'Core',
     'RCTActionSheet',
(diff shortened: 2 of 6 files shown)


pistelak:
FYI: I've updated `cocoapods` and there are still only these two issues. I don't know what is different on my second mac... 🤔

package.json
```diff
@@ -15,7 +15,7 @@
     "start": "yarn react-native start",
     "test": "node node_modules/jest/bin/jest.js --config=.jest.json",
     "test-ci": "./scripts/test.sh",
-    "postinstall": "node ./scripts/configureApplication.js"
+    "postinstall": "node ./scripts/configureApplication.js && sed -i '' 's\/#import <fishhook\\/fishhook.h>\/#import \"fishhook.h\"\/' ./node_modules/react-native/Libraries/WebSocket/RCTReconnectingWebSocket.m && sed -i '' 's\/#import <RCTAnimation\\/RCTValueAnimatedNode.h>\/#import \"RCTValueAnimatedNode.h\"\/' ./node_modules/react-native/Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.h"
```
tbergq: could this rather be added to the configureApplication script?
`const { exec } = require('child_process');`
`exec('sed -i ...')`

mrtnzlml: I can tweak these minor issues during integration. The same applies to squashing all commits and similar. We just need to have a working example and not copy-paste... 😞

pistelak: Thanks, I wasn't aware of `configureApplication` script... 

Is there anything else we need to do or another method to add? In other case I will create `pod`.., but then adding methods will be more complicated. @mrtnzlml @tbergq

mrtnzlml: Probably no - please extract it into our private `react-native-native-modules` repo and require it here.

tbergq: Debugging menu and code reloading needs to work

pistelak: @tbergq I don't know if I can help you with that. Do you have any thoughts what is wrong?

tbergq: I can try to look into it tomorrow. I don't know why. I cloned your repo, tested from your master branch, and there it works, in the ios-logging-module-dev it does not. So there must be something in this PR that does it.

pistelak: The problem is that I don't know what debug menu is and how to open it. 
Should I read https://facebook.github.io/react-native/docs/debugging.html or is there any better docs?

tbergq: If you are on a fysical device, shake it, if you are on simulator, click on hardware => shake gesture. It should look like this: 

skjermbilde 2018-02-07 kl 10 14 25

pistelak:
@tbergq Please, can you test it. It doesn't hurry... :)

tbergq:
Looks good. Debug menu is up and running.

package.json
```diff
@@ -39,6 +39,7 @@
     "node-fetch": "^1.7.3",
     "prettier": "^1.7.4",
     "react-native-dotenv": "^0.1.0",
-    "react-test-renderer": "16.2.0"
+    "react-test-renderer": "16.2.0",
+    "react-native-native-modules": "^0.1.0"
```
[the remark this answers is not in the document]
pistelak: @mrtnzlml It was me radek.pistelak@kiwi.com (@kiwi.mobile.user). Ehm, package namespace? Should I use @kiwi.mobile.user as prefix (http://blog.npmjs.org/post/116936804365/solving-npms-hard-problem-naming-packages)?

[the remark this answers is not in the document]
mrtnzlml: Package name is going to be `@kiwicom/something`. See: https://github.com/kiwicom/react-native-app/blob/master/app/core/package.json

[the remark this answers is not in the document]
pistelak: Should be done.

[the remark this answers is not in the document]
mrtnzlml: I don't see it yet... :) https://www.npmjs.com/org/kiwicom

Just publish it again and we are good to go. We will probably handle this publish process in the future...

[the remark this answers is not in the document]
pistelak: Published.

mrtnzlml:
@pistelak Integrated into master. Thanks.

PR #238, opened 2018-02-07
Change filter deactivate behavior

tbergq:
Move filter logic to separate reducer

---

### DIFF
--- a/app/hotels/src/filter/FilterParametersType.js
+++ b/app/hotels/src/filter/FilterParametersType.js
@@ -1,19 +1,26 @@
 // @flow
 
-export type FilterParams = {
+export type FilterParams = {|
   starsRating: number[],
   minPrice: number | null,
   maxPrice: number | null,
   freeCancellation: boolean,
   hotelFacilities: string[],
   minScore: number | null,
-};
+|};
 
-export type OnChangeFilterParams = {
-  starsFilter?: number[],
+export type OnChangeFilterParams = {|
+  starsRating?: number[],
   minPrice?: ?number,
   maxPrice?: ?number,
   freeCancellation?: boolean,
   hotelFacilities?: string[],
   minScore?: ?number,
-};
+|};
+
+export type ActiveFilters = {|
+  isPriceFilterActive: boolean,
+  isStarsFilterActive: boolean,
+  isMinScoreActive: boolean,
+  isHotelFacilitiesActive: boolean,
+|};
--- a/app/hotels/src/filter/FilterStripe.js
+++ b/app/hotels/src/filter/FilterStripe.js
@@ -9,11 +9,14 @@ import PriceFilter from './price/PriceFilter';
 import FreeCancellationFilter from './freeCancellation/FreeCancellationFilter';
 import HotelFacilitiesFilter from './hotelFacilities/HotelFacilitiesFilter';
 import ScoreFilter from './score/ScoreFilter';
+import type { FilterReducerState } from './FiltersReducer.js';
+import type { CurrentSearchStats } from './CurrentSearchStatsType';
 import type {
+  ActiveFilters,
   FilterParams,
   OnChangeFilterParams,
 } from './FilterParametersType';
-import type { CurrentSearchStats } from './CurrentSearchStatsType';
+import type { HotelsReducerState } from '../HotelsReducer';
 
 const styles = {
   view: {
@@ -31,6 +34,7 @@ type Props = {|
   filter: FilterParams,
   currency: string,
   currentSearchStats: CurrentSearchStats,
+  activeFilters: ActiveFilters,
 |};
 
 /**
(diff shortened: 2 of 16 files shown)


app/hotels/src/filter/FilterParametersType.js
```diff
@@ -17,3 +17,10 @@ export type OnChangeFilterParams = {
   hotelFacilities?: string[],
   minScore?: ?number,
 };
+
+export type ActiveFilters = {
+  isPriceFilterActive: boolean,
+  isStarsFilterActive: boolean,
+  isMinScoreActive: boolean,
+  isHotelFacilitiesActive: boolean,
+};
```
mrtnzlml: Please use everywhere in this file exact types - should be possible. I noticed there is an error with `OnChangeFilterParams` type.

app/hotels/src/filter/FilterStripe.js
```diff
@@ -31,13 +34,20 @@ type Props = {|
...
  * This filter holds all available hotel filters. Active (selected) filters are
  * rendered first.
  */
 function FilterStripe(props: Props) {
+  const {
+    isPriceFilterActive,
+    isStarsFilterActive,
+    isMinScoreActive,
+    isHotelFacilitiesActive,
+  } = props.activeFilters;
```
mrtnzlml: I am not super big fan of doing this. It's IMHO not very well readable. Example:

```js
<StarsFilter
    key="stars"
    stars={starsRating}
    onChange={props.onChange}
    isActive={isStarsFilterActive}
/>
```

Based on what should I know what is `starsRating` and `isStarsFilterActive`? This information is lost somewhere at the beginning of this component.

app/hotels/src/filter/FilterStripe.js
```diff
@@ -125,8 +134,15 @@ function FilterStripe(props: Props) {
   );
 }
 
-const mapStateToProps = state => ({
-  currentSearchStats: state.hotels.currentSearchStats,
+const mapStateToProps = ({
```
mrtnzlml: Protip (much better naming): screen shot 2017-12-08 at 15 23 05

app/hotels/src/filter/FiltersReducer.js
```diff
@@ -0,0 +1,78 @@
...
+      hotelFacilities,
+    },
+  } = newState;
+  return {
+    ...newState,
+    activeFilters: {
+      isPriceFilterActive: minPrice !== null || maxPrice !== null,
+      isStarsFilterActive: starsRating.length > 0,
+      isMinScoreActive: minScore !== null,
+      isHotelFacilitiesActive: hotelFacilities.length > 0,
```
mrtnzlml: Should be this reducer responsible for "is active" logic of every filter? I am not sure about that...

app/hotels/src/filter/__tests__/FiltersReducer.test.js
```diff
@@ -0,0 +1,52 @@
+// @flow
+
+import FiltersReducer, { InitialFiltersReducerState } from '../FiltersReducer';
+
+const type = 'setFilter';
+
+describe('FiltersReducer', () => {
+  it('should return the initialState', () => {
+    // $FlowFixMe
```
mrtnzlml: See `.flowconfig`:

> ; Please, do not use "fix me" suppress types and comments

If it's necessary (we are in tests so I am ok with that) use something else (`$FlowIssue` or `$FlowExpectedError` with an explanation).

app/hotels/src/filter/hotelFacilities/HotelFacilitiesFilter.js
```diff
@@ -10,6 +10,7 @@ import type { OnChangeFilterParams } from '../FilterParametersType';
 type Props = {|
   facilities: string[],
   onChange: OnChangeFilterParams => void,
+  isActive: boolean,
```
mrtnzlml: Alright, maybe it makes sense to move this logic outside.

app/hotels/src/filter/FilterStripe.js
```diff
@@ -125,8 +134,15 @@ function FilterStripe(props: Props) {
...
+  hotels,
+  filters,
+}: {
+  hotels: HotelsReducerState,
```
mrtnzlml: Do we need this reducer here?

app/hotels/src/filter/__tests__/FiltersReducer.test.js
```diff
@@ -0,0 +1,52 @@
+// @flow
+
+import FiltersReducer, { InitialFiltersReducerState } from '../FiltersReducer';
+
+const type = 'filtersReducer/FILTER_CHANGED';
+
+describe('FiltersReducer', () => {
+  it('should return the initialState', () => {
+    // $FlowExpectedError: Need to test that it returns the inital state when called first time
```
mrtnzlml: Just a small note: this comment is not that helpful because you already revealed your intentions in the test description (`describe` + `it`). It's more important to explain why are you actually suppressing this error. So for example: _"the second argument is intentionally not a valid action"_... :)

I will fix it.

mrtnzlml:
Integrated.

PR #243, opened 2018-02-08
Set max dates for date pickers

### LINKED ISSUE #237: Hotels: server error while trying to search hotels far in the future
Steps to reproduce:

- go to hotels search
- change checkout date to something like 2022
- select date

Server returns "server error". Why? Also, I checked CloudWatch and it seems like our client is not sending correct date (year) to the server/Booking.com.

tbergq:
closes #237

### DIFF
--- a/app/hotels/src/allHotels/searchForm/DateInput.js
+++ b/app/hotels/src/allHotels/searchForm/DateInput.js
@@ -94,11 +94,18 @@ export default class DateInput extends React.Component<Props> {
 
     return (
       <View style={styles.container}>
+        {/* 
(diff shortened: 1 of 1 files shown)


mrtnzlml:
@tbergq I am really starting to be scared of `moment` everywhere. Should we wrap it into our object so we can tweak the implementation if needed? For example, we have date formatting in so many places - should be in one class IMHO so we can adjust it according to selected user interface language (so many use-cases like this)...

tbergq:
Yes, I think that makes sense.

PR #254, opened 2018-02-13
Add pagination container for AllHotelsSearchList

### LINKED ISSUE #249: Investigate why we get only small amount of hotels from API
There are complains that we are showing only small amount of hotels compare to Booking.com. Please investigate what is going on. Are there default limits? (I don't think so) Is Booking.com API returning only limited results? How to fix that? Thanks...

tbergq:
closes #249 

https://github.com/kiwicom/graphql/pull/4
must be merged before this PR


app/hotels/src/allHotels/AllHotelsSearchList.js
```diff
@@ -42,47 +106,91 @@ export class AllHotelsSearchList extends React.Component<Props> {
...
-          })}
-        </ScrollView>
+        <View style={styles.container}>
+          <FlatList
+            data={hotels}
+            renderItem={this.renderListItem}
+            keyExtractor={this.keyExtractor}
+            onEndReached={this.loadMore}
+          />
+          {this.renderLoadingIndication()}
```
mrtnzlml: I will have to stop accepting this pattern. I think we have components so let's use them... 🙂

app/hotels/src/allHotels/AllHotelsSearchList.js
```diff
@@ -2,38 +2,102 @@
 
 import * as React from 'react';
 import idx from 'idx';
-import { createFragmentContainer, graphql } from 'react-relay';
-import { ScrollView, Text } from 'react-native';
-import { CenteredView, Logger } from '@kiwicom/react-native-app-shared';
+import { createPaginationContainer, graphql } from 'react-relay';
+import type { Relay } from 'react-relay';
```
mrtnzlml: Are you sure this type actually exists? It seems that I can replace the props with this:

```diff
- relay: Relay,
+ relay: any,
```

and it still works. This is what actually happens when you import non-existent type. I would replace it with something more obvious if it's intentional:

```diff
- relay: Relay,
+ relay: $FlowFixMeProps,
```

app/hotels/src/allHotels/AllHotelsSearchList.js
```diff
@@ -2,38 +2,102 @@
...
+    <AllHotelsSearchRow
+      data={item.node}
+      openSingleHotel={this.props.openSingleHotel}
+    />
+  );
+
+  renderLoadingIndication = () => {
+    if (this.state.isLoading) {
+      return (
+        <View style={styles.loadingView}>
```
mrtnzlml: I noticed that the loading indicator sometimes hovers over the last hotel like when you have a footer with CSS fixed position. Is it intentional?

app/hotels/src/allHotels/AllHotelsSearchList.js
```diff
@@ -42,47 +106,91 @@ export class AllHotelsSearchList extends React.Component<Props> {
...
+    direction: 'forward',
+    getConnectionFromProps(props) {
+      return props.data && props.data.allAvailableHotels;
+    },
+    getFragmentVariables(prevVars, totalCount) {
+      return {
+        ...prevVars,
+        count: totalCount,
+      };
+    },
```
mrtnzlml: Options `direction`, `getConnectionFromProps` and `getFragmentVariables` are optional and I think unnecessary in this case. What do you think about removing them?

tbergq: Nice, I did not know they were optional, I will remove them

app/hotels/src/allHotels/AllHotelsSearchList.js
tbergq: Though it was the only way, but with some googling I found ListFooterComponent, so I can render it at the end of the list.

app/hotels/src/allHotels/AllHotelsSearchList.js
tbergq: I think this should have been `RelayPaginationProp`

mrtnzlml: Same problem, is it? Relay is not exporting Flow types as far as I know. From my point of view, this is still effectively `any` type.

mrtnzlml: Well easy to verify that types doesn't work here:

```diff
- if (this.props.relay.hasMore() && !this.props.relay.isLoading()) {
+ if (this.props.relay.hasMore() && !this.props.relay.isLoadingShit()) {
```

```
▲ react-native-app at 249-hotels-pagination ✖ y flow
yarn run v1.3.2
$ /Users/mrtnzlml/Work/kiwi-opensource/react-native-app/node_modules/.bin/flow
No errors!
✨  Done in 0.21s.
```

mrtnzlml:
Integrated, thanks. I didn't use the flow-typed for Relay yet because there is a lot of errors.

PR #255, opened 2018-02-13
Add padding to BarPopup link buttons

### LINKED ISSUE #245: Enlarge touchable areas for buttons in `BarPopup` components
Maybe even add a bigger vertical padding. Maybe make the whole 50% of the bar touchable. It's hard to hit them by fat finger:

screen_shot_2018-02-06_at_12 01 57

img_5692

tbergq:
closes #245

### DIFF
--- a/app/shared/src/popup/BarPopup.js
+++ b/app/shared/src/popup/BarPopup.js
@@ -29,13 +29,35 @@ export default class BarPopup extends React.Component<Props> {
   render = () => (
     <Popup isVisible={this.props.isVisible} onClose={this.onClose}>
       <View style={styles.bar}>
-        <LinkButton title="Cancel" onPress={this.onClose} />
-        <LinkButton title={this.props.buttonTitle} onPress={this.onSave} />
+        <LinkButton
+          title="Cancel"
+          onPress={this.onClose}
+          styles={cancelLinkButtonStyles}
+        />
+        <LinkButton
+          title={this.props.buttonTitle}
+          onPress={this.onSave}
+          styles={saveLinkButtonStyles}
+        />
       </View>
       {this.props.children}
     </Popup>
   );
 }
+
+const saveLinkButtonStyles = StyleSheet.create({
+  buttonWrapper: {
+    paddingRight: 10,
+    paddingLeft: 50,
+  },
+});
+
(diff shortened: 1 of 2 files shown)


app/shared/src/popup/BarPopup.js
```diff
@@ -15,6 +15,20 @@ type Props = {|
   onSave: Function,
 |};
 
+const saveLinkButtonStyles = {
```
mrtnzlml: I would prefer `StyleSheet.create`. Also, we should not reinvent a wheel here so this may be better (`LinkButton`):

```diff
- styles={{ ...styles, ...buttonProps.styles }}
+ styles={[ styles, buttonProps.styles ]}
```

app/shared/src/popup/BarPopup.js
```diff
@@ -29,8 +43,16 @@ export default class BarPopup extends React.Component<Props> {
   render = () => (
     <Popup isVisible={this.props.isVisible} onClose={this.onClose}>
       <View style={styles.bar}>
-        <LinkButton title="Cancel" onPress={this.onClose} />
-        <LinkButton title={this.props.buttonTitle} onPress={this.onSave} />
+        <LinkButton
+          title="Cancel"
+          onPress={this.onClose}
+          styles={cancelLinkButtonStyles}
```
mrtnzlml: Also, I would change "styles" => "style" to stick with the common naming: `<View style={styles.view}>`...

app/shared/src/popup/BarPopup.js
tbergq: This will not work with how the button component is now

app/shared/src/popup/BarPopup.js
tbergq: LinkButton expects ButtonProps which is:
```
export type Props = {
  title: string,
  onPress?: Function,
  // $FlowFixMeProps
  styles?: {|
    buttonWrapper?: Object,
    button?: Object,
    buttonText?: Object,
    icon?: Object,
  |},
  icon?: Object,
};
```

mrtnzlml: Oh yeah, `Button` component design is pretty shitty. I already started to work on refactoring. Alright, thanks for the explanation.

PR #257, opened 2018-02-14
Move google logo out of view

### LINKED ISSUE #251: Map preview: try to get rid of Google logo
It's probably not possible to turn it off so maybe just move the map a little bit:

screenshot_20180207-092011

tbergq:
closes #251


mrtnzlml:
I need to resolve #258 first. I am not sure if this solution is correct because there is also map marker and I need to verify that the position of this marker is still correct.

tbergq:
I could still see the map marker on iOS, but I do agree, we should resolve #258 first!

mrtnzlml:
Merged, thanks. Interestingly the drop marker is not visible at all on Android devices. See: #265

PR #263, opened 2018-02-14
Ready use for google maps

### LINKED ISSUE #258: Google map doesn't work
Can you please help me with that? I was not able to make it work even though it should be only about adding one key into `AndroidManifest.xml`...

tbergq:
closes #258

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -2,10 +2,13 @@
 
 buildscript {
     repositories {
+        // this line will fix Could not find com.android.tools.build:gradle:3.0.1
+        // https://stackoverflow.com/questions/47446681/react-native-run-android-could-not-find-com-android-tools-buildgradle3-0-1#answer-47456492
+        google()
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.2.3'
+        classpath 'com.android.tools.build:gradle:3.0.1'
 
         // NOTE: Do not place your application dependencies here; they belong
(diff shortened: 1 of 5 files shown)


tbergq:
You need to use an Emulator with API version 24

mrtnzlml:
@tbergq Always exactly 24? This should be in the docs. What about the `yarn android`?

android/.project
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mrtnzlml: What are these files?!

android/build.gradle
```diff
@@ -2,10 +2,11 @@
 
 buildscript {
     repositories {
+        google()// this is the magic line which fixes build failure on yarn android
```
mrtnzlml: Can you please add a better explanation? I don't like the fact that we have to use magic here but we don't know what it actually means.

tbergq:
I have misread the documentation. It says `The Google Play images are available starting with Android Nougat (API 24)` . So there can be newest API Version. The emulator image though needs to support google play

skjermbilde 2018-02-14 kl 14 02 26

I will add this information to the readme

android/.project
tbergq: I checked with our android guy, and it seems this should be in the `.gitignore` file. I will add it.

mrtnzlml:
This is not fix at all. I managed to make it work even without this fix... 🙄

PR #275, opened 2018-02-20
Make large photo show on android, fix top position for iphoneX

### LINKED ISSUE #269: Large hotel photos are not displaying on Android
But only sometimes - see: https://github.com/leecade/react-native-swiper/issues/227

tbergq:
closes #269

### DIFF
--- a/app/hotels/src/gallery/PhotosStripe.js
+++ b/app/hotels/src/gallery/PhotosStripe.js
@@ -1,7 +1,7 @@
 // @flow
 
 import * as React from 'react';
-import { StyleSheet, View } from 'react-native';
+import { StyleSheet, View, Platform } from 'react-native';
 import Swiper from 'react-native-swiper';
 import { NetworkImage } from '@kiwicom/react-native-app-shared';
 
@@ -14,7 +14,25 @@ type Props = {|
   onClose: () => void,
 |};
 
-export default class PhotosStripe extends React.Component<Props> {
+type State = {|
+  key: number,
+|};
+
+/* 
+  Need to trigger a re-render on android to show photo
+  https://github.com/leecade/react-native-swiper/issues/227
+*/
+export default class PhotosStripe extends React.Component<Props, State> {
+  state = {
+    key: Math.random(),
+  };
+
+  componentDidMount = () => {
+    if (Platform.OS === 'android') {
+      setTimeout(() => this.setState({ key: Math.random() }));
+    }
+  };
+
   renderPagination = (index: number, total: number) => {
     return (
       <View style={{ position: 'absolute', top: 20, width: '100%' }}>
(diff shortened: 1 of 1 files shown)


app/hotels/src/gallery/PhotosStripe.js
```diff
@@ -14,10 +14,28 @@ type Props = {|
...
+  https://github.com/leecade/react-native-swiper/issues/227
+*/
+export default class PhotosStripe extends React.Component<Props, State> {
+  state = {
+    key: Math.random(),
+  };
+
+  componentDidMount = () => {
+    if (Platform.OS === 'android') {
+      setTimeout(() => this.setState({ key: Math.random() }), 200);
```
mrtnzlml: @tbergq Is this solution sufficient? Do you know why 200 and not something else?

tbergq: I have tested it, and the photo always shows. I guess the main idea here is to trigger re-render. Could be that some other value lower than 200ms also will do the trick. 

Do you want me to investigate that further?

mrtnzlml: Yes please, give it few more minutes to understand this. For example - does it work when the image is not in the cache (i.e. we need to wait longer for the download)? Does it matter?

tbergq: removing the setState call from timeout makes the opening of the modal laggy. Having it in a timeout without any delay (making it execute async) prevents the laggy behaviour and renders the image.

app/hotels/src/gallery/PhotosStripe.js
```diff
@@ -14,10 +14,28 @@ type Props = {|
...
+  componentDidMount = () => {
+    if (Platform.OS === 'android') {
+      setTimeout(() => this.setState({ key: Math.random() }));
+    }
+  };
+
   renderPagination = (index: number, total: number) => {
     return (
-      <View style={{ position: 'absolute', top: 20, width: '100%' }}>
+      <View style={{ position: 'absolute', top: 40, width: '100%' }}>
```
mrtnzlml: Is this good solution? This is why I suggested `SafeAreaView` because now it looks really weird on normal devices... 🙂

mrtnzlml:
Also, I added this style as a first attempt to solve it but it's probably not needed now:

```diff
networkImage: {
  flex: 1,
- width: '100%',
- height: '100%',
},
```

app/hotels/src/gallery/PhotosStripe.js
tbergq: Ok, I can see if we can make it work with SafeAreaView, I will remove it from this PR.
SafeAreaView will not work with `position: 'absolute'`

mrtnzlml: Leave it if not possible - it's just that the behavior is much more sophisticated... :)

tbergq: Yeah, I will leave it out of this PR, and make a new one for the Header on iphone X

mrtnzlml:
Integrated, thanks.

PR #286, opened 2018-02-21
Device store dimensions from rootview

### LINKED ISSUE #278: Split view doesn't work with correct dimensions
See: https://slack-files.com/T024Z3H2Y-F93TJECJX-c73e2d3cd4 (it should be responsive - change to mobile view). I tried to investigate a little bit few weeks ago but it seems like it's broken in RN: https://github.com/facebook/react-native/issues/16152

tbergq:
closes #278

### DIFF
--- a/app/shared/src/Device.js
+++ b/app/shared/src/Device.js
@@ -2,30 +2,43 @@
 
 import { Dimensions, Platform } from 'react-native';
 
-function renewDimensions() {
-  try {
-    return Dimensions.get('screen');
-  } catch (error) {
-    // screen is not available in the 'test' environment
-    return { width: 0, height: 0 };
-  }
-}
+type ScreenDimensions = {
+  width: number,
+  height: number,
+};
 
 export default {
+  dimensions: {
+    width: 0,
+    height: 0,
+  },
+  renewDimensions() {
+    try {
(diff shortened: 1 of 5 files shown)


app/shared/src/Device.js
```diff
@@ -2,39 +2,55 @@
...
   },
   isLandscape() {
-    const { height, width } = renewDimensions();
+    const { height, width } = this.renewDimensions();
     return width >= height;
   },
   isTablet() {
-    if (Platform.OS === 'ios' && Platform.isPad) {
-      return true;
-    }
```
mrtnzlml: Why did you remove this condition? Maybe we should refactor it a little bit more - tablet is still tablet even though it's in the multitasking mode and it's a mistake we are adjusting the design based this I think.

tbergq: Maybe, now isTable is used to decide if we should show map with hotel list, render map view icon in header etc. Maybe we should add another method to decide if we should show map with hotel list?

mrtnzlml: The logic of showing the icon is wrong. It doesn't depend on the tablet/phone but on the available width. Let me think about it for a while. I am not very happy with this implementation yet... 🙂

PR #288, opened 2018-02-22
Script for publish workspaces

tbergq:


mrtnzlml:
I think you should revert back all this:

```diff
- "version": "0.0.2",
+ "version": "0.0.0",
- "private": false,
+ "private": true,
```

and unpublish all the helper packages. Correct?

tbergq:
Hmm, I think maybe unpublishing the helper packages will fail the download of the hotels package. Since these helper packages listed as dependencies of hotels package. 

- Could this be solved as having all dependencies as devDependencies in the hotel package?
- Should we create a new workspace without dependencies where we put the build files, and just publish this package?

mrtnzlml:
I think the best would be to have new `package.json` for this:

```
{
  "name": "@kiwicom/react-native-app-hotels",
  "version": "0.0.2",
  "private": false,
  "dependencies": {
    "react-native": "^132"
  }
}
```

It's because we are basically creating completely independent library they need to just require and use. Therefore, you don't need dependencies at all (expect RN). I would not create new workspace for this though. Just make it part of the build folder?

mrtnzlml:
Integrated, thanks.

PR #289, opened 2018-02-23
Use native translations

### LINKED ISSUE #268: Use native translations library
- [ ] Android
- [x] iOS

Idea: _maybe_ we can add translations native code call into `<Text />` component from our shared package because every text needs to be wrapped into this component anyway.

tbergq:
closes #268

### DIFF
--- a/app/shared/src/translations/__tests__/Translate.test.js
+++ b/app/shared/src/translations/__tests__/Translate.test.js
@@ -1,0 +1,13 @@
+// @flow
+
+import { NativeModules } from 'react-native';
+
+import Translate from '../Translate';
+
+describe('Translate', () => {
+  it('should call NativeModules.RNTranslationManager.translate with the key paramter', () => {
+    const string = Translate('test');
+    expect(string).toEqual('test');
+    expect(NativeModules.RNTranslationManager.translate).toBeCalledWith('test');
+  });
+});
--- a/android/app/src/main/java/com/reactnativeapp/MainApplication.java
+++ b/android/app/src/main/java/com/reactnativeapp/MainApplication.java
@@ -11,12 +11,16 @@ import com.kiwi.mynativemodule.MyNativeModulePackage;
 import com.oblador.vectoricons.VectorIconsPackage;
 import com.airbnb.android.react.maps.MapsPackage;
 import com.trinerdis.skypicker.logging.RNLoggingPackage;
+import com.trinerdis.skypicker.translation.*;
+
+import org.jetbrains.annotations.NotNull;
 
 import java.util.Arrays;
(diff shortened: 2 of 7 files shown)


app/shared/src/services/Translate.js
```diff
@@ -0,0 +1,7 @@
+// @flow
```
mrtnzlml: For the future reference: I would prefer to think about better structure instead of adding everything to the `services`. It's like a bag for everything. I think something like `shared/translations` or even new package for translations is a little bit better approach. What do you think?

tbergq: I thank that makes more sense, I will implement that structure

app/shared/src/services/__tests__/Translate.test.js
```diff
@@ -0,0 +1,16 @@
+// @flow
+import { NativeModules } from 'react-native';
```
mrtnzlml: My OCD is going crazy every time I see import under `// @flow` without an empty line. I know - it should not be in the code review but in eslint somehow... 😄

app/shared/src/services/Translate.js
mrtnzlml: Now or in the future? Should I merge it or wait?

tbergq: I will do it now for translations

app/shared/src/translations/__tests__/Translate.test.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+
+import { NativeModules } from 'react-native';
+
+import Translate from '../Translate';
+
+NativeModules.RNTranslationManager = {
+  translate: jest.fn(key => key),
+};
```
mrtnzlml: Sorry, one last thing - shouldn't this be part of the `jestSetup` script?

app/shared/src/translations/__tests__/Translate.test.js
```diff
@@ -0,0 +1,17 @@
...
+
+describe('Translate', () => {
+  it('should call NativeModules.TranslationManager.translate with the key paramter', () => {
```
mrtnzlml: OCD trigger:

mrtnzlml: OCD trigger:

mrtnzlml:
```

app/shared/src/translations/__tests__/Translate.test.js
mrtnzlml: Thanks. It will actually happen very soon because `Text` component will require the translations...

PR #295, opened 2018-02-26
Add currency native module

tbergq:


android/app/src/main/java/com/reactnativeapp/MainApplication.java
```diff
@@ -19,9 +21,10 @@
 import java.util.Arrays;
 import java.util.List;
 
-public class MainApplication extends Application implements ReactApplication, ResourceStringCallback {
+public class MainApplication extends Application implements ReactApplication, ResourceStringCallback, CurrencyChangeCallback {
 
     RNTranslationManagerPackage translationPackage = new RNTranslationManagerPackage(this);
+    RNCurrencyManagerPackage currencyPackage = new RNCurrencyManagerPackage(this);
```
mrtnzlml: Why do we have to do it like this?

android/app/src/main/java/com/reactnativeapp/MainApplication.java
```diff
@@ -63,4 +67,10 @@ public void onCreate() {
     public String getTranslation(@NotNull String key) {
         return key;
     }
+
+    @NotNull
+    @Override
+    public String getFormatCurrency(double amountInEur, @NotNull String currency) {
+        return amountInEur + " " +  currency;
```
mrtnzlml: Please return an object with these values as we discussed.

android/app/src/main/java/com/reactnativeapp/MainApplication.java
tbergq: Because the native module we get is just a mock, we have to add our actual own implementation 🥇

app/shared/src/services/__tests__/CurrencyFormatter.test.js
```diff
@@ -1,9 +0,0 @@
-// @flow
-
-import CurrencyFormatter from '../CurrencyFormatter';
-
-it('works', () => {
-  expect(CurrencyFormatter(10)).toBe('10');
-  expect(CurrencyFormatter(12.3)).toBe('12.3');
-  expect(CurrencyFormatter(72.8499999)).toBe('72.85');
```
mrtnzlml: Please try not to delete tests like this. For example, this one is quite important because it was created after bugfix (fixing the .84999999 rounding). Could you revert it, please?

tbergq: Add them to `app/shared/src/currency/__tests__/CurrencyFormatter.test.js`

PR #296, opened 2018-02-26
Add documentation for native modules

### LINKED ISSUE #284: Documentation: add info about native modules
Add info about what native modules we can use and how to use them. Most of the work will be probably in the private repo containing native modules but we should mention it here as well.

tbergq:
closes #284


mrtnzlml:
Awesome. Now just let's keep it updated. We'll split this README into proper docs in the future... :)

PR #301, opened 2018-02-27
Add support for multitasking on iOS devices

mrtnzlml:
There is one disadvantage of this solution - the screen width is by
design asynchronous value so you always have to use `AdaptableLayout`
component. Things like `Device.getWidth` will never work because this
value won't get updated. This is unfortunate consequence of this
workaround because we are using `onLayout` at the very top of this
application and the elements must be rendered first in order to get this
 value.


app/hotels/src/map/allHotels/HotelSwipeList.js
```diff
@@ -94,29 +68,40 @@ class HotelSwipeList extends React.Component<Props, State> {
...
     const { data, selectedIndex, onSnapToItem } = this.props;
 
+    const child = (
+      <BottomSheet openHeight={openHeight} closedHeight={closedHeight}>
+        <BottomSheetHandle />
+        <View style={styles.sliderWrapper}>
+          <Carousel
+            data={data}
+            renderItem={this.renderItem}
+            sliderWidth={Device.getWideDeviceThreshold()}
```
mrtnzlml: This is still not perfect but I don't know how to solve it better now.

app/shared/src/view/AdaptableLayout.js
```diff
@@ -0,0 +1,45 @@
...
+    if (this.state.wideLayout === true) {
+      if (this.props.renderOnWide) {
+        return this.props.renderOnWide;
+      }
+    } else {
+      if (this.props.renderOnNarrow) {
+        return this.props.renderOnNarrow;
+      }
+    }
+
```
tbergq: Could this be instead
```
if (this.state.wideLayout && this.props.renderOnWide) {
  ...
 } else if (!this.state.wideLayout && this.props.renderOnNarrow) {
...
}
```

mrtnzlml: Improved.

PR #305, opened 2018-02-28
Search with coordinates

tbergq:


app/hotels/src/allHotels/AllHotelsSearch.js
```diff
@@ -58,13 +59,49 @@ export class AllHotelsSearch extends React.Component<Props> {
...
+  hasCoordinates = () => {
+    const latitude = idx(this.props, _ => _.coordinates.latitude);
+    const longitude = idx(this.props, _ => _.coordinates.longitude);
 
     return (
-      Boolean(this.getCityIdFromData(data)) &&
-      s.checkin !== null &&
-      s.checkout !== null
+      latitude !== null &&
+      isUndefined(latitude) === false &&
```
mrtnzlml: Come on, I don't think you need Lodash function for this, do you? 😊  https://github.com/lodash/lodash/blob/c1f805f4972843b675056b2786f1165f7db81737/isUndefined.js#L17

Also, sounds like you are looking for `latitude != null` which handles `null` and `undefined` only?

PR #310, opened 2018-03-01
Fix android back button to work inside native app

### LINKED ISSUE #309: Clicking android back button from native app leaves entire app
- Start native app like `react-native-app-connected` in android
- Navigate to hotels
- Navigate to maps
- Click android back button
- You end up in native app, expected to end up in hotels search list

tbergq:
closes #309


app/hotels/package.json
```diff
@@ -7,11 +7,14 @@
     "@kiwicom/react-native-app-config": "^0",
     "@kiwicom/react-native-app-redux": "^0",
     "@kiwicom/react-native-app-shared": "^0",
+    "enzyme": "^3.3.0",
+    "enzyme-adapter-react-16": "^1.1.1",
```
mrtnzlml: Should be part of the root package JSON if you decided to use Enzyme for this project.

tbergq: Ok, I will create an issue to move it to root package

app/hotels/package.json
```diff
@@ -7,11 +7,14 @@
...
     "geolib": "2.0.24",
     "idx": "2.2.0",
     "lodash": "^4.17.4",
     "moment": "^2.20.1",
     "querystring": "^0.2.0",
+    "react-dom": "^16.2.0",
```
mrtnzlml: 3 questions... 😊 

1. Why do we need it? We do not work with DOM as you know it. Sounds odd.
2. Why 16.2? We are using React 16.0. (we have the same error with `react-test-renderer`)

tbergq: 1. Enzyme needs it
2. I could probably downgrade it
3. I will move it to root

app/hotels/src/navigation/AllHotelsNavigationScreen.js
```diff
@@ -87,7 +89,19 @@ class AllHotelsNavigationScreen extends React.Component<Props> {
...
+    if (Platform.OS === 'android' && this.backButtonListner === null) {
+      this.backButtonListner = BackHandler.addEventListener(
+        'hardwareBackPress',
+        () => {
+          this.props.onBackClicked();
+          return true;
+        },
+      );
+    }
+  };
```
mrtnzlml: Just out of curiosity - why does it work in our development environment? It looks like you are workarounding something you should not. Do they extend `ReactActivity`?

tbergq: Yes, in their activity they have to define
```
@Override
    public void onBackPressed() {

        if (mReactInstanceManager != null) {
            mReactInstanceManager.onBackPressed();
        } else {
            super.onBackPressed();
        }
    }
```

Which basically lets react-native handle navigation. When back button is pressed in the root view, react-native has no more views to go back to. Nothing happens when clicking back button, so we have to trigger the onBackClick callback.

tbergq: Which means that clicking the android backbutton leads back to this
```
onBackClicked = () => {
    if (NativeModules.RNNavigationModule) {
      NativeModules.RNNavigationModule.leaveHotels();
    } else {
      this.props.onBackClicked();
    }
  };
```
which is why it works in development

app/hotels/src/navigation/__tests__/AllHotelsNavigationScreen.test.js
```diff
@@ -0,0 +1,59 @@
+// @flow
+
+import * as React from 'react';
+import { Platform, BackHandler } from 'react-native';
+import { shallow } from 'enzyme';
```
mrtnzlml: Just `react-test-renderer/shallow` is not good enough?

app/hotels/package.json
mrtnzlml: > Enzyme needs it

app/hotels/src/navigation/__tests__/AllHotelsNavigationScreen.test.js
tbergq: I crashes, don't know why. I struggle finding documentation for react-test-renderer

mrtnzlml: Example from this repo: https://github.com/kiwicom/react-native-app/blob/2c9c102e3d65135a19db0a671dea619b2e1937dd/app/hotels/src/filter/stars/__tests__/StarsPopup.test.js#L4

Docs: https://reactjs.org/docs/shallow-renderer.html

tbergq: Sure, I will have a look at it.....

tbergq: The real issue here is that I want to test `componentDidMount` function. 
Using `react-test-renderer` crashes because a child component connects to redux store. I know with enzyme, I can pass the store like `{ context: { store } }`
But that does not work with `react-test-renderer`; 

with `react-test-renderer/shallow` it seems like I just get a snapshot (again cannot find any documentation), I can not get an instance and call `componentDidMount`

`enzyme shallow` allows me to get an instance and test `componentDidMount`

So if you have some hints or tips that can help me solve this, i would be happy :)

mrtnzlml: > Using react-test-renderer crashes because a child component connects to redux 

Test only what you need to test and don't involve Redux. I usually export default component with HOC components and I also export the component itself just to get access to the class directly.

I am not sure if `componentDidMount` works with shallow rendering or you have to use normal renderer. But again - I usually test only what I want to test so no need to call lifecycle methods.

It's just odd that you have to pass context to the Enzym (why?) and that you need React DOM reconciler for this.

tbergq: The problem is that `react-test-renderer` calls render, and renders all child components also. I do test the exported component like `import { AllHotelsNavigationScreen } from '../AllHotelsNavigationScreen'` still it gives me `Invariant Violation: Could not find "store" in either the context or props of "Connect(FilterStripe)"` because a child of the component connects to redux store.

As for `react-dom` there is an open issue for it, https://github.com/airbnb/enzyme/issues/1390

tbergq:
```
let backButtonListner = null;

export const registerListener = (onBackClicked: Function) => {
  if (Platform.OS === 'android' && this.backButtonListner === null) {
    backButtonListner = BackHandler.addEventListener(
      'hardwareBackPress',
      () => {
        onBackClicked();
        return true;
      },
    );
  }
};

export class AllHotelsNavigationScreen extends React.Component<Props> {
  

  componentDidMount = () => {
    registerListener(this.props.onBackClicked);
  };
```

Maybe  this for a workaround? So I can test the `registerListener` isolated?

mrtnzlml: Why do you even need to render it? :)

PR #311, opened 2018-03-01
Search with coordinates from map view

### LINKED ISSUE #299: Passing parameters to hotel search screen
_copy-pasted from the specs (section Passing parameters to hotel search screen)_

Search results should be opened in one of following modes:

- for active booking (destination city coordinates should be provided)
- for explore destination and airport (destination city coordinates should be provided)
- for explore my surroundings (actual location coordinates should be provided)
- location unknown

Hotels search results screen should accept parameter with location (latitude + longitude). If such parameters are passed, results should be displayed for accordingly and search input text should be nearest city (city specified for nearest hotel). If no parameters are provided, hotel search dialog should be empty and wait for user input.

---

It's probably going to be necessary to do it in multiple PRs. Please. I don't know what changes are necessary at this moment.

tbergq:
closes #299


app/hotels/src/search/__tests__/SearchQueryHelpers.test.js
```diff
@@ -0,0 +1,87 @@
...
+        hasCoordinates({
+          latitude: 34.3,
+          longitude: 12.12312,
+        }),
+      ).toBe(true);
+    });
+
+    it('returns false if latitude is not defined', () => {
+      expect(
+        // $FlowExpectedError: Expect error since it breaks with flow type
```
mrtnzlml: Of course it breaks with flow type - this is why you need to use the `// $FlowExpectedError` comment. It's usually better to explain why. Especially why you decided not to fix it here or why you are OK with this hack.

Look at this from my point of view - I don't know what is the reason for this. So I have to remove the comment, run Flow and only after that I'll realize it's because `hasCoordinates` accepts object with all fields and you are intentionally not passing them (you can improve the readability with `latitute: undefined` to make it really obvious).

Just a tip - think about it... :)

PR #315, opened 2018-03-02
Remove enzyme dependency

### LINKED ISSUE #313: Move enzyme and react-dom to root package.json

tbergq:
closes #313


app/hotels/src/navigation/__tests__/AllHotelsNavigationScreen.test.js
```diff
@@ -1,37 +1,22 @@
...
+  resetBackButtonListener,
+} from '../AllHotelsNavigationScreen';
 
 describe('AllHotelsNavigationScreen', () => {
+  afterEach(() => {
+    resetBackButtonListener();
+  });
   it('registers an event listener if platform is android', () => {
     const onBackClicked = jest.fn();
     Platform.OS = 'android';
```
mrtnzlml: I didn't notice this before but it's necessary to reset `OS` back when we overwite global variable.

app/hotels/src/navigation/AllHotelsNavigationScreen.js
```diff
@@ -59,8 +59,26 @@ type NavigationProps = {|
...
+      'hardwareBackPress',
+      () => {
+        onBackClicked();
+        return true;
+      },
+    );
+  }
+};
+
+// Needed for test purposes
```
mrtnzlml: It's almost always wrong to do something like this. Tests should guide you towards better application design but it doesn't mean you should tweak application code in order do be able to test it in this ugly way.. 🙂

mrtnzlml:
Integrated, thanks.

PR #322, opened 2018-03-02
Show persons and rooms on book now button

### LINKED ISSUE #319: Improve text on book now button
[Trello](https://trello.com/c/fB6jWpxI/112-bug-inconsistency-in-prices-on-search-and-in-hotel-booking)
[Zeplin](https://app.zeplin.io/project/5a1308cdae8ae89cae5c15af/screen/5a1edf0356c752b71a613634)

tbergq:
closes #319

### DIFF
--- a/app/hotels/src/singleHotel/bookNow/BookNow.js
+++ b/app/hotels/src/singleHotel/bookNow/BookNow.js
@@ -6,12 +6,12 @@ import { View } from 'react-native';
 import {
   StyleSheet,
   Color,
-  Price,
   TouchableItem,
-  ButtonText,
+  Price,
 } from '@kiwicom/react-native-app-shared';
 import idx from 'idx';
 
+import BookNowText from './BookNowText';
 import countBookingPrice from './countBookingPrice';
 import convertRooms from './convertRooms';
 import type { BookNow_availableRooms } from './__generated__/BookNow_availableRooms.graphql';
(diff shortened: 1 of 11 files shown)


app/hotels/src/singleHotel/bookNow/BookNow.js
```diff
@@ -82,11 +96,16 @@ export class BookNow extends React.Component<Props> {
...
             <View style={styles.buttonInnerWrapper}>
               <Text style={styles.buttonTitle}>Book Now</Text>
-              <Price
-                style={styles.buttonPrice}
-                amount={price.amount}
-                currency={price.currency}
-              />
+              <View style={styles.priceWrapper}>
+                <Text style={styles.buttonPrice}>
+                  {this.getBookNowHelpText()}
```
mrtnzlml: I am still not sure about this pattern. I am not sure what is this rerendering behavior here but what do you think just from a design perspective - is it better to call functions like this or to create a component and encapsulate everything like we do everywhere else?

So I am trying to compare this:

mrtnzlml: I am still not sure about this pattern. I am not sure what is this rerendering behavior here but what do you think just from a design perspective - is it better to call functions like this or to create a component and encapsulate everything like we do everywhere else?

So I am trying to compare this:

```
<Text style={styles.buttonPrice}>
  {this.getBookNowHelpText()}
</Text>
```

and this basically:

```
<BookNowText price={<Price />} />
```

This component can be part of this component (you don't need to export/import it).

tbergq: Well, you made me think, and it does make sense to extract it to a separate component. :)

PR #328, opened 2018-03-05
Uppercase button text on android

### LINKED ISSUE #327: Buttons should be all caps
[Trello](https://trello.com/c/wXLEkEw1/170-buttons-should-be-all-caps)

tbergq:
closes #327

### DIFF
--- a/app/hotels/src/singleHotel/bookNow/BookNow.js
+++ b/app/hotels/src/singleHotel/bookNow/BookNow.js
@@ -7,8 +7,8 @@ import {
   StyleSheet,
   Color,
   Price,
-  Text,
   TouchableItem,
+  ButtonText,
 } from '@kiwicom/react-native-app-shared';
 import idx from 'idx';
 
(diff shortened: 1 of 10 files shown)


app/hotels/src/singleHotel/bookNow/BookNow.js
```diff
@@ -81,7 +82,9 @@ export class BookNow extends React.Component<Props> {
         <View style={styles.buttonWrapper}>
           <TouchableItem onPress={this.handleGoToPayment} activeOpacity={0.6}>
             <View style={styles.buttonInnerWrapper}>
-              <Text style={styles.buttonTitle}>Book Now</Text>
+              <Text style={styles.buttonTitle}>
+                {TransformButtonText('Book Now')}
```
mrtnzlml: Isn't better to use components? Something like `<ButtonText />` sounds better to me than functions again.

app/hotels/src/singleHotel/description/Description.js
```diff
@@ -48,11 +49,11 @@ const Link = ({
 );
 
 const renderTruncatedFooter = (handlePress: () => void) => (
-  <Link label="Show More" handlePress={handlePress} />
+  <Link label={TransformButtonText('Show More')} handlePress={handlePress} />
```
mrtnzlml: Does this issue apply even to these links? I am not sure about that. I thought it's only about real buttons?

app/shared/src/buttons/__tests__/TransformButtonText.test.js
```diff
@@ -0,0 +1,19 @@
+// @flow
+
+import { Platform } from 'react-native';
+
+import TransformButtonText from '../TransformButtonText';
+
+describe('TransformButtonText', () => {
+  it('changes text to upper case on android', () => {
+    Platform.OS = 'android';
```
mrtnzlml: You are leaking changed `Platform.OS` value to other tests. This is potentially dangerous...

Issue #246, opened 2018-02-08
Create wrapper (component?) around Moment library


mrtnzlml:
See: https://github.com/kiwicom/react-native-app/pull/243#issuecomment-364096793

PR #333, opened 2018-03-05
Hotels: add styles to the search datepickers

mrtnzlml:
See: https://trello.com/c/TYCGDaYg/129-bug-date-pickers-are-not-optimized-for-small-devices

Note: I am not using correct icons because I need to update Icon
component first. This change will follow.

### DIFF
--- a/app/hotels/src/allHotels/searchForm/DateInput.js
+++ b/app/hotels/src/allHotels/searchForm/DateInput.js
@@ -15,7 +15,7 @@ type Props = {|
 
 type DateInputType = 'checkin' | 'checkout';
 
-export const DISPLAY_DATE_FORMAT = 'DD/MM/YYYY';
+export const DISPLAY_DATE_FORMAT = 'MMM DD';
 
 export const checkinAndCheckoutToDate = (
   checkin: moment,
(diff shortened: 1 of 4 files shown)


app/hotels/src/allHotels/searchForm/DateInput.js
```diff
@@ -15,7 +15,7 @@ type Props = {|
 
 type DateInputType = 'checkin' | 'checkout';
 
-export const DISPLAY_DATE_FORMAT = 'DD/MM/YYYY';
+export const DISPLAY_DATE_FORMAT = 'MMM DD';
```
tbergq: If we are going to use this kind of date formatting, we also need to set 
```
import moment from 'moment/min/moment-with-locales`
const deviceLocale = //getfrom native code
moment.locale(deviceLocale);
```

 Since on my device with norwegian, it shows `Dec 18`,  while expected output would be `des. 18`;
Which probably just means that we need to do #246

tbergq: Seems like this is more of a `react-native-datepicker` issue

https://github.com/xgfe/react-native-datepicker/issues/71

mrtnzlml: I would prefer not to fix it now since we didn't do localization at all yet. What do you think?

PR #334, opened 2018-03-06
Adjust hotel detail preview styling

### LINKED ISSUE #331: WIP: Price on detail - map
[Trello](https://trello.com/c/tn3nxkyt/192-bug-price-on-detail-map)

tbergq:
closes #331

### DIFF
--- a/app/hotels/src/map/HotelDetailPreview.js
+++ b/app/hotels/src/map/HotelDetailPreview.js
@@ -37,21 +37,29 @@ const styles = StyleSheet.create({
   description: {
     flex: 1,
     justifyContent: 'space-between',
-    marginVertical: 3,
   },
   hotelName: {
     fontSize: 16,
     color: '#30363d',
+    android: {
+      lineHeight: 17,
+    },
   },
   metainfo: {
     color: Color.grey.$700,
     opacity: 0.8,
     fontSize: 12,
+    android: {
+      lineHeight: 13,
+    },
   },
   price: {
     color: Color.brand,
     fontWeight: 'bold',
     fontSize: 12,
+    android: {
+      lineHeight: 13,
+    },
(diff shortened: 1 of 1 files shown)


app/hotels/src/map/HotelDetailPreview.js
```diff
@@ -37,21 +37,29 @@ const styles = StyleSheet.create({
...
     flex: 1,
     justifyContent: 'space-between',
-    marginVertical: 3,
   },
   hotelName: {
-    fontSize: 16,
     color: '#30363d',
+    fontSize: 17,
+    android: {
+      lineHeight: 16,
```
mrtnzlml: I think this is no go. Line height cannot be smaller than font size on Android otherwise the characters are going to be cropped (characters like "ěščřžýáíé"). I was thinking about extending our PlatformStyleSheet with validation but we are still using this for +- buttons hack...

Same even for other lines.

tbergq: I can also set line height higher, this was just the values I found in Zeplin.

Let's see if we can find some better solution anyways...

mrtnzlml: I am ok with line-height solution just not with the number. Zeplin design is not very RN compatible... 🙂

mrtnzlml: See: https://trello.com/c/S69utn2g/143-hotel-name-on-detail or https://trello.com/c/ObSFWfrs/149-bug-description-string-in-search-is-not-fully-visible

tbergq: Ok, let me have a second look at it when they have fixed the maven issue, now I can't run the app on android...

PR #339, opened 2018-03-06
Hotels: make the whole search page scrollable

mrtnzlml:
The search form and search tags should be scrollable as well according
to the specification (not only search results).

TODO:

- [x] map of all hotels doesn't work


mrtnzlml:
@tbergq Could you please review this? I will still have to tweak it but the main idea is there.

app/hotels/src/allHotels/index.js
```diff
@@ -0,0 +1,161 @@
...
+      />
+    );
+  };
+
+  renderAllHotelsSearchPublicRenderer = (
+    rendererProps: allHotels_cityLookup_QueryResponse,
+  ) => {
+    const cityId = idx(rendererProps, _ => _.city.edges[0].node.id) || null;
+
+    // TODO: cannot query with non-existing cityId!
```
tbergq: Now typing in the location search bar closes the keyboard because component re-renders. I think you need to separate location from the SearchList, and rather just pass it the cityId, so re-render happens only when cityId changes.

mrtnzlml:
@tbergq Ready to check.

PR #340, opened 2018-03-07
Update price component and currency formatter

### LINKED ISSUE #338: Update Price component implementation
Mobile guys realized this implementation is way to complicated so they simplified the price formatting:

screen shot 2018-03-06 at 15 56 03

There is also new interface (bridges 0.1.19) for Java:

```
fun formatAmount(amount: Double, currency: String): String
```

and iOS: _probably not ready yet but in the same way?_

tbergq:
closes #338

### DIFF
--- a/app/shared/src/__tests__/Price.test.js
+++ b/app/shared/src/__tests__/Price.test.js
@@ -2,25 +2,30 @@
 
 import * as React from 'react';
 import ShallowRenderer from 'react-test-renderer/shallow';
+import testRenderer from 'react-test-renderer';
 
 import Price from '../Price';
+import CurrencyFormatter from '../currency/CurrencyFormatter';
+
+jest.mock('../currency/CurrencyFormatter', () => jest.fn());
 
 const renderer = new ShallowRenderer();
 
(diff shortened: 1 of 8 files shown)


app/shared/src/__tests__/Price.test.js
```diff
@@ -2,25 +2,30 @@
 
 import * as React from 'react';
 import ShallowRenderer from 'react-test-renderer/shallow';
+import testRenderer from 'react-test-renderer';
 
 import Price from '../Price';
+import CurrencyFormatter from '../currency/CurrencyFormatter';
 
-const renderer = new ShallowRenderer();
+jest.mock('../currency/CurrencyFormatter', () => jest.fn());
```
mrtnzlml: Why not to mock `RNCurrencyManager` like every other native module?

tbergq: The RNCurrencyManager mocks ´NativeModules.RNCurrencyManager`. Here I want to test that we actually call the module in `app/shared/src/currency/CurrencyFormatter`

PR #346, opened 2018-03-08
Add getDimensions to device and use it in gallery grid

### LINKED ISSUE #345: Gallery - Size change
[Trello](https://trello.com/c/FuRRnBka/198-bug-gallery-size-change)

tbergq:
closes #345


mrtnzlml:
Integrated, thanks.

I think we should replace this workaround (async device dimensions) completely with our custom native module. I hope that we can measure the true screen dimensions in native code without problems...

PR #347, opened 2018-03-08
Add roboto font

### LINKED ISSUE #307: Text: use custom font
We need to distribute our own font to prevent these issues on weird devices: https://github.com/facebook/react-native/issues/15114

System fonts:
- iOS: SanFran
- Android: Roboto

Related issue: https://trello.com/c/sYuG1zzB/159-bug-uncompleted-strings-and-layouts-in-hotels

tbergq:
closes #307 

Fonts are not beeing bundled as assets (same with `react-native-vector-icons` fonts)
I have added roboto font to `@kiwicom/react-native-native-modules` and extended configureApplication script with a script that copies all fonts to the correct android folder. 

Native needs to do the same on their side.


mrtnzlml:
@tbergq Do we have to fetch the font from the native modules? What about the approach introduced in #353?

tbergq:
@mrtnzlml No, we do not have to. We can just add it as you did in #353. 
The issue is that the font is not bundled together with the jsbundle when we release. 

This means that IOS and Android would have to manually add the font file as well. 

Since `react-native-native-modules` is our shared repo between native and react-native. I think it makes sense to put all custom fonts there.

PR #353, opened 2018-03-09
Shared: add new TextIcon component

### LINKED ISSUE #330: Date pickers are not optimized for small devices
[Trello](https://trello.com/c/TYCGDaYg/129-bug-date-pickers-are-not-optimized-for-small-devices)

mrtnzlml:
Closes #330


tbergq:
@mrtnzlml This font will not be bundled and deployed with the npm package. How will native guys add it to the app?

mrtnzlml:
@tbergq That is true and I didn't solve it yet. I believe they already have this font anyway. I would like to:

1. solve it in the future
2. move it to the different repo like you did with Roboto

Currently, I have no idea which way to go.

tbergq:
I think we should agree on how to to it before we merge. It should be solved in the same way in this MR and this #347.

mrtnzlml:
@tbergq I added new commands to copy static assets to the build folders. Please check it again.

PR #359, opened 2018-03-12
Fix invalid dates issue that arises when added to native app

tbergq:


mrtnzlml:
@tbergq Can you please help me to understand what was wrong? Why does it work for us and not for native guys? I remember I moved the default dates to the initial reducer state so what did I do wrong?

tbergq:
When I did some console logging, I saw that props.checkin and props.checkout were strings. I fail to understand why it works in develop, and not in the native app though.

PR #360, opened 2018-03-13
Hotels: refactor room picker button

mrtnzlml:
It now uses `TouchableItem` instead of `TouchableOpacity`, the high is
finally same for selected and not selected buttons and the design is
update according to the new design changes.

screen shot 2018-03-13 at 09 58 47


app/hotels/src/singleHotel/roomPicker/RoomPicker.js
```diff
@@ -71,34 +29,110 @@ export default function RoomPicker({
...
+  const buttonStyles = [styles.button, styles.borderRadiusRight];
+
+  if (disabled) {
+    return (
+      <View style={[...buttonStyles, styles.disabledButton]}>{child}</View>
+    );
+  }
+
+  return (
+    <TouchableItem style={buttonStyles} onPress={onPress}>
```
tbergq: There is a disabled prop `<TouchableItem style={buttonStyles} onPress={onPress} disabled={disabled}>`

mrtnzlml: Good point - even better. I just copy pasted the previous 💩 solution... :)

mrtnzlml: @tbergq Should be fixed.

PR #362, opened 2018-03-13
Single hotel header: use 'TouchableItem' component

mrtnzlml:
One small step to completely deprecate direct usage of touchable opacity
 and native feedback.


tbergq:
Should we restrict import in `eslintrc.js` also?

mrtnzlml:
@tbergq I would love to do that but it's not possible yet because we still use touchable opacity in some places.

PR #366, opened 2018-03-14
Create moment wrapper

### LINKED ISSUE #246: Create wrapper (component?) around Moment library

tbergq:
closes #246


app/hotels/src/GraphQLSanitizers.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 
-import moment from 'moment';
+import { moment } from '@kiwicom/react-native-app-shared';
```
mrtnzlml: @tbergq I think it's quite a good idea to create completely new packages if it makes sense ([not sure if our shared package is the right place](https://github.com/kiwicom/react-native-app/blob/master/app/shared/README.md)). What do you think?

app/shared/src/datetime/__tests__/__snapshots__/Date.test.js.snap
```diff
@@ -17,7 +17,7 @@ exports[`formats date 1`] = `
     ]
   }
 >
-  2017-11-17
+  11/17/2017
```
mrtnzlml: This test looks fragile. Will it break with a different setting? 🙂 Maybe the test is complete nonsense and we should rewrite it to actually test something (instead of snapshoting).

tbergq: Since `device-info` is mocked, I am not sure it actually returns anything, so moment falls back to default locale. I do think you are correct, this test really does not test anything right now

app/hotels/src/GraphQLSanitizers.js
tbergq: I think that sounds like a goodt idea, naming? just `@kiwicom/react-native-app-moment`?

app/shared/src/datetime/__tests__/__snapshots__/Date.test.js.snap
mrtnzlml: > this test really does not test anything right now

@tbergq Get rid of it.

app/hotels/src/GraphQLSanitizers.js
mrtnzlml: I actually wanted to hide the `moment` implementation detail. So I would maybe consider renaming `-translations` to `*-localization` and add it there with exports of just `DateFormatter` instead of `moment`? Something like that sounds good to me.

tbergq: Ok, adding it to `localization` makes sense. Not sure about naming it `DateFormatter`, We are basically just renaming `moment`. I think it makes sense when new people look at the code, that they actually understand at once that we are using `moment`

tbergq:
@mrtnzlml please have another look

app/shared/src/image/NetworkImage.js
```diff
@@ -2,7 +2,7 @@
 
 import * as React from 'react';
 import Image from 'react-native-image-progress';
-import { connect } from '@kiwicom/react-native-app-redux';
+import { connect } from 'react-redux'; // TODO: Find out why this crashes when imported from @kiwicom/react-native-app-redux
```
mrtnzlml: I don't want to merge this... :)

What problem is here? Is `connect` undefined? It may be a circular reference somewhere.

PR #367, opened 2018-03-14
Eslint: enforce usage of 'TouchableItem' component

mrtnzlml:


.eslintrc.js
```diff
@@ -39,7 +39,9 @@ module.exports = {
             importNames: [
               'Dimensions', // Device
               'StyleSheet', // StyleSheet
-              'Text' // Text
+              'Text', // Text
+              'TouchableNativeFeedback', // TouchableItem
+              'TouchableOpacity', // TouchableItem
```
tbergq: Should `Button` and `TouchableHighlight` also be here? They are also button components from react native.

mrtnzlml: You are right - updated. Check it, please.

PR #368, opened 2018-03-15
Use one instance of relay environments

### LINKED ISSUE #101: Use a single instance of the Relay Environment
Currently we create a new environment every time when we call `commitMutation` or when we render `QueryRenderer`. It's not too often, but still it's not a best practice.

We should use a single instance for the whole application.

tbergq:
closes #101


mrtnzlml:
Thanks - this is what I expected. Integrated.

PR #370, opened 2018-03-15
Check checkin date when application launches

### LINKED ISSUE #369: Checkin date can be before todays date
Checkin and checkout dates are stored by redux persist. Screenshot below is taken 14th of March

screenshot_20180314-174213

tbergq:
closes #369

### DIFF
--- a/app/hotels/src/allHotels/AllHotels.js
+++ b/app/hotels/src/allHotels/AllHotels.js
@@ -5,7 +5,7 @@ import { ScrollView } from 'react-native';
 import idx from 'idx';
 import { graphql } from 'react-relay';
 import { PublicApiRenderer } from '@kiwicom/react-native-app-relay';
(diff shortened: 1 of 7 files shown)


mrtnzlml:
I am afraid this solution is not good enough. For example, I broke it just by setting the date to today, changing the laptop date to tomorrow and restoring the app from the background. The value is still stuck in the UI and `componentDidMount` is not called. But something like this could be a better solution:

```js
componentDidMount = () => {
  AppState.addEventListener('change', state => {
    if (state === 'active') {
      // the app was restored from the background - refresh UI
    }
  });
}
```

tbergq:
@mrtnzlml Updated with a check also on app state change. We still need to check on componentDidMount since user can leave the app, date gets outdated, and he comes back to the hotels app from the native app.

mrtnzlml:
Integrated, thanks.

PR #374, opened 2018-03-16
Fix hotel title line break

tbergq:
Change this
image

to this

image


mrtnzlml:
Integrated, thanks. I just removed the wrapping view.

PR #376, opened 2018-03-16
Add date transform to redux persist

tbergq:
Checks all fields when going from Async store => redux and transforms date strings into date objects.


mrtnzlml:
Integrated, thanks.

PR #390, opened 2018-03-20
Increase room row margin

tbergq:


app/hotels/src/singleHotel/roomList/RoomRow.js
```diff
@@ -90,7 +90,7 @@ export class RoomRow extends React.Component<Props> {
     const selectedCount = idx(this.props.selected, _ => _[originalId]) || 0;
     const room = idx(availableRoom, _ => _.room);
     return (
-      <SimpleCard>
+      <SimpleCard style={{ marginVertical: 5 }}>
```
mrtnzlml: We should maybe start thinking about inline objects the same ways as about inline callback. I think it has the same effect - always a new instance of the object with every re-render. What do you think? (just an idea at this moment)

tbergq: I really wanted to do that, but the implementaition of SimpleCard does not work with it
```
return StyleSheet.create({
    wrapper: {
      backgroundColor: '#fff',
      marginVertical: 3,
      paddingVertical: 10,
      paddingHorizontal: 10,
      ...additionalWrapperStyles,
      ...extraStyles,
    },
  });
```
Can that maybe be fixed with `StyleSheet.flatten`?

mrtnzlml: Sure, don't deal with it now. We will have to fix few components and introduce new eslint rule. I am just interested in your opinion.

tbergq: Yes, I always prefer StyleSheet.create over inline styling

PR #392, opened 2018-03-20
Refactor read more to shared component apply correct read more styles in room info

tbergq:


.eslintrc.js
```diff
@@ -48,6 +48,11 @@ module.exports = {
             message:
               "Please use '@kiwicom/react-native-app-shared' package instead.",
           },
+          {
+            name: 'react-native-read-more-text',// ReadMore
```
mrtnzlml: OCD 😊

```diff
- 'react-native-read-more-text',// ReadMore
+ 'react-native-read-more-text', // ReadMore
```

app/hotels/src/singleHotel/hotelInformation/description/Description.js
```diff
@@ -91,8 +60,8 @@ export class Description extends React.Component<Props> {
       <View style={isWide ? styles.wideCardChildrenWrapper : null}>
         <ReadMore
           numberOfLines={5}
-          renderTruncatedFooter={renderTruncatedFooter}
-          renderRevealedFooter={renderRevealedFooter}
+          truncatedText="Show More"
+          relvealedText="Show Less"
```
mrtnzlml: Typo - should be `revealed`.

app/hotels/src/singleHotel/roomList/RoomDescription.js
```diff
@@ -28,16 +32,26 @@ const styles = StyleSheet.create({
...
   const description = idx(props, _ => _.room.description.text);
 
   if (description != null) {
     return (
       <View>
         <View style={styles.delimiter} />
-        <ReadMore numberOfLines={2}>
+        <ReadMore
+          numberOfLines={2}
+          linkStyle={styles.readMore}
```
mrtnzlml: Can this be just `style` as usual? I know there is also view around but you cannot style it from outside anyway.

app/hotels/src/singleHotel/roomList/__tests__/RoomDescription.test.js
```diff
@@ -0,0 +1,18 @@
+// @flow
+
+import * as React from 'react';
+import { PlaygroundRenderer } from '@kiwicom/react-native-app-playground';
+
+import { RoomDescription } from '../RoomDescription';
+
+describe('RoomDescription', () => {
```
mrtnzlml: Why do you use these describers? Isn't this obvious from the test name?

app/shared/src/Color.js
```diff
@@ -75,6 +75,18 @@ const ColorPalette = {
...
+ * from RNColors
+ *      "brand":           @"#0097A9",
+        "brandSecondary":  @"#0CB3C7",
+        "buttercup":       @"#EB9D08",
+        "sun":             @"#FBAD18",
+        "backgroundGray":  @"#EDEFF2",
+        "textDark":        @"#30363D",
+        "textMedium":      @"#596066",
+        "textLight":       @"#79818A",
+ */
```
mrtnzlml: Maybe we should remap it manually so we'll be sure that we are actually using the right name in the code and the autocomplete will work?

app/shared/src/ReadMore.js
```diff
@@ -0,0 +1,73 @@
+// @flow
+
+import * as React from 'react';
+import { View } from 'react-native';
+import OriginalReadMore from 'react-native-read-more-text'; // eslint-disable-line
```
mrtnzlml: Please use this format:

```js
// eslint-disable-line no-restricted-imports
```

It will catch also other rules and it's more clear what rule is disabled here.

app/hotels/src/singleHotel/roomList/__tests__/RoomDescription.test.js
tbergq: Guess this is my OCD, `it` looks so naked without a describe.

Just how I was taught writing tests really. I can skip them if you think it is better.

app/shared/src/Color.js
tbergq: Makes sense, it is not that many colors either. I will do it.

tbergq:
@mrtnzlml please check again

app/hotels/src/singleHotel/roomList/__tests__/RoomDescription.test.js
mrtnzlml: I like it more without the description because 1) it's less code and 2) I think it's redundant (verbose):

```
 PASS  app/hotels/src/singleHotel/roomList/__tests__/RoomDescription.test.js
  RoomDescription
    ✓ renders (7ms)
```

Failure (not verbose):

```
 FAIL  app/hotels/src/singleHotel/roomList/__tests__/RoomDescription.test.js
  ● RoomDescription › renders

    expect(value).toMatchSnapshot()
```

However, it's definitely not wrong and we don't have a rule for this so I will accept it. I just wanted to start this discussion. Thank you.

PR #393, opened 2018-03-20
Translations: translate all buttons

mrtnzlml:
This is just a first batch. Next step is to enforce every Text component.


app/shared/src/translations/Translate.js
```diff
@@ -1,7 +0,0 @@
-// @flow
-
-import { NativeModules } from 'react-native';
-
-export default function(key: string) {
```
tbergq: Not sure we should delete this entirely, What about cases like this? Or 3rd party components that expects strings.
```
return {
      headerLeft: renderHeaderLeft(),
      headerTitle: 'Hotels',
      headerRight: renderHeaderRight(),
    };
```

app/translations/index.js
```diff
@@ -0,0 +1,45 @@
...
+import Translation from './src/Translation';
+
+export default Translation;
+
+export type TranslationType =
+  | React.Element<typeof Translation>
+  | React.Element<typeof DummyTranslation>
+  | React.Element<typeof TranslationFragment>;
+
+export class TranslationFragment extends React.Component<{|
```
tbergq: Can we put this in its own file? If I do a file search for TranslationFragment, I wont find it.

app/translations/index.js
```diff
@@ -0,0 +1,45 @@
...
+|}> {
+  render = () => this.props.children;
+}
+
+/**
+ * This translation component only returns the ID back without translating
+ * it. It should therefore used in tests and other places where we don't
+ * want to deal with translations but the Flow requires `Translation`.
+ */
+export class DummyTranslation extends React.Component<{|
```
tbergq: Seperate file also please.

app/shared/src/translations/Translate.js
mrtnzlml: My idea behind these changes is to change Flow types to enforce the translations. So for example in case of `headerTitle` the type is not going to be `string` but `TranslationType`. This way you cannot really make a mistake. Do you know about any 3rd party library where we need to use string and there is no other solution (so I can see it)? Because I think [we can use component in case of `headerTitle`](https://reactnavigation.org/docs/stack-navigator.html#headertitle).

tbergq: > Do you know about any 3rd party library where we need to use string

Nothing comes to mind. I am sure this is ok. If need be, we can always make it work

mrtnzlml:
@tbergq I rebased it and moved all the components into separate files but I didn't revert the `Translate` method (let's solve it when needed). Please check it again.

PR #400, opened 2018-03-21
Un nest text for android on hotel title

tbergq:
Probably not the optimal solution, but the best we can do for the moment.

Text will not crash on android, but stars renders far to the right.

Looks good on ios.


mrtnzlml:
> Text will not crash on android, but stars renders far to the right.

I think we need to change the design (because of implementation difficulties). What about moving the stars above or below the hotel name?

tbergq:
Sounds doable. Maybe we should consult some designer?
In my head above sounds good.

tbergq:
Fixed with #401

PR #401, opened 2018-03-21
HotelTitle: remove line heights to allow free text expansion

mrtnzlml:
Line height doesn't work very well for nested text strings. So I removed
 them completely in order to improve the title design on Android. We
 should I general be very careful with line heights especially on
 Android devices.

Replaces: https://github.com/kiwicom/react-native-app/pull/400

@tbergq Please check it if everything is still alright (I don't care about line heights in Zeplin).


tbergq:
Looks good on both platforms

PR #404, opened 2018-03-22
Add location picker modal screen

tbergq:


app/core/src/navigation/HomepageStack.js
```diff
@@ -18,5 +19,22 @@ export default StackNavigator(
...
+  },
+);
+
+export default StackNavigator(
+  {
+    Main: {
+      screen: HomeStack,
+    },
+    LocationPicker: {
+      screen: withMappedProps(LocationPicker),
```
mrtnzlml: @tbergq This is only temporary, right? Will this be part of the hotels' navigation stack?

tbergq: Yes, that was the idea, when it is all done, we connect it to hotels navigation stack

app/shared/src/Color.js
```diff
@@ -100,4 +100,5 @@ export default {
   textLight,
 
   ...ColorPalette,
+  white: '#fff',
```
mrtnzlml: This is already in the `ColorPalette`.

app/hotels/src/allHotels/searchForm/locationPicker/LocationPickerScreen.js
```diff
@@ -0,0 +1,185 @@
...
+    color: Color.white,
+    lineHeight: 20,
+    fontFamily: 'SFProText-Regular',
+  },
+  confirmButtonText: {
+    fontFamily: 'SFProText-Semibold',
+  },
+  headerTitle: {
+    color: Color.white,
+    fontFamily: 'SFProText-Semibold',
```
mrtnzlml: What about Android?

app/hotels/src/allHotels/searchForm/locationPicker/LocationPickerScreen.js
```diff
@@ -0,0 +1,185 @@
...
+          borderlessRipple
+        >
+          <Text style={styles.headerButtonText}>Confirm</Text>
+        </Touchable>
+      );
+    }
+
+    return {
+      headerLeft: headerLeft(),
+      title: <Text style={styles.headerTitle}>Where</Text>,
```
mrtnzlml: Why do you use the `Text` component here? Our default setting looks alright...

app/hotels/src/allHotels/searchForm/locationPicker/LocationPickerScreen.js
tbergq: There really is no design for android. The font will work on android, since it is custom font. Should I rather use default font on android?

app/hotels/src/allHotels/searchForm/locationPicker/LocationPickerScreen.js
tbergq: It is a custom font, default is SFUI, this header has SFProText-Semibold,

app/hotels/src/allHotels/searchForm/locationPicker/LocationPickerScreen.js
mrtnzlml: It's definitely not SF* font on Android.

app/hotels/src/allHotels/searchForm/locationPicker/LocationPickerScreen.js
mrtnzlml: Seems like a design bug to me - why to complicate it? I would prefer to have one title font everywhere...

app/hotels/src/allHotels/searchForm/locationPicker/LocationPickerScreen.js
tbergq: Ok, I will change to default font then

app/hotels/src/allHotels/searchForm/locationPicker/LocationPickerScreen.js
tbergq: Ok, I will leave as the others

app/hotels/src/allHotels/searchForm/locationPicker/LocationPickerScreen.js
mrtnzlml: Well, at least setup this only for iOS. We use Roboto on Android.

PR #409, opened 2018-03-23
Hotel location suggestions: make the search case insensitive

mrtnzlml:


app/hotels/src/allHotels/searchForm/locationPicker/SuggestionListItem.js
```diff
@@ -43,25 +43,31 @@ export default class SuggestionListItem extends React.Component<Props> {
...
+    const [, before, match, after] = matchResult;
+    return [before, match, after];
   };
 
   render = () => {
-    const { before, match, after } = this.splitText();
+    const [before, match, after] = this.findStringToHighlight(
+      idx(this.props, _ => _.city.name) || '',
+      this.props.search,
+    );
```
tbergq: Why do you pass these strings to to class method, they are available in that method as well

mrtnzlml: Because I believe it should be obvious what is happening - I was very confused by `this.splitString()` because the interface is not clear what string I am actually splitting. It also simplifies the test. Should I revert it?

tbergq: I like the renaming of the function and the regex. I get the benefit for the readability of the testing. 

we could do
```
findStringToHighlight = (): [string, string, string] => {
    const fullText = idx(this.props, _ => _.city.name) || '';
    const stringToHighlight = this.props.search;
...
}
```

But if you think it is best the way you have it in this PR, I am fine with it.

mrtnzlml: @tbergq Improved.

PR #410, opened 2018-03-23
Replace react-native-device-info with react-native-native-modules

tbergq:


app/translations/index.js
```diff
@@ -6,8 +6,14 @@ import Translation from './src/Translation';
...
 import TranslationFragment from './src/TranslationFragment';
 import DateFormatter from './src/DateFormatter';
+import GetDeviceLocale from './src/GetDeviceLocale';
 
-export { DummyTranslation, TranslationFragment, DateFormatter };
+export {
+  DummyTranslation,
+  TranslationFragment,
+  DateFormatter,
+  GetDeviceLocale,
```
mrtnzlml: Is it necessary to make it public? I would prefer to keep it private as long as possible...

tbergq:
@mrtnzlml updated, keeping it local.

PR #411, opened 2018-03-26
Add translations in location picker

tbergq:


tbergq:
@mrtnzlml please check again

PR #414, opened 2018-03-26
WIP: Pick first location on confirm if none is selected

tbergq:


tbergq:
I would like to wait a little bit with this one. I think the problem where the styling of the confirm button does not update is linked to our text component. I thought it might be a rerender issue with `react-navigation`

mrtnzlml:
@tbergq Can you please replicate this issue in a minimal example so we can fix it?

tbergq:
Yes, I will look into it soon, make a playground for it, and try to solve it

tbergq:
closing this since we confirm on click on first location

PR #417, opened 2018-03-26
Connect location picker

### LINKED ISSUE #402: Suggestion modal
[Zeplin](https://zpl.io/bzYogR4)

tbergq:
closes #402

### DIFF
--- a/app/core/src/navigation/HomepageStack.js
+++ b/app/core/src/navigation/HomepageStack.js
@@ -5,9 +5,8 @@ import { StackNavigator } from 'react-navigation';
 import { StackNavigatorOptions } from '@kiwicom/react-native-app-navigation';
 
 import Homepage from '../screens/homepage/Homepage';
-import LocationPicker from '../../../hotels/src/allHotels/searchForm/locationPicker/LocationPickerScreen';
 
-const HomeStack = StackNavigator(
+export default StackNavigator(
   {
     Home: {
       screen: withMappedProps(Homepage),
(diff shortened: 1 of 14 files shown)


tbergq:
I ran in to some issue when connection it to the hotels stack. Suddenly calls to `setParams` resulted in undefined values. I solved this by connecting it to redux instead

mrtnzlml:
@tbergq There are still 3 issues I don't like from the UX perspective:

- the "Where" input (weird placeholder BTW) is not focused by default
- I believe it's better to click to "Rome" and this should select and close the modal - now it's little bit odd: you type "Rome", you see "Rome", you click "Rome" and 1) nothing happens but you'll lost all the suggestions and 2) you need to click "Confirm" even though this is what I expected by clicking "Rome"
- if you open the modal for the second time and you already searched for Rome it will open with empty "Where" input again - this is unexpected from my point of view

For this reason, I would prefer not to replace it yet. Maybe just create something like `if (experimentalFeaturesEnabled)` otherwise let's stick with the old solution until we'll make the basics good.

tbergq:
As for the first, I do agree, but it is according to design 🙈 , and we can make it focuesed, that is an easy fix. 
Second, i absolutely agree, and this is also how other parts of the native app, I say we just implement it like this. 
Third, I can also fix this

I will fix these issues tomorrow.

tbergq:
@mrtnzlml UX improvements implemented. Please check again

PR #423, opened 2018-03-27
Text: improve styles inheritance to be RN compatible

mrtnzlml:
Now I probably finally understand how does it work. Our default style
must be applied on the root (top) Text component only and inner deeply
nested components are inheriting previous styles only if their style is
undefined... :)

@tbergq Please also try your use-case. I rethought the internal implementation completely so now it behaves the same as RN implementation and it should just work. (I also don't know how to unit test this style changes on state change)


tbergq:
I can confirm that it works for my use case as well

mrtnzlml:
@tbergq Cool. We also talked about the fact that our Text behaves differently from RN Text component. This is now resolved and it's no longer true. The only exception is that we are adding our default styles. Please, let me know if you spot any inconsistency.

PR #427, opened 2018-03-27
Translate all navigation titles

mrtnzlml:
Closes: https://github.com/kiwicom/react-native-app/issues/285

Now we just need to prepare Phraseapp deployment.

### DIFF
--- a/app/navigation/index.js
+++ b/app/navigation/index.js
@@ -1,20 +1,21 @@
 // @flow
 
+import * as React from 'react';
 import { Platform, StatusBar } from 'react-native';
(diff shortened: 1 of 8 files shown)


app/navigation/index.js
```diff
@@ -28,6 +24,43 @@ const createNavigationOptions = () => {
   return navigationOptions;
 };
 
+export const HeaderTitle = ({ children }: {| children: TranslationType |}) => {
+  return (
+    <Text
+      style={{
+        color: '#fff',
```
tbergq: Use `StyleSheet.create({})` instead

mrtnzlml: fixed

PR #430, opened 2018-03-27
Dependencies: upgrade Relay to the latest version

mrtnzlml:
@tbergq I need your help here. Please check the [failing tests](https://circleci.com/gh/kiwicom/react-native-app/832). The problem is that Relay generates additional `$refType` prop (I don't know why exactly) but we are not using it in our manual tests. How to solve it?


tbergq:
I have not been able to find out why this gives errors either. Maybe we could register an issue with flow(maybe there is a bug, maybe they can help us resolve it), and use `$FlowIssue:` in the meanwhile, since they are just tests. Or wait with the upgrade until we know the answer.

mrtnzlml:
> I have not been able to find out why this gives errors either.

It's because it generates exact types like this:

```
type XYZ = {|
  +$refType: AllHotelsSearchRow$ref,
|}
```

and we are not using this property in our manual Relay data mocks. Now you understand it so my question is how to test our components... :)

mrtnzlml:
@tbergq Please check it again. Note this issue: https://github.com/facebook/relay/issues/2394

PR #435, opened 2018-03-28
Map language locales to booking.com language

### LINKED ISSUE #434: Map from language locales to booking.com language
locale can be passed as `nb-NO` but booking.com expects `no`
image uploaded from ios

tbergq:
closes #434

### DIFF
--- a/app/translations/src/MapLocaleToLanguageQuery.js
+++ b/app/translations/src/MapLocaleToLanguageQuery.js
@@ -1,0 +1,67 @@
+// @flow
+
+import idx from 'idx';
+
+// Copied from grahpql.schema
+const languages = [
+  'ar',
+  'bg',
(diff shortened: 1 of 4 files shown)


app/translations/src/MapLocaleToLanguageQuery.js
```diff
@@ -0,0 +1,67 @@
+// @flow
+
+import idx from 'idx';
+
+// Copied from grahpql.schema
```
mrtnzlml: Should we update our GraphQL API to be less embarrassing? Or how to keep this up to date?

mrtnzlml: For example, we have now Locale and Language types in the API. What a mess... :-/

PR #436, opened 2018-03-28
Translations: fix transformation prop propagation

mrtnzlml:
This props was used only in Translation component but it should be used
everywhere. Moreover the TranslationFragment should pass all props down.


mrtnzlml:
@tbergq Huh? That was extremely fast... 😄

PR #439, opened 2018-03-29
Upgrade rn version

### LINKED ISSUE #433: Upgrade React Native to the latest version

tbergq:
closes #433 
We need to let ios developers know about the changes in Podfile
also note that `react-native-maps` does only work with `react-native: 0.54.0-rc.3`
https://github.com/react-community/react-native-maps/issues/2051#issuecomment-370119631

There still is some cropped texts that needs to be adjusted, there are a lot of `componentWillMount` warnings, mainly from react-native itself, and other 3rd party modules. We also have som `componentWillReceiveProps` that should be refactored.


tbergq:
@mrtnzlml ready to be review, please have a look for some UI peculiarities I might have missed

PR #442, opened 2018-03-29
Single hotel page: make navigation header transparent on Android

mrtnzlml:
@tbergq I am not really confident about the implementation. I had to combine 3 designs... 😢

### DIFF
--- a/app/hotels/src/navigation/NavigationStack.js
+++ b/app/hotels/src/navigation/NavigationStack.js
@@ -1,6 +1,7 @@
 // @flow
 
 import * as React from 'react';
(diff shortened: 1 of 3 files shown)


app/hotels/src/navigation/NavigationStack.js
```diff
@@ -51,6 +53,25 @@ const AllHotelsStack = StackNavigator(
   },
 );
 
+const createTransparentHeaderStyle = () => {
+  if (Device.isWideLayout()) {
+    return {};
+  }
+
+  return Platform.select({
```
tbergq: Use `StyleSheet.create` if it works with header style

PR #443, opened 2018-03-29
Remove inline styles

tbergq:
I left tests out of it, and `Translations` there were some tests failing when I used StyleSheet in translations, saying create is undefined. Probably circular references?


mrtnzlml:
> Probably circular references?

Yes.

PR #444, opened 2018-04-03
Remove componentWillReceiveProps

tbergq:


app/translations/src/TranslationFragment.js
```diff
@@ -25,12 +25,14 @@ export default class TranslationFragment extends React.Component<{|
...
-
-    return React.Children.map(this.props.children, child =>
-      React.cloneElement(child, {
-        ...this.props,
-        children: null,
-      }),
-    );
+    return React.Children.map(this.props.children, child => {
+      if (child) {
+        return React.cloneElement(child, {
```
tbergq: This change is necessary because condintional useage like 
```
<TranslationFragment>
        {stars.includes(0) && (
          <TranslationFragment>
            <Translation
              id="HotelsSearch.Filter.StarsFilter.Unrated"
              key="unratedTranslation"
            />
            {stars.length > 1 && <DummyTranslation id="," />}
          </TranslationFragment>
        )}
        <DummyTranslation
          id={stars
            .sort()
            .filter(star => star > 0)
            .join(',')}
        />
      </TranslationFragment>
```
gives false as a child, and React.cloneElement(false) crashes.

app/translations/src/TranslationFragment.js
```diff
@@ -25,12 +25,14 @@ export default class TranslationFragment extends React.Component<{|
...
-      }),
-    );
+    return React.Children.map(this.props.children, child => {
+      if (React.isValidElement(child)) {
+        return React.cloneElement(child, {
+          ...this.props,
+          children: null,
+        });
+      }
+      return null;
```
mrtnzlml: Sorry, I didn't realize one thing. Maybe we should return the children here? What about plain string for example? It's not a valid element but it's valid children. This would remove it. Try `<TranslationFragment>just a string</TranslationFragment>`. If I am correct please add a test for it.

app/hotels/src/map/allHotels/MapView.js
```diff
@@ -64,7 +64,8 @@ export class Map extends React.Component<Props, State> {
     this.markers[selectedIndex] && this.markers[selectedIndex].showCallout();
   };
 
-  componentWillReceiveProps = (nextProps: Props) => {
+  UNSAFE_componentWillReceiveProps = (nextProps: Props) => {
+    // TODO: refactor, not quite sure how to do this in static method
```
mrtnzlml: I think you have to do it the same way like in the `MapScreen`. See: https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#updating-state-based-on-props

mrtnzlml: Separate PR, please.

PR #445, opened 2018-04-03
Show no results found message on map

tbergq:


app/hotels/src/map/allHotels/HotelSwipeList.js
```diff
@@ -77,23 +82,34 @@ class HotelSwipeList extends React.Component<Props, State> {
...
-            decelerationRate="fast"
-            activeSlideAlignment="start"
-            containerCustomStyle={styles.slider}
-            removeClippedSubviews={false}
-            onSnapToItem={onSnapToItem}
-          />
-        </View>
-        <Address address={this.getSelectedAddress()} />
+        {data.length ? (
+          [
```
mrtnzlml: You can use React Fragments now. We upgraded our React... 🎉

PR #446, opened 2018-04-03
Fix snap to item in map view

tbergq:


app/hotels/src/map/allHotels/HotelSwipeList.js
```diff
@@ -71,26 +77,31 @@ class HotelSwipeList extends React.Component<Props, State> {
...
             itemWidth={this.getCardItemWidth()}
-            firstItem={selectedIndex}
             inactiveSlideScale={1}
             inactiveSlideOpacity={0.5}
             decelerationRate="fast"
             activeSlideAlignment="start"
             containerCustomStyle={styles.slider}
             removeClippedSubviews={false}
             onSnapToItem={onSnapToItem}
+            useScrollView
```
tbergq: We need to use scroll view with this approach. Default is list view, and list view keeps limited amount in memory. If we snap to an item not in memory, we will se a blank slide only.

mrtnzlml:
Integrated, thanks.

PR #451, opened 2018-04-04
Fix datepicker button styling

tbergq:
Datepicker confirm and cancel button had disappeared due to too much paddingVertical


mrtnzlml:
@tbergq Haha, thanks. I was not able to fix it yesterday so I postponed it until the next day. This is how I like to solve the issues... 😄

PR #465, opened 2018-04-06
Dependencies: upgrade React Navigation to the latest version (^1.5.11)

mrtnzlml:
@tbergq Please test it. I didn't notice a single problem.


tbergq:
Looks good, can't find any issues either

PR #469, opened 2018-04-09
Add header to all hotels map

### LINKED ISSUE #468: Navigation is not visible on all hotels map
screen shot 2018-04-07 at 14 42 20

tbergq:
closes #468


app/hotels/src/navigation/AllHotelsMapNavigationScreen.js
```diff
@@ -16,6 +17,14 @@ type Props = {|
...
   Props,
 > {
+  static navigationOptions = ({ navigation }: Props) => {
+    function goBack() {
+      navigation.goBack(null);
+    }
+    return {
+      headerLeft: <HeaderBackButton tintColor="#fff" onPress={goBack} />,
+    };
+  };
```
mrtnzlml: @tbergq I don't understand our navigation system. Do we have to call this on every screen?

tbergq: Not on every screen, but on the first level of each nested stack yes.

mrtnzlml: That's silly. So why is there `StackNavigator` only for one map screen?

PR #472, opened 2018-04-09
Show recent searches

### LINKED ISSUE #471: Show recent searches in location picker
[Trello](https://trello.com/c/wEzJeazf/239-search-overlay-should-be-displayed-with-recent-search)
[Zeplin](https://zpl.io/bAGyo5n)

tbergq:
closes #471

### DIFF
--- a/app/hotels/src/allHotels/searchForm/locationPicker/RecentSearchItem.js
+++ b/app/hotels/src/allHotels/searchForm/locationPicker/RecentSearchItem.js
@@ -1,0 +1,47 @@
+// @flow
+
+import * as React from 'react';
+import {
+  Touchable,
+  StyleSheet,
+  Text,
+  Color,
+} from '@kiwicom/react-native-app-shared';
+import Translations from '@kiwicom/react-native-app-translations';
+
+import Row from './Row';
+
+const styles = StyleSheet.create({
+  text: {
+    fontSize: 14,
+    fontWeight: '600',
+    color: Color.textDark,
+  },
+});
+
+type Location = {|
+  id: string,
+  name: string,
+|};
+
+type Props = {|
+  location: Location,
+  onCitySelected: (cityId: string, cityName: string) => void,
+|};
+
+export default class RecentSearchItem extends React.Component<Props> {
+  onPress = () => {
+    const { id, name } = this.props.location;
+    this.props.onCitySelected(id, name);
+  };
+
+  render = () => (
+    <Touchable onPress={this.onPress}>
+      <Row>
+        <Text style={styles.text}>
+          <Translations passThrough={this.props.location.name} />
+        </Text>
+      </Row>
+    </Touchable>
+  );
+}
--- a/app/hotels/src/allHotels/searchForm/locationPicker/__tests__/__snapshots__/RecentSearches.test.js.snap
+++ b/app/hotels/src/allHotels/searchForm/locationPicker/__tests__/__snapshots__/RecentSearches.test.js.snap
@@ -1,0 +1,46 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`RecentSearches renders 1`] = `
+<UNDEFINED>
+  <Component
+    style={
+      Object {
+        "padding": 10,
+      }
+    }
+  >
+    <Translation
+      id="hotels_search.location_picker.recent_search"
+    />
+  </Component>
+  <FlatList
+    data={
+      Array [
+        Object {
+          "id": "1",
+          "name": "Rome",
+        },
+        Object {
+          "id": "2",
+          "name": "Oslo",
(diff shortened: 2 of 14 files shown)


app/hotels/src/allHotels/searchForm/locationPicker/__tests__/__snapshots__/RecentSearches.test.js.snap
```diff
@@ -0,0 +1,46 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`RecentSearches renders 1`] = `
+<UNDEFINED>
```
tbergq: Looks like `react-test-renderer` does not know about `React.Fragment` yet

mrtnzlml: I am fine to accept this. Thanks for pointing it out.

ios/reactNativeApp.xcodeproj/project.pbxproj
```diff
@@ -1176,7 +1177,17 @@
 			);
 			name = "[CP] Copy Pods Resources";
 			outputPaths = (
-				"${TARGET_BUILD_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}",
+				"${TARGET_BUILD_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}/Entypo.ttf",
```
tbergq: Looks like we are finally synced right?

mrtnzlml: Hopefully, yes.

app/hotels/src/allHotels/searchForm/locationPicker/__tests__/__snapshots__/RecentSearches.test.js.snap
mrtnzlml: This will be fixed in the future version of Jest: https://github.com/facebook/jest/issues/5786

app/hotels/src/allHotels/searchForm/locationPicker/RecentSearchItem.js
```diff
@@ -0,0 +1,28 @@
+// @flow
+
+import * as React from 'react';
+import { Touchable, Text } from '@kiwicom/react-native-app-shared';
+import Translations from '@kiwicom/react-native-app-translations';
+
+import type { Location } from './Types';
+import styles from './Styles';
```
mrtnzlml: @tbergq Could you please make it more reusable? I would prefer to have the style encapsulated into component instead of requiring and using this style directly. One reason for this is a question from yesterday: "Could we reuse the suggestions picker for something else? For example flight locations?".

So instead of `Touchable` with required style let's introduce `Row` component related to our picker. What do you think?

app/hotels/src/allHotels/searchForm/locationPicker/Types.js
```diff
@@ -0,0 +1,6 @@
+// @flow
```
mrtnzlml: I really don't like doing this. It's confusing. Every component defines its own properties and I don't see a reason why should they be shared (they look the same but there is no relation between these components, right?). It's fine to repeat yourself here. But maybe you have a different opinion on this...

tbergq: > they look the same but there is no relation between these components
That is not how I see it. 

`LocationPickerScreen` loads locations and passes them to `RecentSearches` who passes a location to `RecentSearchItem`,  this are the same locations. If this object changes in the future, say we add a `searchedDate` we would have to change the type in 3 places.

Maybe a separate file is not the way to do it, maybe export it directly from `LocationPickerScreen`?

app/hotels/src/allHotels/searchForm/locationPicker/RecentSearchItem.js
tbergq: yes makes sense

app/hotels/src/allHotels/searchForm/locationPicker/Types.js
mrtnzlml: I understand that but my point is this: component `RecentSearches` defines its properties and `RecentSearchItem` defines its properties. From the component perspective, there is no relation between these two components because they are encapsulated and if something change in the parent component it cannot affect the inner component - you don't have to change everything, just change one level and keep to contract with the second level. We did this mistake with `AvailableHotelSearchInput` type and it's the biggest mess ever.

Anyway, I think it's better to export the type from the other component.

tbergq:
@mrtnzlml please check again

mrtnzlml:
Integrated, thanks.

PR #474, opened 2018-04-10
Add version to payment screen label

### LINKED ISSUE #464: Add build version into ancillaries logging
Currently, we use something like `kiwi-ios-react` label. The new format should contain even the version so for example `kiwi-ios-react-0.0.18`. The easiest solution is probably to require `package.json` directly... (Because hey! We are in JS! 🙂)

Alternative inspiration: https://github.com/facebook/react-native/blob/v0.55.2/Libraries/Core/ReactNativeVersion.js

tbergq:
closes #464

### DIFF
--- a/app/hotels/src/singleHotel/PaymentScreen.js
+++ b/app/hotels/src/singleHotel/PaymentScreen.js
@@ -7,6 +7,7 @@ import { DateFormatter } from '@kiwicom/react-native-app-translations';
 import querystring from 'querystring';
 
 import { sanitizeDate } from '../GraphQLSanitizers';
+import hotelPackage from '../../package.json';
 
 export type PaymentParameters = {|
   hotelId: number,
@@ -60,7 +61,7 @@ export function createURI(pp: PaymentParameters): string {
       ...roomsQuery,
       rt_pos_selected: '', // ???
       aid: pp.affiliateId,
-      label: `kiwi-${Platform.OS}-react`,
+      label: `kiwi-${Platform.OS}-react-${hotelPackage.version}`,
       lang: pp.language,
       selected_currency: pp.currency,
     })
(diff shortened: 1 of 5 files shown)


scripts/releaseHotels.js
```diff
@@ -2,13 +2,26 @@
...
 const child_process = require('child_process');
+const fs = require('fs');
 
 const exec = (command, options) =>
   child_process.execSync(command, {
     stdio: 'inherit',
     ...options,
   });
 
+const getFileContentAsObject = (filePath, options) => {
```
mrtnzlml: I will show you a magic here:

```
▲ react-native-app at master ✔ node                                    
> const packageInfo = require('./app/hotels/package.json')
undefined
> console.log(packageInfo.version)
0.0.0
undefined
> 
```

This is what I meant with:

> Because hey! We are in JS! 🙂

app/hotels/src/GetVersion.js
```diff
@@ -0,0 +1,5 @@
+// @flow
+
+import hotelPackage from '../package.json';
+
+export const getVersion = () => hotelPackage.version;
```
mrtnzlml: Hm, but you did it here... 🤔

I need to read it again what are you doing there actually... 😄

tbergq:
@mrtnzlml  ready to be reviewed

app/hotels/src/singleHotel/PaymentScreen.js
```diff
@@ -7,6 +7,7 @@ import { DateFormatter } from '@kiwicom/react-native-app-translations';
 import querystring from 'querystring';
 
 import { sanitizeDate } from '../GraphQLSanitizers';
+import hotelPackage from '../../package.json';
```
mrtnzlml: @tbergq Is this going to work after we'll bundle this package?

PR #476, opened 2018-04-11
Dependencies: upgrade RN to the latest version (^0.55.2)

mrtnzlml:


mrtnzlml:
@tbergq Fixed. I think we should start doing the podspec file...

tbergq:
>  I think we should start doing the podspec

And what about android?

mrtnzlml:
We should swallow their Maven builder script because they cannot handle the versioning so we never know what's happening and what version are they using (they don' know it either, unfortunately 😞). Dunno what's the podspec equivalent there...

PR #483, opened 2018-04-12
Add new datepicker component

tbergq:


app/shared/src/forms/datePicker/DatePicker.ios.js
```diff
@@ -0,0 +1,81 @@
...
+          onSave={this.onSave}
+          onClose={this.togglePopup}
+        >
+          <DatePickerIOS
+            {...rest}
+            minimumDate={minDate}
+            maximumDate={maxDate}
+            date={this.state.date}
+            onDateChange={this.onDateChange}
+            locale={NativeModules.RNDeviceInfo.Locale}
```
tbergq: Should we rather expose the `GetDeviceLocale` method?

app/shared/src/forms/datePicker/DatePicker.android.js
```diff
@@ -0,0 +1,40 @@
...
+  date: Date,
+  format?: string,
+  onDateChange: (date: Date) => void,
+  minDate?: Date,
+  maxDate?: Date,
+};
+
+export default class DatePicker extends React.Component<Props> {
+  openDatePicker = () => {
+    // Touchable needs a method not returning a promise
```
mrtnzlml: I do not understand this comment. Maybe please try to clarify it differently.

tbergq: I'll sign up for the english course as well 😞 I will improve the comment

app/shared/src/forms/datePicker/DatePicker.android.js
```diff
@@ -0,0 +1,40 @@
...
+import { DatePickerAndroid } from 'react-native';
+
+import DatePickerButton from './DatePickerButton';
+
+type Props = {
+  date: Date,
+  format?: string,
+  onDateChange: (date: Date) => void,
+  minDate?: Date,
+  maxDate?: Date,
```
mrtnzlml: I prefer to keep required params first so it's clear that you need to use `date` and `onDateChange`. What do you think?

app/shared/src/forms/datePicker/DatePicker.android.js
mrtnzlml: @tbergq No, it's fine. I am just not sure about the reason and why you actually pointed this out... :) Just make it more clear for your future @tbergq so he will understand immediately what is happening here... :)

app/shared/src/forms/datePicker/DatePicker.ios.js
```diff
@@ -0,0 +1,81 @@
...
+};
+
+type State = {|
+  isPickerOpen: boolean,
+  date: Date,
+|};
+
+export default class DatePicker extends React.Component<Props, State> {
+  static defaultProps = {
+    format: 'MMM DD',
```
mrtnzlml: This whole component seems like it behaves completely differently from Android (for example this format). I would expect to have same interfaces and behavior just platform specific...

tbergq: I started making the ios picker, this I moved this defaultProps down to `DatePickerButton` and forgot to remove it here. 

The components needs to behave differently since the underlying code is different, `DatePickerIOS` vs `DatePickerAndroid`

They take the same props so they are used the same from the outside

mrtnzlml: > They take the same props so they are used the same from the outside

This is what I mean by "same interfaces" and by the "same behavior" I mean "do not format the dates here and not in the Android component"... :)

tbergq:
@mrtnzlml please check again

app/shared/src/forms/datePicker/__tests__/DatePicker.test.js
```diff
@@ -0,0 +1,64 @@
...
+
+describe('DatePicker', () => {
+  it('renders in playground', () => {
+    PlaygroundRenderer.render(<PickerWrapper />);
+  });
+
+  it('works with min date', () => {
+    PlaygroundRenderer.render(
+      <View>
+        <Translation passThrough="With min date" />
```
mrtnzlml: Please don't do this. Always think about this file as about the test file. This is absolutely not relevant to the test. We should use the description from `describe` or `it` if you would like to see the description in the Playground.

mrtnzlml:
Could you please also fix the vertical alignment?

screen shot 2018-04-13 at 10 31 26

screen shot 2018-04-13 at 10 31 26 2

app/translations/index.js
```diff
@@ -6,8 +6,14 @@ import Translation from './src/Translation';
...
 import DateFormatter from './src/DateFormatter';
 import MapLocaleToLanguageQuery from './src/MapLocaleToLanguageQuery';
+import GetDeviceLocale from './src/GetDeviceLocale';
 
-export { TranslationFragment, DateFormatter, MapLocaleToLanguageQuery };
+export {
+  TranslationFragment,
+  DateFormatter,
+  MapLocaleToLanguageQuery,
+  GetDeviceLocale,
```
mrtnzlml: Should we rename this package to the "localization"? (not in this PR) Name "translations" seems to be too specific.

mrtnzlml: OK: https://github.com/kiwicom/react-native-app/issues/484

mrtnzlml:
Integrated, thanks.

PR #485, opened 2018-04-13
WIP: Replace sync native methods with async methods

tbergq:


android/gradle.properties
```diff
@@ -19,4 +19,6 @@
 
 android.useDeprecatedNdk=true
 MYAPP_RELEASE_STORE_FILE=kiwi-rn-hotels.keystore
-MYAPP_RELEASE_KEY_ALIAS=kiwi-rn-hotels
\ No newline at end of file
+MYAPP_RELEASE_KEY_ALIAS=kiwi-rn-hotels
+android.enableAapt2=false
```
tbergq: This was added because of this issue
https://github.com/facebook/react-native/issues/16906

android/gradle.properties
```diff
@@ -19,4 +19,6 @@
...
+org.gradle.jvmargs=-Xmx3072m
```
tbergq: See 
https://stackoverflow.com/questions/37090135/to-run-dex-in-process-the-gradle-daemon-needs-a-larger-heap-it-currently-has-a

app/hotels/src/filter/price/PricePopup.js
```diff
@@ -54,14 +49,12 @@ export default class PricePopup extends React.Component<Props, State> {
   renderLabel = () => {
     const { start, end } = this.state.price;
     const { currency } = this.props;
-    const endPrice = <Price amount={end} currency={currency} />;
-    const startPrice = <Price amount={start} currency={currency} />;
```
tbergq: This is currently only working with EUR anyway

tbergq:
@mrtnzlml Have a look at it and decide wether we should proceed with this PR or not

android/gradle.properties
mathieumg: The better fix is to use the Gradle file from the PR in the issue you linked, and apply that one for the time being. It was mentioned that not using AAPT2 could cause some older devices to crash!

tbergq: @mathieumg, thanks for pointing this out. I think in our case, it it still ok. We are currently just providing a bundle to be consumed by our native app. 

I build this apk to test performance with production build.

tbergq:
@mrtnzlml did you have time to look at this?

app/shared/src/errors/__tests__/__snapshots__/PartialFailure.test.js.snap
```diff
@@ -61,7 +61,7 @@ Array [
           ]
         }
       >
-        Some parts of the page may be missing due to partial server error.
```
mrtnzlml: @tbergq I don't understand this change. Can you please explain it? (fix it)

tbergq: `Translation` renders first with `this.state.translatedString = ''` then async calls native, gets the response and calls `this.setState({translatedString})`

So on first render, translation just renders empty string, that is why ^^ has changed

mrtnzlml: In this case, I am afraid I cannot accept this PR since we are not able to test simple pass-through type of translation... 😞 

This is for me obviously failing test.

tbergq: > This is for me obviously failing test

This is really another component, mocking it should solve this

mrtnzlml: I see it's not pass through translation in this case... :-/

PR #494, opened 2018-04-18
Upgrade react native version

tbergq:
Fixes React/Fabric issue

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   },
   "dependencies": {
     "react": "^16.3.1",
-    "react-native": "^0.55.2"
+    "react-native": "^0.55.3"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


mrtnzlml:
What issue? Could you link it here as well? Seems like this fix didn't affect our code at all?

package.json
```diff
@@ -22,7 +22,7 @@
   },
   "dependencies": {
     "react": "^16.3.1",
-    "react-native": "^0.55.2"
+    "react-native": "^0.55.3"
```
mrtnzlml: @tbergq Do we have to upgrade our `package.json` from build folder as well? Maybe no, because there is `^...` anyway? :)

tbergq:
> What issue

https://github.com/facebook/react-native/issues/18683

PR #499, opened 2018-04-20
Fix 'babel-eslint' dependency

mrtnzlml:
We are using caret operator everywhere else and this is a mistake.


package.json
```diff
@@ -26,7 +26,7 @@
   },
   "devDependencies": {
     "babel-cli": "^6.26.0",
-    "babel-eslint": "~8.2.2",
+    "babel-eslint": "^8.2.2",
     "babel-jest": "22.4.3",
     "babel-plugin-idx": "^2.2.0",
```
tbergq: what about this?

package.json
```diff
@@ -26,7 +26,7 @@
   },
   "devDependencies": {
     "babel-cli": "^6.26.0",
-    "babel-eslint": "~8.2.2",
+    "babel-eslint": "^8.2.2",
     "babel-jest": "22.4.3",
```
tbergq: Should this also be `^`?

package.json
```diff
@@ -26,7 +26,7 @@
...
     "babel-plugin-idx": "^2.2.0",
     "babel-preset-react-native": "4.0.0",
```
tbergq: and this?

mrtnzlml:
@tbergq WTF happened with our dependencies? 😄 Fixed hopefully everywhere...

tbergq:
> WTF happened with our dependencies

👻 commiter 🤔

PR #500, opened 2018-04-23
Inject playground tests and start playground

tbergq:


app/playground/src/PlaygroundList.js
```diff
@@ -5,27 +5,29 @@ import { FlatList, View } from 'react-native';
...
-import '../../shared/src/__tests__/Text.test.js';
-import '../../shared/src/badge/__tests__/AdaptableBadge.test.js';
-import '../../shared/src/buttons/__tests__/Button.test.js';
-import '../../shared/src/buttons/__tests__/IncrementDecrementButtons.test.js';
-import '../../shared/src/cards/__tests__/SimpleCard.test.js';
-import '../../shared/src/forms/__tests__/Slider.test.js';
-import '../../shared/src/popup/__tests__/ButtonPopupPlayground.test.js';
 import PlaygroundListItem from './PlaygroundListItem';
 import PlaygroundRenderer from './PlaygroundRenderer';
+// PlaygroundImportsStart
```
mrtnzlml: @tbergq What about creating one file only with these autogenerated imports and include it here?  I don't like this semi-automatic tweaking of one part of our code...

tbergq: Yeah, that sounds like a better idea. 👍

tbergq:
@mrtnzlml check again please

PR #502, opened 2018-04-24
Add checkin and checkout date props to all hotels package

tbergq:


app/hotels/src/allHotels/AllHotels.js
```diff
@@ -40,21 +44,39 @@ type Props = {|
...
+        checkout: DateFormatter(checkout).toDate(),
+      });
+    } else {
+      this.validateCheckinDate();
+    }
+  };
+
+  componentWillUnmount = () => {
+    this.props.onLocationChange('');
+    this.props.onCityIdChange(null);
```
tbergq: I noticed that if you chose a location from the location picker, clicked back to homepage, then clicked to to hotels search, it still searched with the last picked location, which is not what you would expect. So therefore I added this calls on unmount

tbergq:
Close this and make new PR without redux dependency

PR #504, opened 2018-04-24
Show cityName from allHotelsQuery when available

### LINKED ISSUE #501: Add more props to KiwiHotels package
- checkin and checkout date
- destination

tbergq:
closes #501

### DIFF
--- a/app/hotels/src/allHotels/AllHotels.js
+++ b/app/hotels/src/allHotels/AllHotels.js
@@ -119,7 +119,7 @@ export class AllHotels extends React.Component<PropsWithContext> {
             onChange={this.props.setFilter}
             currency={this.props.currency}
           />
-          {this.props.cityId ? (
+          {this.props.cityId || this.props.coordinates ? (
             <AllHotelsSearch
               coordinates={this.props.coordinates}
               openSingleHotel={this.props.openSingleHotel}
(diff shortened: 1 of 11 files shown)


tbergq:
Setting cityName to location in context triggers a new search using cityId since current logic is 
```
if (hasCoordinates(coordinates) && location === '') {
    params.latitude = idx(coordinates, _ => _.latitude);
    params.longitude = idx(coordinates, _ => _.longitude);
  } else if (cityId) {
    params.cityId = cityId;
  }
```

cityId is always available. This logic is becoming painful to deal with. I suggest we add this to `Biggest pain` issue and somehow improve it.

mrtnzlml:
> I suggest we add this to Biggest pain issue and somehow improve it.

Do so please and try to provide info how to improve it and what's the pain actually.

tbergq:
Rebased and pushed. I will describe the pain in the issue.

tbergq:
I found a way to improve this a bit, still leaves us with the ugly ternary in `AllHotels`
```
{this.props.cityId || this.props.coordinates ? (
            <AllHotelsSearch
              coordinates={this.props.coordinates}
              openSingleHotel={this.props.openSingleHotel}
              cityId={this.props.cityId}
              currency={this.props.currency}
            />
          ) : (
            <PublicApiRenderer
...
```
Which we should improve

PR #506, opened 2018-04-24
Replace Redux with React Context components

mrtnzlml:
Please read this: https://reactjs.org/docs/context.html


app/hotels/src/allHotels/AllHotelsSearch.js
```diff
@@ -14,21 +13,23 @@ import type { Coordinates } from '../CoordinatesType';
...
-type Props = {|
-  search: SearchParams,
+type PropsWithContext = {
+  // external:
   openSingleHotel: (searchParams: AvailableHotelSearchInput) => void,
-  onCityIdChange: (cityId: string | null) => void,
   cityId: string,
   coordinates: Coordinates | null,
+  currency: string,
+  // from context
```
tbergq: Why are you not using `...Props` here as done in the other components?

app/hotels/src/HotelsFilterContext.js
```diff
@@ -0,0 +1,89 @@
...
+  actions: {|
+    setFilter: OnChangeFilterParams => void,
+  |},
+|};
+
+class Provider extends React.Component<Props, State> {
+  setFilter = (action: OnChangeFilterParams) => {
+    const newState = {
+      ...this.state,
+      filterParams: {
```
tbergq: Shouldn't you be using functional set state here?

app/hotels/src/allHotels/AllHotelsSearch.js
mrtnzlml: I tried but I had to use non-exact props and for some reason, it doesn't work. Can you please check it what's wrong? I wasn't able to fix it this way... 🙈

tbergq: I will have a look and see if I can find out

tbergq: Ok, I can't really find out why this doesn't work. We could merge it and try to make it exact type later in another PR.

mrtnzlml: Should be possible really soon: https://github.com/facebook/flow/issues/2405#issuecomment-379175609

app/hotels/src/HotelsFilterContext.js
mrtnzlml: Yes, fixed - please check it again.

PR #509, opened 2018-04-25
Add checkin and checkout props to hotels package

tbergq:


app/hotels/src/HotelsSearchContext.js
```diff
@@ -92,7 +92,7 @@ class Provider extends React.Component<Props, State> {
     });
   };
 
-  setLocationAndCityId = (cityId: string, location: string) => {
+  setCityIdAndLocation = (cityId: string | null, location: string) => {
```
tbergq: Renamed because first parameter is city id, so naming was confusing

app/hotels/src/allHotels/AllHotels.js
```diff
@@ -40,15 +43,30 @@ type PropsWithContext = {
...
+        checkout: checkout,
+      });
+    } else {
+      this.validateCheckinDate();
+    }
+  };
+
+  componentWillUnmount = () => {
+    this.props.setCityIdAndLocation(null, '');
   };
```
tbergq: I noticed that if you chose a location from the location picker, clicked back to homepage, then clicked back to hotels search, it still searched with the last picked location, which is not what you would expect. So therefore I added this calls on unmount

tbergq:
Seems our datepicker did not handle date changes when they came from outside it self. This has never been possible before. But it is fixed now. 

Please check again

PR #512, opened 2018-04-26
Make tests that render translations async

### LINKED ISSUE #510: Components are leaking memory or doing no-ops
You can reproduce this issue by running `yarn test`. You'll see two warning messages:

```
 PASS  app/hotels/src/navigation/allHotels/__tests__/GuestsModalScreen.test.js
  ● Console

    console.error node_modules/fbjs/lib/warning.js:1
      Warning: Can't call setState (or forceUpdate) on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
          in Translation (created by GuestsModal)
          in Text (created by Text)
          in Text (created by Text)
          in Text (created by GuestsModal)
          in View (created by Component)
          in Component (created by GuestsModal)
          in View (created by Component)
          in Component (created by GuestsModal)
```

And:

```
 PASS  app/localization/src/__tests__/Translation.test.js
  ● Console

    console.error node_modules/fbjs/lib/warning.js:1
      Warning: Can't call setState on a component that is not yet mounted. This is a no-op, but it might indicate a bug in your application. Instead, assign to `this.state` directly or define a `state = {};` class property with the desired state in the Translation component.
```

Let's fix it, please.

tbergq:
closes #510

### DIFF
--- a/app/hotels/src/navigation/allHotels/__tests__/GuestsModalScreen.test.js
+++ b/app/hotels/src/navigation/allHotels/__tests__/GuestsModalScreen.test.js
@@ -82,23 +82,23 @@ describe('GuestsModalScreen', () => {
     expect(navigation.goBack).not.toHaveBeenCalled();
   });
 
-  it('should remove error when children is set to 0', () => {
+  it('should remove error when children is set to 0', async () => {
     const onClose = jest.fn();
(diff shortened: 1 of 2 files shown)


mrtnzlml:
Thank you.

PR #513, opened 2018-04-26
Refactor all hotels

tbergq:
Use cityId or coordinates if passed, or use device location.
If user does not give permission, show message to ask user to search for city


mrtnzlml:
@tbergq Can we use a similar technique to implement "Your location" feature in the locations picker?

tbergq:
Absolutely, that's why i made the Context in shared. This can be reused anywhere

app/hotels/src/allHotels/AllHotels.js
```diff
@@ -172,16 +169,23 @@ export default function AllHotelsWithContext(props: Props) {
...
-              {...props}
-              location={location}
-              cityId={cityId}
-              search={searchParams}
-              filter={filterParams}
-              setSearch={setSearch}
-              setFilter={setFilter}
-              setCityIdAndLocation={setCityIdAndLocation}
-            />
+            <GeolocationContext.Consumer>
```
mrtnzlml: I think this (3 nested context, many props) is a great indicator that we are maybe doing too much in this component. Next refactoring should follow... :)

tbergq: Yes, I agree, it has become a little monster 🙈

app/shared/src/GeolocationContext.js
```diff
@@ -0,0 +1,66 @@
...
+    actions: {
+      updateGeolocation: this.updateGeolocation,
+    },
+  };
+  componentDidMount = () => {
+    this.updateGeolocation();
+  };
+
+  updateGeolocation = () => {
+    navigator.geolocation.getCurrentPosition(
```
mrtnzlml: @tbergq Where is this `navigator` from?

tbergq: https://facebook.github.io/react-native/docs/geolocation.html

tbergq: > this API is available through the navigator.geolocation global - you do not need to import it.

app/hotels/src/allHotels/AllHotels.js
```diff
@@ -77,78 +83,69 @@ export class AllHotels extends React.Component<PropsWithContext> {
...
+              <AllHotelsSearch
+                coordinates={coordinates}
+                openSingleHotel={this.props.openSingleHotel}
+                cityId={this.props.cityId}
+                currency={this.props.currency}
+              />
+            ) : (
+              <View style={styles.searchForCityContainer}>
+                {!this.props.canGetUserLocation ? (
+                  <Translation id="hotels_search.all_hotels.please_search" />
```
mrtnzlml: So this will allow us to avoid the API renderer in order to fetch initial locations, right? I just want to confirm the idea.

tbergq: yes

app/shared/src/GeolocationContext.js
```diff
@@ -0,0 +1,66 @@
...
+  actions: {|
+    updateGeolocation: () => void,
+  |},
+|};
+
+class Provider extends React.Component<Props, State> {
+  state = {
+    ...defaultState,
+    actions: {
+      updateGeolocation: this.updateGeolocation,
```
mrtnzlml: This function is not used outside the context so why is it exposed? It's also `undefined` at this moment...

app/shared/src/GeolocationContext.js
mrtnzlml: Should we set `geo_options` (timeout or maximumAge - dunno what's the difference)? When I change the location it will never return the new one and I think it's because of `maximumAge = INFINITY`...

app/shared/src/GeolocationContext.js
tbergq: the idea was to offer a way for components to refresh lat, lng if needed. Why it is undefined, I don't know 🙈

mrtnzlml: > the idea was to offer a way for components to refresh lat, lng if needed

app/shared/src/GeolocationContext.js
tbergq: timeout is the the timeout for getting gps locations I guess. maxiumumAge sounds like, how long it will allow you to use lat, lng. Infinity sounds ok to me. I can set timeout.

mrtnzlml: So you'll open the app in Prague and you'll have Prague there forever? (try to open the app, change the search location, go back to the homepage and go back to the search...)

tbergq: If you close the app, then open the app, you would get new gps coordinates.
You are always free to search for a location.

Since we have the app state change, we could always refresh gps coordinates there. Sound ok?

PR #514, opened 2018-04-27
Add newline-after-flow-annotation rule

tbergq:


.eslintrc.js
```diff
@@ -94,5 +94,9 @@ module.exports = {
     'react-native/no-unused-styles': ERROR,
     'react-native/no-inline-styles': WARNING,
     'flowtype/require-valid-file-annotation': [ERROR, 'always'],
+    'flowtype/newline-after-flow-annotation': [
+      2,
```
mrtnzlml: `ERROR` to be more obvious, please... :)

mrtnzlml:
@tbergq Now the question is - is this rule (one line) really stupid and we should remove it as @oreqizer suggested?

.eslintrc.js
tbergq: Sorry, copy/paste `rookie-mistake`

tbergq:
Really, its your OCD. For me it is the same

PR #517, opened 2018-05-02
Add setup for MMB package

tbergq:


app/MMB/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "@kiwicom/mobile-manage-by-booking",
```
mrtnzlml:
```diff
-   "name": "@kiwicom/mobile-manage-by-booking",
+   "name": "@kiwicom/mobile-manage-my-booking",
```

app/core/package.json
```diff
@@ -4,6 +4,7 @@
   "private": true,
   "dependencies": {
     "@kiwicom/react-native-app-hotels": "^0",
+    "@kiwicom/mobile-manage-by-booking": "^0",
```
mrtnzlml:
```diff
-     "@kiwicom/mobile-manage-by-booking": "^0",
+     "@kiwicom/mobile-manage-my-booking": "^0",
```

app/core/src/navigation/index.js
```diff
@@ -30,16 +29,32 @@ const VoidStack = StackNavigator(
...
       screen: HomepageStack,
     },
     HotelsPackage: {
-      screen: withMappedProps(HotelsPackageWrapper),
+      screen: HotelsStack,
     },
-    SingleHotelPackage: {
-      screen: withMappedProps(SingleHotelsPackageWrapper),
+    MMBPackage: {
+      screen: tabNavigator,
```
mrtnzlml: I don't understand this change. My initial intention was to go closer to our native app by creating the tab navigator (as we have in our app) so we can say later - _"look, this is what we have and it's basically the same with native"_. So I wanted to use even the homepage as a homepage (even though we have only links there so far). What do you think about that?

app/MMB/src/navigation/MMBScreen.js
```diff
@@ -0,0 +1,42 @@
+// @flow
+
+import * as React from 'react';
+import { View } from 'react-native';
+import { Translation } from '@kiwicom/mobile-localization';
+import type { NavigationType } from '@kiwicom/mobile-navigation';
+import { HeaderBackButton } from 'react-navigation';
+
+const isDevelopment = process.env.NODE_ENV;
```
mrtnzlml: Just a silly question - `NODE_ENV` is defined only during development and therefore it's safe to use it like this?

app/MMB/src/navigation/NavigationStack.js
```diff
@@ -0,0 +1,28 @@
...
+
+import {
+  StackNavigator,
+  StackNavigatorOptions,
+  type NavigationType,
+} from '@kiwicom/mobile-navigation';
+
+import MMBScreen from './MMBScreen';
+
+export type NavigationProps = {|
```
mrtnzlml: This type is not used.

app/MMB/package.json
tbergq: 🙈

app/core/src/navigation/index.js
tbergq: Ait, I guess I can work with that.

mrtnzlml: The change is simple - just keep the homepages as we do (we could actually remove some links) and let's work on the second tab (the same with our native app).

Later on, we will have to introduce the horizontal scroll view with suggested hotels and we can directly use it on our homepage...

app/MMB/src/navigation/MMBScreen.js
tbergq: It should be `development` in dev-mode and `production` when deployed. 

This is just a silly developer-mistake 🙈

app/MMB/src/navigation/NavigationStack.js
tbergq: copy/paste 🙈

tbergq:
@mrtnzlml please check again

PR #533, opened 2018-05-04
MMB: create MenuItem and MenuGroup components

mrtnzlml:
Just a design at this moment (it's not clickable).

### DIFF
--- a/app/MMB/src/components/MenuGroup.js
+++ b/app/MMB/src/components/MenuGroup.js
@@ -1,0 +1,68 @@
+// @flow
+
+import * as React from 'react';
+import { View } from 'react-native';
+import { Color, StyleSheet, Text } from '@kiwicom/mobile-shared';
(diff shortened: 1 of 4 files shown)


app/MMB/src/components/MenuGroup.js
```diff
@@ -0,0 +1,68 @@
...
+ * |  ICON                        |
+ * |                              |
+ * |        ----------------------|
+ * |                              |
+ * |  ICON                        |
+ * |                              |
+ * `------------------------------`
+ */
+export default function MenuGroup(props: Props) {
+  let separator = null;
```
tbergq: Use const when not reassigning

Which reminds me, we should add
https://eslint.org/docs/rules/prefer-const

tbergq:
Should these menu components rather be in the shared folder? Say we should implement profile tab next, it uses the same menu items.

mrtnzlml:
Yes, but I will move it when needed.

PR #534, opened 2018-05-04
Add city image component

tbergq:


app/MMB/src/mmbList/ImageBadges.js
```diff
@@ -0,0 +1,75 @@
...
+        style={styles.bookingIdBadge}
+        textStyle={styles.badgeText}
+      />
+    </React.Fragment>
+  );
+}
+
+const styles = StyleSheet.create({
+  confirmBadge: {
+    backgroundColor: 'rgba(51, 189, 27, 0.5)',
```
mrtnzlml: Badges should be partially transparent (with blur but we can skip the blur for now)...

app/MMB/src/mmbList/CityImage.js
```diff
@@ -0,0 +1,122 @@
...
+  };
+
+  render = () => (
+    <View style={styles.container}>
+      <NetworkImage
+        source={{
+          uri: this.props.imageUrl,
+        }}
+        style={styles.image}
+        resizeMode="stretch"
```
mrtnzlml: Is this really the best mode? It looks little bit funny on large/small devices. I would go for one with _"scale the image uniformly"_ (see: https://facebook.github.io/react-native/docs/image.html#resizemode).

app/shared/src/badge/AdaptableBadge.js
```diff
@@ -50,7 +62,11 @@ export default function AdaptableBadge(props: Props) {
     <View style={[style.container, props.style]}>
       {props.icon}
       <Text style={[style.text, props.textStyle]}>
-        <Translation passThrough={props.text} />
+        {props.text !== undefined ? (
+          <Translation passThrough={props.text} />
+        ) : (
+          props.translation
```
mrtnzlml: Can we just be strict here and accept only translations instead of this? (maybe send separate PR first)

app/shared/src/image/StretchedImage.js
```diff
@@ -14,5 +14,11 @@ const styles = StyleSheet.create({
 });
 
 export default function StretchedImage(props: Object) {
```
mrtnzlml: Time to write exact type with allowed values we actually use? 🙂

app/MMB/src/mmbList/CityImage.js
```diff
@@ -0,0 +1,122 @@
+// @flow
```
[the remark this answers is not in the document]
tbergq: Suggestions?

app/MMB/src/mmbList/CityImage.js
```diff
@@ -0,0 +1,122 @@
...
+        </View>
+      </View>
+    </View>
+  );
+}
+
+const styles = StyleSheet.create({
+  container: {
+    height: 152,
+    width: '100%',
```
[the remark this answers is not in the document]
tbergq: Wouldn't this make it take the size of its parent? so <View style={{width : 400}}>

app/MMB/src/mmbList/CityImage.js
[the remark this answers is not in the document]
mrtnzlml: Just list for example. I don't think it's that important to have MMB twice in the file name/import.

app/MMB/src/mmbList/CityImage.js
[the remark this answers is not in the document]
mrtnzlml: Ok, looks like a good plan... 🙂

app/MMB/src/mmbList/CityImage.js
mrtnzlml: 🙈

app/shared/src/badge/AdaptableBadge.js
tbergq: Yes, I will create separate PR for this 👍

app/MMB/src/mmbList/CityImage.js
```diff
@@ -0,0 +1,122 @@
...
+
+import TravelTypeIcon from './TravelTypeIcon';
+import ImageBadges from './ImageBadges';
+
+type Props = {|
+  imageUrl: string,
+  bookingId: string,
+  type: string,
+  departureCity: string,
+  arrivalCity: string,
```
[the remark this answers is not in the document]
tbergq: What's the benefit?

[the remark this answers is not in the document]
mrtnzlml: There is probably none... 😛 OK. Can I check it again?

tbergq:
@mrtnzlml please check again

app/MMB/src/mmbList/ImageBadges.js
mrtnzlml: @tbergq What about the opacity? Not now? :)

mrtnzlml: Oh I see - you used opacity but it didn't work? Or it was not so obvious? 🤔

tbergq: it looks different using opacity: 0.5, don't know why 🤷‍♂️

tbergq: > we can skip the blur for now
Also I though you meant to remove it

mrtnzlml: Ok, I'll rephrase: there should be an opacity (I didn't see it) but let's skip the blur effect. Sorry... 🙈 (https://zpl.io/2Gmv4n4)

tbergq: Ok, I will check it on monday

mrtnzlml:
Integrated, thanks.

PR #539, opened 2018-05-07
MMB: introduce new 'TitledMenuGroup' component

mrtnzlml:
I prepared every submenu of the "Manage" group in the main menu. It
should be fine from the design point of view but it's not clickable yet.
 This logic will follow...

### DIFF
--- a/app/MMB/src/components/menu/MenuItem.js
+++ b/app/MMB/src/components/menu/MenuItem.js
@@ -7,7 +7,7 @@ import { Translation } from '@kiwicom/mobile-localization';
 
 type Props = {|
   title: React.Element<typeof Translation>,
-  icon: React.Element<typeof TextIcon>,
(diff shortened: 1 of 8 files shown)


app/MMB/src/components/menu/MenuItem.js
```diff
@@ -30,9 +30,12 @@ type Props = {|
  * `------------------------------`
  */
 export default function MenuItem(props: Props) {
+  const styleSheet = createStyleSheet(props);
```
tbergq: Isn't this really just the same as using inline styles? You create a new object on each render

mrtnzlml: What do you propose? I have to create a new object on each render because this object depends on props.

mrtnzlml: @tbergq Updated - please check again.

PR #540, opened 2018-05-07
Add mmb flight list

tbergq:
First step of adding mmb flight list for mobile

Still to go
- tablet layout
- pull to refresh
- separate future & past trips
- Surely something else 🤔

### DIFF
--- a/app/MMB/src/appRegistry/ManageMyBookingPackage.js
+++ b/app/MMB/src/appRegistry/ManageMyBookingPackage.js
@@ -7,6 +7,7 @@ import NavigationStack from '../navigation/NavigationStack';
 type Props = {|
   language: string,
   currency: string,
+  accessToken: string,
 |};
 
 export default class ManageMyBookingPackage extends React.Component<Props> {
--- a/app/MMB/src/list/CityImage.js
+++ b/app/MMB/src/list/CityImage.js
@@ -18,12 +18,13 @@ import ImageBadges from './ImageBadges';
 
 type Props = {|
   imageUrl: string,
-  bookingId: string,
+  bookingId: ?number,
   type: string,
   departureCity: string,
   arrivalCity: string,
-  date: Date,
+  date: ?Date,
   status: string,
+  passengerCount: ?number,
 |};
 
 export default class CityImage extends React.Component<Props> {
@@ -66,7 +67,7 @@ export default class CityImage extends React.Component<Props> {
           </Text>
           <View style={styles.row}>
             <Text style={[styles.text, styles.dateText]}>
-              <Translation passThrough="2" />
+              <Translation passThrough={this.props.passengerCount} />
             </Text>
             <Icon
               name="people"
@@ -115,8 +116,9 @@ const styles = StyleSheet.create({
   },
   dateText: {
     fontSize: 12,
+    opacity: 0.7,
   },
   passengerIcon: {
-    opacity: 1,
+    opacity: 0.7,
   },
 });
--- a/app/MMB/src/list/FlightList.js
+++ b/app/MMB/src/list/FlightList.js
@@ -1,0 +1,125 @@
+// @flow
+
+import * as React from 'react';
+import { ScrollView, View } from 'react-native';
+import { graphql, createRefetchContainer } from 'react-relay';
+import { Translation } from '@kiwicom/mobile-localization';
+import idx from 'idx';
+import { Text, StyleSheet, Color } from '@kiwicom/mobile-shared';
+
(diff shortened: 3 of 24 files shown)


app/MMB/src/list/FlightList.js
```diff
@@ -0,0 +1,274 @@
...
+    finalDestination: idx(flight, _ => _.outbound.arrival.airport.city.name),
+  });
+
+  multicityToFlightDetail = (flight: ?MulticityType) => ({
+    departureDate: idx(flight, _ => _.start.time),
+    departureCity: idx(flight, _ => _.start.airport.city.name),
+    status: idx(flight, _ => _.status),
+    bookingId: idx(flight, _ => _.databaseId),
+    finalDestination: idx(flight, _ => _.end.airport.city.name),
+  });
```
mrtnzlml: Is it possible to split it into 3 components? It would be great even from the Relay fragment perspective (this is actually how we designed the graph). I would use this component only for the decision which one to render. Makes sense?

app/MMB/src/list/FlightList.js
```diff
@@ -0,0 +1,274 @@
...
+                    }
+                  }
+                }
+                arrival {
+                  airport {
+                    city {
+                      name
+                    }
+                  }
+                }
```
mrtnzlml: This repetition leads to the custom fragment which leads to a separate component. Try to **compose** it a little bit differently, please. 🙂

app/MMB/src/list/CityImage.js
```diff
@@ -18,12 +18,13 @@ import ImageBadges from './ImageBadges';
 
 type Props = {|
   imageUrl: string,
-  bookingId: string,
+  bookingId: ?number,
   type: string,
   departureCity: string,
   arrivalCity: string,
-  date: Date,
+  date: ?Date,
```
mrtnzlml: Just to make sure I understand - these changes are because of IDX nullable output, right?

app/MMB/src/appRegistry/ManageMyBookingPackage.js
```diff
@@ -7,6 +7,7 @@ import NavigationStack from '../navigation/NavigationStack';
 type Props = {|
   language: string,
   currency: string,
+  accessToken: string,
```
mrtnzlml: Shouldn't we use context for this? I am not sure if prop-drilling for every `PrivateApiRenderer` is a good idea...

mrtnzlml:
> separate future & past trips

This is luckily ready in GraphQL. Idea was to use it like this:

```
{
  future: allBookings(only: FUTURE) {
    edges {
      cursor
    }
  }
  past: allBookings(only: PAST) {
    edges {
      cursor
    }
  }
}
```

This way you'll get `future` and `past` keys directly in your response... 🎉

app/MMB/src/list/FlightList.js
tbergq: Yes, and it looks a lot nicer 🙂

app/MMB/src/list/CityImage.js
tbergq: Yes, I could do some fallback with new Date(), but anyways, it is passed through moment() later, and moment(null/undefined) => moment(new Date())

app/MMB/src/appRegistry/ManageMyBookingPackage.js
mrtnzlml: Ok, also please try to fix the alignment of oneway/round/multicity arrows...

app/MMB/src/list/MulticityFlight.js
```diff
@@ -0,0 +1,53 @@
...
+  <CityImage
+    imageUrl={props.imageUrl}
+    type={props.type}
+    passengerCount={props.passengerCount}
+    bookingId={idx(props.booking, _ => _.databaseId)}
+    status={idx(props.booking, _ => _.status) || ''}
+    departureCity={idx(props.booking, _ => _.start.airport.city.name) || ''}
+    date={idx(props.booking, _ => _.start.time)}
+    arrivalCity={idx(props.booking, _ => _.end.airport.city.name) || ''}
+  />
```
mrtnzlml: Great simplification would be to allow `CityImage` to fetch data (create fragment). But maybe you want it pure - that's fine. In this case, you can have `CityImage` and `CityImageContainer` (fragment itself). The benefit of this approach is that you'll do just `...CityImage_data` in every component (oneway/multicity/return). Please think about it during your following refactorings...

mrtnzlml:
Integrated, thanks.

PR #541, opened 2018-05-07
Eslint: remove 'react-native/no-inline-styles' rule

mrtnzlml:


tbergq:
Why do we want to remove this, should we rather change it to error and fix the 12 tests that are using inline styles?

mrtnzlml:
The question is - is it really such a big issue for us? What do you prefer?

PR #544, opened 2018-05-08
Align city image icon to center

### LINKED ISSUE #542: Fix city image arrow icon alignment

tbergq:
closes #542


app/MMB/src/list/cityImage/CityImage.js
```diff
@@ -52,31 +50,15 @@ export default class CityImage extends React.Component<Props> {
...
-        </View>
+        <FromToRow
+          departureCity={this.props.departureCity}
+          arrivalCity={this.props.arrivalCity}
+          type={this.props.type}
+        />
+        <DateAndPassengerCount
+          formattedDate={this.formatDate()}
+          passengerCount={this.props.passengerCount}
+        />
```
mrtnzlml: This is way more readable... 👍

mrtnzlml:
Integrated, thanks. I still think that you could take bigger advantage from Relay fragments though... 🙂

tbergq:
> Integrated, thanks. I still think that you could take bigger advantage from Relay fragments though

I am open for suggestions, out of my head, I don't know how this fragment component should look, since there are 3 different flight types

mrtnzlml:
A good way how to start is to check [these fragments](https://github.com/kiwicom/mobile/blob/aef7cab814e9509eda5c203f71a445267f78cd63/app/MMB/src/list/__generated__/FlightsQuery.graphql.js#L53). You can spot that they are very similar. At least these parts:

```
arrival {
  airport {
    city {
      name
    }
  }
}
```

And:

```
departure {
  time
  airport {
    city {
      name
    }
  }
}
```

It looks different for multicity but it's still the same in fact because you can easily create fragment on `RouteStop` type. This always helps you to design your components better. In this case, the arrival/destination row could fetch data needed to render itself and similar applies to `DateAndPassengerCount` component. I hope I am not wrong but looks like it's easily possible.

I am usually trying to decompose it and always fetch only what's needed for the component (via fragment). I can try to send PR if you are lost... 🙂

PR #545, opened 2018-05-08
Add accessToken to context

tbergq:
closes #545

### DIFF
--- a/app/MMB/src/appRegistry/ManageMyBookingPackage.js
+++ b/app/MMB/src/appRegistry/ManageMyBookingPackage.js
@@ -1,15 +1,19 @@
 // @flow
 
 import * as React from 'react';
+import { AuthContext } from '@kiwicom/mobile-relay';
 
 import NavigationStack from '../navigation/NavigationStack';
 
 type Props = {|
-  language: string,
   currency: string,
   accessToken: string,
 |};
 
 export default class ManageMyBookingPackage extends React.Component<Props> {
-  render = () => <NavigationStack screenProps={this.props} />;
+  render = () => (
+    <AuthContext.Provider accessToken={this.props.accessToken}>
+      <NavigationStack />
+    </AuthContext.Provider>
+  );
 }
--- a/app/MMB/src/list/Flights.js
+++ b/app/MMB/src/list/Flights.js
@@ -7,9 +7,7 @@ import { graphql } from 'react-relay';
 import type { FlightsQueryResponse } from './__generated__/FlightsQuery.graphql';
 import FlightList from './FlightList';
 
-type Props = {|
-  accessToken: string,
-|};
+type Props = {||};
 
 export default class Flights extends React.Component<Props> {
   renderInner = (innerProps: FlightsQueryResponse) => {
@@ -19,7 +17,6 @@ export default class Flights extends React.Component<Props> {
   render = () => (
     <PrivateApiRenderer
       render={this.renderInner}
-      accessToken={this.props.accessToken}
       query={graphql`
         query FlightsQuery {
           ...FlightList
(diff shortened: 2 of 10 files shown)


app/MMB/src/__tests__/MMBContext.test.js
```diff
@@ -0,0 +1,39 @@
...
+    expect(consumerProps.setAccessToken).toBeDefined();
+    expect(consumerProps.accessToken).toBe(null);
+  });
+
+  it('sets the accessToken', () => {
+    const wrapper = renderer.create(
+      <MMBContext.Provider>{null}</MMBContext.Provider>,
+    );
+    const instance = wrapper.getInstance();
+    instance.setAccessToken('myToken');
```
mrtnzlml: @tbergq What about using context action directly? This way you'll actually test the real API because nobody is going to use it like this in real code...

app/MMB/src/appRegistry/ManageMyBookingPackage.js
```diff
@@ -11,5 +12,9 @@ type Props = {|
 |};
 
 export default class ManageMyBookingPackage extends React.Component<Props> {
-  render = () => <NavigationStack screenProps={this.props} />;
+  render = () => (
+    <MMBContext.Provider>
+      <NavigationStack screenProps={this.props} />
+    </MMBContext.Provider>
```
mrtnzlml: This is little bit odd. You should set `accessToken` here directly and completely get rid of `setAccessToken` action. We wanted to avoid prop-drilling but this is not very convincing...

tbergq:
@mrtnzlml please check again

app/MMB/src/appRegistry/ManageMyBookingPackage.js
```diff
@@ -11,5 +12,13 @@ type Props = {|
 |};
 
 export default class ManageMyBookingPackage extends React.Component<Props> {
-  render = () => <NavigationStack screenProps={this.props} />;
+  render = () => (
+    <MMBContext.Provider
+      accessToken={this.props.accessToken}
+      language={this.props.language}
+      currency={this.props.currency}
```
mrtnzlml: Where is `language` and `currency` actually used? For example, the way how we should handle languages completely changed in our GraphQL. Are we preparing here for something that is maybe already invalid? 🙂

tbergq: currency is beeing used atleast here
zep

I will remove language, I guess you are right. It is not in use.

app/MMB/src/list/Flights.js
```diff
@@ -6,20 +6,21 @@ import { graphql } from 'react-relay';
...
+export class Flights extends React.Component<PropsWithContext> {
   renderInner = (innerProps: FlightsQueryResponse) => {
     return <FlightList data={innerProps} />;
   };
 
   render = () => (
     <PrivateApiRenderer
       render={this.renderInner}
-      accessToken={this.props.accessToken}
+      accessToken={this.props.accessToken || ''}
```
mrtnzlml: It should work a little bit differently. You didn't maybe notice but we have 2x `PrivateApiRenderer` component. You are using the one directly from `mobile-relay` but you should never do that. There is also other one in `app/core/src/components/relay/PrivateApiRenderer.js` - this component should be consuming the context (probably not MMB context). This way you don't have to deal with access tokens in your code at all (look at how are you prop-drilling the `accessToken` property here just because of `PrivateApiRenderer` component). 🙂

Changes are probably going to be necessary but the idea is clear - do not work with access tokens in your components. I can explain in person if you want...

tbergq:
@mrtnzlml  please check again

PR #547, opened 2018-05-08
MMB: create PoC of the navigation

mrtnzlml:
It behaves differently on mobile and tablet: mobile always open a new
stack but tablet only changes the container content. This makes it quite
 challenging because tablet have to change the title as well as
 sometimes allow to open new scene in stack (for map for example).

### DIFF
--- a/app/MMB/src/navigation/MMBScreen.js
+++ b/app/MMB/src/navigation/MMBScreen.js
@@ -1,11 +1,13 @@
 // @flow
 
 import * as React from 'react';
-import type { NavigationType } from '@kiwicom/mobile-navigation';
+import { HeaderTitle, type NavigationType } from '@kiwicom/mobile-navigation';
+import { AdaptableLayout, WebView } from '@kiwicom/mobile-shared';
+import { Translation } from '@kiwicom/mobile-localization';
 
 import Layout from '../components/Layout';
 import MainMenu from '../MainMenu';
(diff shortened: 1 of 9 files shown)


app/MMB/src/navigation/NavigationStack.js
```diff
@@ -6,20 +6,35 @@ import {
...
     MMB: {
-      screen: MMBScreen,
+      screen: withMappedProps(MMBScreen),
     },
+    ...Screens,
   },
   {
     ...StackNavigatorOptions,
-    initialRouteName: 'MMBList',
+    initialRouteName: 'MMB',
```
tbergq: The actual entry point is the `MMBList`

app/MMB/src/navigation/MMBScreen.js
```diff
@@ -16,16 +18,105 @@ type Props = {|
...
-    />
-  );
+type State = {|
+  activeContainerComponent: string,
+|};
+
+export const MenuComponents = {
+  'mmb.help': {
+    screen: HelpSubmenu,
+    headerTitle: function HelpSubmenuHeaderTitle() {
```
tbergq: Any reason for not using arrow function here?

mrtnzlml: Every component (function) must have a name.

PR #550, opened 2018-05-08
MMB: reduce props drilling

mrtnzlml:


tbergq:
Circle-ci times out, all tests passes locally so I will merge it.

PR #553, opened 2018-05-09
Split past and future flights

tbergq:
Next up => add gray overlay over cityImage past flight

### DIFF
--- a/app/MMB/src/list/FlightList.js
+++ b/app/MMB/src/list/FlightList.js
@@ -1,11 +1,10 @@
 // @flow
 
 import * as React from 'react';
-import { ScrollView, View } from 'react-native';
+import { View } from 'react-native';
 import { graphql, createRefetchContainer } from 'react-relay';
-import { Translation } from '@kiwicom/mobile-localization';
 import idx from 'idx';
-import { Text, StyleSheet, Color } from '@kiwicom/mobile-shared';
+import { StyleSheet } from '@kiwicom/mobile-shared';
 
 import OneWayFlight from './OneWayFlight';
 import ReturnFlight from './ReturnFlight';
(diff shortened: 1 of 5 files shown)


app/MMB/src/list/FlightList.js
```diff
@@ -89,28 +107,45 @@ const styles = StyleSheet.create({
...
+          destinationImageUrl(dimensions: _375x165)
+          type
+          oneWay {
+            ...OneWayFlight_booking
+          }
+          return {
+            ...ReturnFlight_booking
+          }
+          multicity {
+            ...MulticityFlight_booking
```
mrtnzlml: I am sure you don't like these two identical fragments as well. It, in my opinion, indicates wrong components structure. You should compose it a little bit differently: there should be something like `FlightList` with only one fragment and only one responsibility (render the list). Another component should care about the future and past trips. Working example:

```js
export default class Flights extends React.Component<Props> {
  renderInner = (innerProps: FlightsQueryResponse) => {
    return (
      <ScrollView>
        <Translation id="mmb.my_bookings.future_trips" />
        <FlightList data={innerProps.future} />
        <Translation id="mmb.my_bookings.past_trips" />
        <FlightList data={innerProps.past} />
      </ScrollView>
    );
  };

  render = () => (
    <PrivateApiRenderer
      render={this.renderInner}
      query={graphql`
        query FlightsQuery {
          future: allBookings(only: FUTURE) {
            ...FlightList
          }
          past: allBookings(only: PAST) {
            ...FlightList
          }
        }
      `}
    />
  );
}
```

app/MMB/src/list/FlightList.js
```diff
@@ -31,6 +33,7 @@ export class FlightList extends React.Component<Props> {
             type={type}
             imageUrl={imageUrl}
             booking={idx(item, _ => _.node.oneWay)}
+            isPastFlight={isPastFlight}
```
mrtnzlml: This is another indicator of odd design - these conditional props affecting internal logic. Just keep it simple and use one component in multiple places... :)

tbergq:
@mrtnzlml please check again

app/MMB/src/list/FlightList.js
```diff
@@ -63,51 +62,39 @@ export class FlightList extends React.Component<Props> {
...
-        {flights.map(this.renderItem)}
-      </ScrollView>
-    );
+    const flights = idx(this.props.data, _ => _.edges) || [];
+
+    if (flights.length === 0) {
+      return null;
+    }
+
+    return <React.Fragment>{flights.map(this.renderItem)}</React.Fragment>;
```
mrtnzlml: @tbergq Why fragment? Can I remove it?

tbergq: Yes, you are right, the fragment is abundant.

mrtnzlml:
Integrated, thanks. Looks much better now... 🙂

PR #555, opened 2018-05-09
Add styling for past flight on city image

tbergq:
I know it looks nothing like the design, neither does the existing app. I think this is (close to) the best we can do without creating some native code.


tbergq:
We will see if we can handle this on the server. I will close this issue.

mrtnzlml:
@tbergq So can we delete the branch or not yet?

tbergq:
yes, I deleted it now

PR #557, opened 2018-05-10
MMB: introduce new deeplink component

mrtnzlml:
Deeplink component is responsible for fetching the right booking URL and
 displaying MMB webview afterwards.


app/MMB/src/menu/MMBDeeplink.js
```diff
@@ -0,0 +1,47 @@
+// @flow
+
```
tbergq: I think we should try to avoid prefixing all components with MMB

mrtnzlml: @tbergq Improved. Also please check it more heavily since my webview is not working properly again (problem with my machine probably)...

tbergq:
Thanks, tested and works fine as before

PR #559, opened 2018-05-10
Add pull to refresh on mmb list

tbergq:


app/MMB/src/list/FlightListContainer.js
```diff
@@ -0,0 +1,95 @@
+// @flow
+
+import * as React from 'react';
+import {
+  graphql,
+  createRefetchContainer,
+  type RelayRefetchProp,
```
mrtnzlml: @tbergq Where did you find this type? It is not exported from the Relay package...

mrtnzlml: _(and therefore it's exported as `any` and you can call whatever you want on the relay property)_

tbergq:
```
export type RelayRefetchProp = RelayProp & {
    refetch(
        refetchVariables:
            | RelayRuntimeTypes.Variables
            | ((fragmentVariables: RelayRuntimeTypes.Variables) => RelayRuntimeTypes.Variables),
        renderVariables?: RelayRuntimeTypes.Variables,
        callback?: (error?: Error) => void,
        options?: RefetchOptions
    ): RelayRuntimeTypes.Disposable;
};
```
This is in index.ts file from `react-realy`

mrtnzlml:
Integrated, thanks.

PR #560, opened 2018-05-10
TextIcon: use 'code' property instead of 'children'

mrtnzlml:


tbergq:
Looks good, please rebase

PR #561, opened 2018-05-10
Add tablet layout

tbergq:


tbergq:
> Right, let's keep the issue open until these tasks are resolved.

Of course, still one check box unchecked 🙂

mrtnzlml:
I added the sorting.

tbergq:
Sorry, there is one more thing missing, navigation to booking on click, I will update the task

PR #562, opened 2018-05-10
Remove MMB prefix from components

tbergq:
Removing useless prefix naming


app/MMB/src/navigation/__generated__/MMBScreenQuery.graphql.js
```diff
@@ -1,119 +0,0 @@
-/**
```
mrtnzlml: Orphaned file again?

tbergq: yes

PR #564, opened 2018-05-11
Add pdf viewer component and add translations to General error

tbergq:


ios/reactNativeApp/Info.plist
```diff
@@ -35,6 +35,11 @@
 				<key>NSExceptionAllowsInsecureHTTPLoads</key>
 				<true/>
 			</dict>
+			<key>samples.leanpub.com</key>
```
tbergq: Need to whitelist the domain since the pdf in the test is loaded from `http`

app/localization/src/DefaultVocabulary.js
```diff
@@ -30,6 +30,8 @@ const Translations = {
   'core.authentication.login': 'Login!',
   'core.authentication.logout': 'Logout',
 
+  'gallery.gallery_grid.no_images': 'No images available.',
```
mrtnzlml: This gallery is still only for hotels so it should be prefixed as well (see `hotels.gallery.pagination`).

app/shared/index.js
```diff
@@ -70,6 +70,7 @@ export { default as AppStateChange } from './src/AppStateChange';
 export { default as ReadMore } from './src/ReadMore';
 export { default as WithStorage } from './src/WithStorage';
 export { default as GeolocationContext } from './src/GeolocationContext';
+export { default as PdfViewer } from './src/PdfViewer';
```
mrtnzlml: Can we maybe create separate PDF package (workspace)? I would like to start splitting the shared package into multiple workspaces because of our problem with circular dependencies.

app/shared/src/__tests__/__snapshots__/PdfViewer.test.js.snap
```diff
@@ -0,0 +1,39 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`handles errors 1`] = `
```
mrtnzlml: This test is weird - it doesn't really test `handles errors` case, right?

tbergq: You are right, it works as a visual test when rendered in the playground, but as a snapshot it is weird, is it better to remove it?

mrtnzlml: I would prefer to remove it or really test the `onError` callback.

tbergq:
@mrtnzlml please check again

mrtnzlml:
I am not able to start playground:

screen shot 2018-05-11 at 11 16 38

app/shared/package.json
```diff
@@ -8,8 +8,10 @@
     "@kiwicom/mobile-localization": "^0",
     "@kiwicom/react-native-native-modules": "^0.1.25",
     "@ptomasroos/react-native-multi-slider": "^0.0.12",
+    "react-native-fetch-blob": "^0.10.8",
     "react-native-image-progress": "^1.1.0",
     "react-native-modal": "^5.4.0",
+    "react-native-pdf": "^3.0.8",
```
mrtnzlml: This can be probably removed now.

app/shared/src/errors/__tests__/GeneralError.test.js
```diff
@@ -2,11 +2,16 @@
...
+import { Translation } from '@kiwicom/mobile-localization';
 
 import GeneralError from '../GeneralError';
 
 it('renders as expected', () => {
   expect(
-    renderer.create(<GeneralError errorMessage="Error Message" />),
+    renderer.create(
+      <GeneralError
+        translation={<Translation passThrough="Error Message" />}
```
mrtnzlml: I am not sure if this property renaming is a good idea. It should say what is this property about and not what type should you use...

app/shared/package.json
tbergq: I forgot, it should be in pdf package.json yes

tbergq:
@mrtnzlml please check again

PR #567, opened 2018-05-11
MMB menu: make 'MenuItem' component activable

mrtnzlml:


app/MMB/src/menu/components/MenuItem.js
```diff
@@ -35,20 +37,35 @@ type Props = {|
...
+ *  - Android: only left box border is in the brand color
  */
-export default function MenuItem(props: Props) {
+function MenuItem(props: Props) {
+  const styleSheet = createStyleSheet(props);
+
+  let styledIcon = null;
+  if (props.icon) {
+    styledIcon = React.cloneElement(props.icon, {
+      style: Platform.select({
```
tbergq: Why is not the `normal` PlatforStyleSheet enough here?

mrtnzlml:
@tbergq Much better with the suggested changes - ready to be checked again.

PR #569, opened 2018-05-11
MMB: complete "Other" menu

mrtnzlml:
Detail screen is now connected to the real booking ID and everything in
"Other" section should be finished.


tbergq:
Please fix the bookingId null issue

mrtnzlml:
@tbergq Fixed - I didn't use the context because the fix was really simple.

PR #576, opened 2018-05-14
Add flight services component

tbergq:


app/MMB/src/menu/submenu/FlightServices.js
```diff
@@ -0,0 +1,60 @@
+// @flow
```
tbergq: I though it made sense to add a submenu folder, so we might skip suffixing every component with submenu, agree or you think we should keep current structure?

mrtnzlml: Yes, makes sense. We have to evolve the structure as we go...

PR #581, opened 2018-05-14
MMB: update menu design to be Android compatible

mrtnzlml:


tbergq:
Will this work for groups with one child?

mrtnzlml:
I think this bug is not fixed yet - this just updates the design to the Android-friendly:

screen shot 2018-05-14 at 15 13 27

tbergq:
Ok, will you please make a task for it, so I can keep track of it?

PR #583, opened 2018-05-14
MMB: evolve structure of all files to be more readable

mrtnzlml:
This is my new proposed structure:

```
src/
├── components/                       - generic components reused accross all scenes in MMB
├── navigation/                       - React Navigation related files
├── scenes/                           - navigation independent (more or less) scenes
└── MainMenu.js                       - main file with defined items
```


tbergq:
Definitely a big improvement, thanks 🙂

PR #588, opened 2018-05-15
Wrap scroll view with refresh options

tbergq:


mrtnzlml:
I think we should probably move our view to the separate package. What is the best name for it? Should we create something only for UI and therefore `@kiwicom/mobile-ui`? Or something more obvious? I would also like to create "typography" package...

tbergq:
Yes, I think you are right, it could be good to split this package yet again...

Not sure `mobile-ui` is the best name, could be almost anything? Not sure what is the better name yet though...
´@kiwicom/mobile-typography` is descriptive and good for the text components :)

PR #589, opened 2018-05-15
Pass booking opaque id to detail screen

tbergq:


mrtnzlml:
@tbergq Why is it actually needed?

tbergq:
🙈 You are right I was fooled that the type says `ID!`

but I see now in the description `id: ID! Both Booking id & databaseId can be used as ID.`

mrtnzlml:
🙈 Ah, sorry - next time you can write your intention into the description and I can help you faster... 😄

PR #595, opened 2018-05-15
Not showing status open as ordered service

### LINKED ISSUE #522: Flight services
[Specification in confluence](https://confluence.kiwi.com/display/MOB/Manage+my+booking+specification#Managemybookingspecification-Flightservices)
[Zeplin](https://zpl.io/b6JPdGy)
img

## Screenshot from current app

img_5070

## Tasks 
- [x] add component
- [x] connect webviews
- [x] separate ordered/not ordrered
- [x] pull to refresh

## Questions
- Icons?
- Urls to webviews?
- When should this modal open?
skjermbilde 2018-05-03 kl 15 05 20
- How do we know if a service has been ordered?
- Is there any api that provides these services? So we can dynamically add them instead of hardcoding?

tbergq:
closes #522 

From specs: 
`Ordered services are displayed only if the final_status of the services is "closed", "pending" or "confirmed"`

### DIFF
--- a/app/MMB/src/scenes/flightServices/FlightServicesMenuGroup.js
+++ b/app/MMB/src/scenes/flightServices/FlightServicesMenuGroup.js
@@ -106,13 +106,21 @@ export class FlightServicesMenuGroup extends React.Component<Props, State> {
   getOrderedAndNotOrderedServices = () => {
     const bookedServices =
       idx(this.props.bookedServices, _ => _.bookedServices) || [];
-    const categories = bookedServices.map(item => idx(item, _ => _.category));
 
+    // A service is considered to be ordered if it is a booked service and status is not open
+    // Here we are extracting the category of each ordered service with status not equal to open
+    const orderedCategories = bookedServices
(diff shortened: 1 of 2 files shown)


app/MMB/src/scenes/flightServices/FlightServicesMenuGroup.js
```diff
@@ -106,13 +106,16 @@ export class FlightServicesMenuGroup extends React.Component<Props, State> {
   getOrderedAndNotOrderedServices = () => {
```
mrtnzlml: Please comment this behavior even in the code - the logic is getting fairly complicated.

tbergq: See if my comments makes sense 🙈

mrtnzlml:
Integrated, thanks - I just changed the comments. The point is not to comment what is this line about because we are programmers. The point is to explain the idea behind this algorithm so everyone will understand during the refactoring that services with status `OPEN` are somehow special...

PR #596, opened 2018-05-15
MMB: add final design for status bar in the header

mrtnzlml:


app/MMB/src/components/header/Header.js
```diff
@@ -21,6 +21,7 @@ export default class Header extends React.Component<Props> {
     return (
       <View style={styleSheet.wrapper}>
         <StatusBar data={booking} />
+        <View style={styleSheet.separator} />
```
tbergq: We have quite a few of these now I think. Does it make sense to make a component, or are they all different somehow?

mrtnzlml: Will do. We have basically 2 and this one for iOS is actually similar to the separator for Android.

mrtnzlml: Done... 🙂

PR #603, opened 2018-05-16
MMB: Add passenger name, nationality, date of birth on passenger card

tbergq:
Adding another piece of the puzzle, still work to be done

### DIFF
--- a/app/MMB/src/scenes/passenger/Passenger.js
+++ b/app/MMB/src/scenes/passenger/Passenger.js
@@ -1,0 +1,122 @@
+// @flow
+
+import * as React from 'react';
+import { graphql, createFragmentContainer } from '@kiwicom/mobile-relay';
+import { Translation, DateFormatter } from '@kiwicom/mobile-localization';
+import { SimpleCard, StyleSheet, Text, Color } from '@kiwicom/mobile-shared';
+import idx from 'idx';
+import { View } from 'react-native';
+import startCase from 'lodash/startCase';
+
+import { SeparatorFullWidth } from '../../components/Separators';
+import type { Passenger_passenger as PassengerType } from './__generated__/Passenger_passenger.graphql';
+
+const Row = ({ children }: {| children: React.Node |}) => (
+  <View style={styles.row}>{children}</View>
+);
+
+type ColumnProps = {|
+  title: React.Element<typeof Translation>,
+  value: React.Element<typeof Translation>,
+|};
+
+const Column = (props: ColumnProps) => (
+  <View style={styles.column}>
+    <Text style={styles.titleLight}>{props.title}</Text>
+    {props.value}
+  </View>
+);
+
+type Props = {|
+  passenger: PassengerType,
+|};
+
+export class Passenger extends React.Component<Props> {
+  render = () => {
+    const { passenger } = this.props;
+    if (passenger == null) {
+      return null;
+    }
+    const title = idx(passenger, _ => _.title) || '';
+    const fullName = idx(passenger, _ => _.fullName) || '';
+    const birthday = idx(passenger, _ => _.birthday) || '';
+    const nationality = idx(passenger, _ => _.nationality) || '';
+    const idNumber = idx(passenger, _ => _.travelDocument.idNumber) || '';
+    const insuranceType = startCase(
+      (idx(passenger, _ => _.insuranceType) || '').toLowerCase(),
+    );
+    return (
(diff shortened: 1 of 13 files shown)


mrtnzlml:
After I click on "Details":

screen shot 2018-05-16 at 17 28 57

tbergq:
Has the graphql passengers commit been deployed on the server? It kind of depends on that this field exists

mrtnzlml:
How did you develop it? 😄 No, private deployment failed again. Deployment in progress...

tbergq:
I used localhost

app/MMB/src/scenes/passenger/Passenger.js
```diff
@@ -0,0 +1,121 @@
...
+    const firstname = idx(passenger, _ => _.firstname) || '';
+    const lastname = idx(passenger, _ => _.lastname) || '';
+    const birthday = idx(passenger, _ => _.birthday) || '';
+    const nationality = idx(passenger, _ => _.nationality) || '';
+    const idNumber = idx(passenger, _ => _.travelDocument.idNumber) || '';
+    const insuranceType = startCase(
+      (idx(passenger, _ => _.insuranceType) || '').toLowerCase(),
+    );
+    return (
+      <SimpleCard style={StyleSheet.flatten(styles.card)}>
```
mrtnzlml: Yeah, we should really refactor the styles of `SimpleCard`... 🙈

app/MMB/src/scenes/passenger/Passenger.js
```diff
@@ -0,0 +1,121 @@
...
+        <Text style={styles.name}>
+          <Translation passThrough={`${title}. ${firstname} ${lastname}`} />
```
mrtnzlml: What about leaving this to our GraphQL? Field `fullName` would be better since this can be quite complicated logic and there is going to be more places where we need the full name...

app/MMB/src/scenes/passenger/Passenger.js
```diff
@@ -0,0 +1,121 @@
...
+          <RowItem
+            title={<Translation id="mmb.passenger.birthday" />}
+            value={
+              <Translation passThrough={DateFormatter(birthday).format('L')} />
+            }
+          />
+        </View>
+        <SeparatorFullWidth />
+        <View style={[styles.row, styles.insuranceRow]}>
+          <RowItem
```
mrtnzlml: I was a little bit confused by `RowItem` because it's actually a column. What about creating `Row` and `Column` components here and wrapping `SeparatorFullWidth` with vertical margin 15?

app/MMB/src/scenes/passenger/Passenger.js
tbergq: sounds doable

app/MMB/src/scenes/passenger/Passenger.js
tbergq: It is an item in a row, the item uses column layout yes

app/MMB/src/components/Separators.js
```diff
@@ -11,8 +11,8 @@ import { Color, StyleSheet } from '@kiwicom/mobile-shared';
  * |---------------------------|
  * |                           |
  */
-export function SeparatorFullWidth() {
-  return <View style={styleSheet.separator} />;
+export function SeparatorFullWidth({ style }: { style?: StylePropType }) {
+  return <View style={[styleSheet.separator, style]} />;
```
mrtnzlml: First of all, I don't like this change. You can just wrap your `SeparatorFullWidth` in a `View` if you need it. No need for another property... 🙂 But more importantly, the FLow type should be exact. Same for `Row` component.

app/MMB/src/scenes/passenger/PassengerDetail.js
```diff
@@ -77,3 +77,9 @@ export default createRefetchContainer(
     }
   `,
 );
+
+const styles = StyleSheet.create({
+  container: {
+    marginTop: 20,
```
mrtnzlml: Maybe we should add this to the content container instead of setting it on every page?

tbergq: That could definetly be an improvement, but separate PR I guess, since it touches also other screens?

mrtnzlml: I believe it affects only `Layout` component, but OK. 🙂

tbergq:
@mrtnzlml please check again

PR #605, opened 2018-05-17
Fix implementation of DateFormatter and DateUtils

mrtnzlml:
DateFormatter now uses Intl behind the scenes and it should therefore be
 locale friendly.


app/localization/src/DateFormatter.js
```diff
@@ -1,17 +1,61 @@
...
+}
+
+/**
+ * Purpose of this wrapper is to make very opinionated and restrictive
+ * formatter. Every function is localized and you should never call formatting
+ * on the raw Date object.
+ *
+ * TODO: we should probably accept custom Value Object
+ */
+function DateFormatter(rawDate: Date) {
```
tbergq: could we do `rawDate: Date = new Date()` here as well?

mrtnzlml: Why? There is not a single use-case for it and more importantly, it looks weird:

```
DateFormatter().formatToDate();
```

_"What date are you formatting here?"_ So, is it a good idea? I wanted to be explicit here...

tbergq: This is how moment works, you can do `moment().format('L')`

mrtnzlml: That's not an argument for me... 😄

tbergq: Well, I will have to live with it then 🙈

mrtnzlml: I will change it. As I said there is not a single use-case of `DateFormatter(new Date())` in our codebase... 🙂

tbergq: Ok, OCD calming.. 😄 Ping me when it is done

mrtnzlml: Your OCD should be calmed down now. I also introduced new `formatToBirthday` in order to format birthdays. Not sure about the name but it's good enough at this point...

mrtnzlml:
@tbergq 🚀

mrtnzlml:
This PR is very approved... 😄

tbergq:
I'm a click-jacker 🙈 🌮 🎉

PR #610, opened 2018-05-17
Whole new iOS infrastructure for embedding within brownfield apps

grabbou:
This pull request is a result of our internal brainstorming on how to fix the existing setup that uses RNHotelsBundler to download this repository, wrap it up within the `framework` and use inside the app. 

This process is cumbersome and slow (around 10 minutes) and makes it really hard for others to contribute to the React Native distribution and the glue code.

As opposed to the existing setup, we would like to propose a whole new infrastructure within this PR that embraces existing iOS design patterns and tooling.

Below is the list of changes, one by one. Please do not get scared of the number of files and lines touched - it's mainly import of `RNHotelsBundler`.

- We have removed unnecessary targets (like empty automatically generated unit tests)
- We have removed the need for RNHotelsBundler
- Framework (called RNKiwiMobile) is now a separate target inside the existing `.xcodeproj` making the integration really easy - it all now lives in the single file
- It reuses the existing CocoaPods installation making the dependencies in sync and easy to manage
- There is also RNNativePlayground (as opposed to existing RNPlayground) that uses the native home screen to demonstrate `brownfield` integration points. We will use this file to keep implementing features requested by the native team, like the bridge caching
- There is `RNKiwiOptions` interface that includes `initialParameters` (previously called `RNHotelsInitiaParameters), `moduleName` (use it to set a name of component you want to run - we will need it to run other modules in the native app once they are ready) and `jsCodeLocation` (in case you want to run from localhost and see the changes to the Javascript live. Otherwise, by default, it will use the bundle location.

The whole process is now super fast (reuses cache) and encourages to prototype - you can change files and see it live in the playground. 

As a next step, we can also automate it and make new distribution of the framework to be published automatically along the existing `fastlane` setup.

To build the framework, you just need to do it from within the Xcode.

**Important note to the iOS team**: Check the `RNNativePlayground` files for how to integrate your existing native app with the React Native one. The name of files are slightly different and the configuration object is now different to allow better extension in the future. Once this PR lands, I am happy to do the integration work for you - it's less than 5 minutes.


tbergq:
> To build the framework, you just need to do it from within the Xcode

Any instructions for this? I am not very familiar with Xcode. 

Also, how can I use the framework to see that it works?

grabbou:
Open up the xcodeproj, select RNNativePlayground from the top where you
choose device to run the app on and see it working.

The framework builds behind the scenes for you.

To do it yourself, select it from that drop down, it's called RNKiwiMobile.

If you have troubles finding that piece, look for the ability to change the
iOS target.

grabbou:
We can attach framework here on Github as a release artifact so that your
native to pository can just get it with one line of code, instead of
downloading and copying file manually.

But that's another story.

IOS team let me know how would you see this automated.

mrtnzlml:
@tbergq Are you going to be able to describe it in our documentation, please? It would great to wait for the review from iOS guys and describe _"how to"_ so everyone will understand it (even me 🙂).

tbergq:
> We can attach framework here on Github as a release artifact so that your
native to pository can just get it with one line of code, instead of
downloading and copying file manually.

That sounds like a very good idea

@mrtnzlml Not sure yet, need to get familiar with the solution, then maybe 🤷‍♂️

tbergq:
skjermbilde 2018-05-17 kl 14 41 18

Build doesn't work

lukewalczak:
Did you run `pod install` ?

tbergq:
yes

grabbou:
Can someone else try running the PR? @tbergq I saw your issue, but we cannot reproduce it locally for now. Let's work on resolving that one tomorrow.

tbergq:
We need to solve it before we can merge it, even if it works for someone else. I also want to keep working on the project 🙂

ferrannp:
Hey @tbergq can you try:

1. `sudo gem install cocoapods` and make sure you have cocoapods version `>=1.5.0`
2.  Make sure you have XCode updated to `>=9.3.x.`
3. `rm -rf node_modules` and `rm -rf ios/Pods` (as pods come from `node_modules` and we do not have `Podfile.lock` that might be a problem
4. `yarn install && cd ios && pod update && pod install`
5. `rm -rf ios/build`
6. Try to build with XCode again with the target `RNNativePlayground` (it should take a while)

grabbou:
If this doesn't work, let's connect on Skype or something and see what's
going on.

tbergq:
I think we need to make a call on slack yes, when do you have time?

tbergq:
Ok, I had to clone the repo into a new folder to make it work 🤷‍♂️ 

Please rebase and squash your commits.

ferrannp:
@tbergq should be ready.

tbergq:
So if you ever come across someone with a similar problem, for me what worked was deleting `~/Library/Developer/Xcode/DerivedData`

tbergq:
Sorry, one more thing before we merge this PR. 

We need to update the readme section of how to build and how to run the project

tbergq:
So what I need to know is how will the native team consume the package, and how do we provide it to them?

pistelak:
@grabbou  Ok, I looked at it and I have a few questions:

At first, for running this project I have to use `yarn` to install all RN (and JS) dependencies, am I right? I think, @ferrannp did a really good job to remove this necessity. Why are we leaving this solution? 

At second, you have removed `.build.sh` script and I don't see any build phase for creating universal (FAT) library, so how this actually works? I mean, there is note "To build the framework, you just need to do it from within the Xcode.", but it is not so easy. You have to create two frameworks one for simulator and one for real devices and then combine them together... Without this we are not able to use simulator (or device)!

`RNNativePlayground` is cool because testing new changes in `RNKiwiMobile` (or `RNHotelsBundled`) is not so simple as it should be, but it would be better without http://localhost:8081/index.bundle?platform=ios&dev=true as `codeLocation`. 

---

I think, `RNNativePlayground` is handy, but the second point has to be solved and I was really happy for @ferrannp's solution, so... :(

ferrannp:
@pistelak we are working on it. We still can provide the `build.sh` to build both frameworks. On the other hand `http://localhost:8081/index.bundle?platform=ios&dev=true as codeLocation.` this is handy because you can change JS code and you will see the changes in the playground without rebuilding the app.

The codeLocation line should be commented by default though, if you comment the line, it will default to the bundle. So this `localhost` should only be used to play with it (dev).

pistelak:
Ok, ok. I've just wanted to point out that this is not **currently** usable (due to the second point) and it is also easier to create framework because you don't have to use `yarn` now.

grabbou:
Thanks Radek for giving some light on what is your use case and the way you
are utilizing this framework.

We will provide you with a script that will do same things as previously as
well as give us the ability to automate it.

Do you think a setup where we create a Git tag with the framework as an
artifact is something you would be interested in using?

pistelak:
Why not 🤔  It sounds like less work for me... :)

grabbou:
> At first, for running this project I have to use yarn to install all RN (and JS) dependencies, am I right? I think, @ferrannp did a really good job to remove this necessity. Why are we leaving this solution?

Previously, we were building native wrapper based on the released to `npm` hotels bundle. This gives the ability to build directly from source hence the necessity of this step. It's not that tricky and has the benefit of being able to include other RN components within the framework as well. 

I just committed the same `buildFramework.sh` script that creates FAT framework automatically w/o need of `yarn` or any other command.

grabbou:
As a preparation to support multiple apps, I decided to implement the `react-native bundle` command as a `Run Script` phase under XCode.

zrzut ekranu 2018-05-20 o 20 14 36

As you can see, it uses `Input files` to determine what entry points you want to include in your framework as well as `Output files` to know under what name we should save them.

Thanks to this setup, we have nice control over what gets included inside the framework and what's not.

grabbou:
In order to build the universal framework, you have to run `./scripts/buildIOSFramework.sh` from the root folder. 

By default, the example uses `pre-bundled` Javascript file to demonstrate that the aforementioned script for bundling works without issues.

grabbou:
Note to the owners of the repo: You can turn on `squash and merge` option and we will not need to squash commits manually. This is a nice improvement to the workflow and makes it easy to perform inline hot-fixes.

grabbou:
This is in my opinion ready to merge. 

Next steps:
- Figure out the automation process (shall we automatically run this on CircleCI when build is green and make it "a release artifact" so you can just download it from there?) Or maybe create a Git release and upload it here.
- Figure out how we are going to support multiple bundles in the future. That is, whether we are going to have separate `jsbundles` for every part of the app vs having single bundle with multiple components. I believe we want the latter since former will make it hard to reuse single bridge and the memory allocations. Either way, our code currently supports both use-cases (multiple bundles via `Input scripts` and `jsCodeLocation` as well as multiple components through `moduleName`).

ferrannp:
@grabbou I also remember @pistelak asking me about removing unnecessary fonts from vector icons. Would you be ok if I add into `scripts/buildIOSFramework.sh` a step where it delete from `Pods` the unnecessary fonts (we just need `MaterialIcons.ttf` right before creating the framework?

grabbou:
You should add a new Run Phase to `Xcode` called `Remove extraneous assets` and reference a newly created `scripts/removeExtraneousAssets.sh` from there. That way, the `IOSFramework.sh` just builds the framework and the preparation steps happen inside Xcode.

grabbou:
FYI I merged PR #618 that adds releases script to fully automate the stuff.

You can run `yarn release` and this will make a Github release and put framework as an artifact. Make sure you have `GITHUB_TOKEN` in your environment.

Otherwise, to do a local release w/o Github, just run `./scripts/buildIOSFramework.sh`.

To see it separately, just check the aforementioned PR.

That said, this is ready to get merged and we will implement automation on CircleCI at some point later.

tbergq:
> This is in my opinion ready to merge.

If someone now asked me to release a new version to our mobile team, I would not know how to. I need this question answered before we can merge.

PR #612, opened 2018-05-17
MMB header: add duration into 'TripTimes' component

mrtnzlml:


tbergq:
Looks good, just a litte observation. Here we have 4h 0m, The existing app has 4h, and design has 6h 00m. Not sure what we should use? 🤷‍♂️ 
skjermbilde 2018-05-18 kl 08 02 19

Also the `h` and `m` should be passed through some translation, in norwegian i expect to see `t` and `m`

mrtnzlml:
> Looks good, just a litte observation. Here we have 4h 0m, The existing app has 4h, and design has 6h 00m. Not sure what we should use? 🤷‍♂️

I chose compromise... 🙂 

Translations updated.

mrtnzlml:
Working on the unhappy tests.

mrtnzlml:
Tests fixed.

PR #616, opened 2018-05-18
MMB: add cities to the trip info panel

mrtnzlml:


app/MMB/src/components/header/Location.js
```diff
@@ -0,0 +1,34 @@
+// @flow
+
+import * as React from 'react';
+import { graphql, createFragmentContainer } from '@kiwicom/mobile-relay';
+import { Translation } from '@kiwicom/mobile-localization';
+import { Text, StyleSheet } from '@kiwicom/mobile-shared';
+
+function Location(props) {
```
tbergq: why is there no flow type here?

app/MMB/src/components/header/Location.js
```diff
@@ -0,0 +1,34 @@
...
+  return (
+    <Text style={styleSheet.city}>
+      <Translation passThrough={props.data.airport.city.name} />
```
tbergq: Shouldn't you use idx here?

mrtnzlml: Why is Flow happy? 🤔

tbergq: flow is broken 🙈

tbergq:
Also, should there be a little more margin between the rows?

app/MMB/src/components/header/Location.js
mrtnzlml: I saw just a few cases where Flow was actually broken in last year. The problem is usually between the keyboard and chair like in this case - our `createFragmentContainer` HOC is the reason why it doesn' work... 🤦‍♂️

mrtnzlml:
> Also, should there be a little more margin between the rows?

Will do. See my standup notes - separators are going to be next (it's a little bit more complicated).

tbergq:
> separators are going to be next

that's fine

mrtnzlml:
@tbergq `Location` types fixed. Thanks...

PR #617, opened 2018-05-21
Shared bridge

grabbou:
This pull request targets my `work in progress` branch (that is subject to another PR). Until it gets merged to master, this pull request should not be merged.

==

This pull request implements shared bridge to improve startup of our iOS applications. It introduces singleton class, called `RNKiwiSharedBridge` that holds all the bridges created for bundle URLs. 

Since we haven't settled yet whether to go with multiple bundles (and bridges) or single bundle (and multiple components), this PR supports both use cases. 

I believe that even if we settle on a single bundle, the current implementation that uses NSDictionary is still reasonable given that the method signature is `bridgeForBundleURL` and assumes we accept returning multiple bridges.

Now, how to load bridge upfront to result from cache benefits - to do so, you have two options, depending on your preferences:

1) Call `[[RNKiwiSharedBridge] sharedInstance] initBridgeWithOptions:options]` where options are `RNKiwiOptions`, same used for `RNKiwiViewController`

2) Init an instance of `RNKiwiViewController` upfront, in constructor of its parent view controller, ahead of time. RNKiwiViewController automatically inits the bridge when it gets created and so, if you do it upfront, you basically mimic the former point.

### DIFF
--- a/ios/RNKiwiMobile/RNKiwiSharedBridge.m
+++ b/ios/RNKiwiMobile/RNKiwiSharedBridge.m
@@ -1,0 +1,53 @@
+#import "RNKiwiSharedBridge.h"
+#import "RNKiwiOptions.h"
+#import <React/RCTBridge.h>
+#import <React/RCTRootView.h>
+
+@interface RNKiwiSharedBridge()
+
+@property (nonatomic, strong) NSMutableDictionary<NSString*,RCTBridge*> *bridges;
+
+@end
+
+@implementation RNKiwiSharedBridge
+
++ (id)sharedInstance {
+  static dispatch_once_t pred = 0;
+  __strong static id _sharedObject = nil;
+  dispatch_once(&pred, ^{
+    _sharedObject = [[self alloc] init];
+  });
+  return _sharedObject;
+}
+
+- (instancetype)init {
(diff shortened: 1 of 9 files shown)


ios/RNKiwiMobile/RNKiwiSharedBridge.m
```diff
@@ -0,0 +1,55 @@
...
+  if (self) {
+    _bridges = [NSMutableDictionary new];
+  }
+  return self;
+}
+
+- (NSURL *)resolveBundleURL:(id<RNKiwiOptions>)options {
+  return [options respondsToSelector:@selector(jsCodeLocation)]
+    ? [options jsCodeLocation]
+    : [[NSBundle bundleForClass:[self class]] URLForResource:@"hotels" withExtension:@"jsbundle"];
```
ferrannp: Mmm maybe it has more sense to make this mandatory (to respond to selector) and do not have `hotels` as default? WDYT?

grabbou: Yeah, I think we need to expose few constants from this module so that users can just write `jsCodeLocation = [RNKiwiOptions hotelsBundle]`.

PR #618, opened 2018-05-21
Better ci

grabbou:
Implements a new workflow for releasing. 

Instead of pushing a new version to `npm`, it makes Github release and then, uploads the framework as an artifact. Once we get Android to good shape, we will get rid of `npm` step entirely in favor of this one (or not - we will see once we dive deep).

This is still WIP - need to fix the step #2, which is to build the framework itself. For now, I tested uploads with a dummy file.

Feedback appreciated in the meantime.


grabbou:
You can see an example release created with this script here: https://github.com/kiwicom/mobile/releases/tag/cd736bc

It contains the iOS universal framework that was build from my other branch.

grabbou:
Merging to ios-cleanup branch. Feel free to look at it separately for better clarity.

PR #619, opened 2018-05-21
Native module to control swiping gestures

lukewalczak:


app/hotels/src/appRegistry/HotelsStandalonePackage.js
```diff
@@ -17,7 +19,18 @@ type Props = {|
   checkout?: string,
 |};
 
+type State = {
+  key: string,
+  isTransitioning: boolean,
+  index: number,
+  routes: mixed[],
+};
```
grabbou: Do we need to type state here? Probably `*` will do just fine? If not, maybe in GestureController. Right now, you have it defined twice.

app/hotels/src/appRegistry/HotelsStandalonePackage.js
```diff
@@ -17,7 +19,18 @@ type Props = {|
...
+
 export default class HotelsStandalonePackage extends React.Component<Props> {
+  navigationController = (previousState: State, currentState: State) => {
```
grabbou: I guess `onNavigationStateChange` makes more sense as a name

app/hotels/src/appRegistry/HotelsStandalonePackage.js
```diff
@@ -17,7 +19,18 @@ type Props = {|
...
+    GestureController(previousState, currentState, 'KiwiHotels');
```
grabbou: I think the logic should happen here. This method should call either `GestureController.addGestures` or `GestureController.removeGestures`. Gesture Controller should have no knowledge of `navigation` state since we want to use it across different navigators in the future.

app/shared/src/GestureController.js
```diff
@@ -0,0 +1,27 @@
...
+type State = {
+  key: string,
+  isTransitioning: boolean,
+  index: number,
+  routes: mixed[],
+};
+
+import { NativeModules } from 'react-native';
+
+export default function GestureController(
```
grabbou: As I said, instead of a function, make it a class or an object with appropriate methods.

ios/RNKiwiMobile/RNHotelsViewController.h
```diff
@@ -0,0 +1,49 @@
+//
+//  RNHotelsViewController.h
+//  RNHotels
+//
+//  Created by Radek Pistelak on 07/02/2018.
```
grabbou: Remove this whole comment block

ios/RNKiwiMobile/RNKiwiGestureController.m
```diff
@@ -0,0 +1,18 @@
+#import <Foundation/Foundation.h>
+#import "RNKiwiGestureController.h"
+
+@implementation RNKiwiGestureController
+
+RCT_EXPORT_MODULE(RNKiwiGestureController);
+
+RCT_EXPORT_METHOD(removeGestures:(NSString *)moduleName) {
+  [[NSNotificationCenter defaultCenter] postNotificationName:@"RNKiwiRemoveGestures"
```
grabbou: Good practice - you can move this to header file as a `constant`, so that we do not use plain strings. You can see this in https://github.com/facebook/react-native/blob/master/React/Base/RCTBridge.h#L41

Look for usage of `RCTJavaScriptDidFailToLoadNotification` to see how it's defined. Also, when you define it, make sure to comment it when it's called ;)

ios/RNKiwiMobile/RNKiwiViewController.m
```diff
@@ -0,0 +1,160 @@
...
+  if ([self.flowDelegate respondsToSelector:@selector(RNKiwiViewControllerDidStopControllingGestures:)]) {
+    [self.flowDelegate RNKiwiViewControllerDidStopControllingGestures:self];
+  }
+}
+
+- (void)startObservingGestures {
+  [[NSNotificationCenter defaultCenter]
+   addObserver:self
+   selector:@selector(removeGestures)
+   name:@"RNKiwiRemoveGestures"
```
grabbou: I think it's a bit confusing the naming convention we have here. We are not removing and adding gestures. In fact, we are just expressing our intent to do so.

I am not sure what's the right approach - let's think together tomorrow. But the fact that we are calling `didStartControlling` from `removeGestures` method makes me confused.

ios/RNKiwiMobile/RNKiwiViewController.m
```diff
@@ -0,0 +1,160 @@
...
+   object:[_options moduleName]];
+  
+  [[NSNotificationCenter defaultCenter]
+   addObserver:self
+   selector:@selector(addGestures)
```
grabbou: Please indent it using XCode. The `:` should be aligned.

ios/RNKiwiMobile/RNKiwiViewController.m
```diff
@@ -0,0 +1,160 @@
...
+   name:@"RNKiwiAddGestures"
+   object:[_options moduleName]];
+}
+
+- (void)stopObservingGestures {
+  [[NSNotificationCenter defaultCenter]
+   removeObserver:self
```
grabbou: You can use `[[NSNotificationCenter defaultCenter] removeObserver:self];` to unsubscrive from all notifications. No need to write it many times.

PR #623, opened 2018-05-22
Create a NativeModule to control swiping gestures

lukewalczak:


app/hotels/src/appRegistry/HotelsStandalonePackage.js
```diff
@@ -18,6 +19,16 @@ type Props = {|
 |};
 
 export default class HotelsStandalonePackage extends React.Component<Props> {
+  navigationController = (previousState: *, currentState: *) => {
+    if (currentState.index === 0) {
+      // enable gestures
+      GestureController.addGestures('KiwiHotels');
```
grabbou: I think enable/disable is better then add/remove if you want to do a quick search repalce :)

app/shared/src/GestureController.js
```diff
@@ -0,0 +1,15 @@
...
+
+export default {
+  addGestures: NativeModules.RNKiwiGestureController
+    ? NativeModules.RNKiwiGestureController.addGestures
+    : noop,
+
+  removeGestures: NativeModules.RNKiwiGestureController
+    ? NativeModules.RNKiwiGestureController.removeGestures
+    : noop,
+};
```
grabbou: Personal preference: 
```js
// This module is only present when this app is running in context of the native mobilje app
const { addGestures = noop, removeGestures = noop } = NativeModules.RNKiwiGestureController || {};
export default { addGestures, removeGestures };
```

app/shared/src/__tests__/GestureController.test.js
```diff
@@ -0,0 +1,14 @@
...
+import GestureController from '../GestureController';
+
+describe('GestureController', () => {
+  it('should call NativeModules.RNKiwiGestureController.addGestures', () => {
+    GestureController.addGestures('KiwiHotels');
+    expect(
+      NativeModules.RNKiwiGestureController.addGestures,
+    ).toHaveBeenCalledWith('KiwiHotels');
+  });
+});
```
grabbou: I think what is more important to test (and nice) would be to make sure that `noop` is called when there is no native module present.

To test it, we should probably not mock `NativeModules.RNKiwiGestureController` in one suite.

ios/RNKiwiMobile/RNKiwiGestureController.m
```diff
@@ -0,0 +1,22 @@
+
+#import <Foundation/Foundation.h>
+#import "RNKiwiGestureController.h"
+
+NSString *const RNKiwiAddGestures = @"RNKiwiAddGestures";
+NSString *const RNKiwiRemoveGestures = @"RNKiwiRemoveGestures";
```
grabbou: I like that :) As a general "good practice", I suggest adding one doc block that describes when this gets called (or should be)

grabbou: We can also mention it is scoped to `moduleName`

app/shared/src/__tests__/GestureController.test.js
```diff
@@ -0,0 +1,28 @@
+// @flow
+
+import { NativeModules } from 'react-native';
+
+describe('GestureController', () => {
+  it('should call NativeModules.RNKiwiGestureController.enableGestures', () => {
+    const GestureController = require('../GestureController').default;
+
```
tbergq: Could this be `Import GestureController from '../GestureController'` instead?

lukewalczak: It's better to stick with my solution because in the second test suite I'm resetting modules, so still, we will have to require `GestureController`

app/shared/src/__tests__/GestureController.test.js
```diff
@@ -0,0 +1,29 @@
...
+      NativeModules.RNKiwiGestureController.enableGestures,
+    ).toHaveBeenCalledWith('KiwiHotels');
+  });
+
+  it('should call noop when there is no native module present', () => {
+    jest.resetModules();
+
+    const OriginalGestureController = NativeModules.RNKiwiGestureController;
+
+    /* $FlowFixMe */
```
tbergq: Please use `// $FlowExpectedError:` and explain why you are suppressing the error

app/shared/src/__tests__/GestureController.test.js
```diff
@@ -0,0 +1,29 @@
...
+    ).toHaveBeenCalledWith('KiwiHotels');
+  });
+
+  it('should call noop when there is no native module present', () => {
+    jest.resetModules();
+
+    const OriginalGestureController = NativeModules.RNKiwiGestureController;
+
+    /* $FlowExpectedError: undefined is incompatible with object from NativeModules */
+    NativeModules.RNKiwiGestureController = undefined;
```
tbergq: Obviously it is incompatible with object from native modules, that's why we are suppressing it.

lukewalczak: I'm deleting now `NativeModules.RNKiwiGestureController`

app/hotels/src/appRegistry/SingleHotelStandalonePackage.js
```diff
@@ -21,6 +22,14 @@ type Props = {|
 export default class SingleHotelStandAlonePackage extends React.Component<
   Props,
 > {
+  onNavigationStateChange = (previousState: *, currentState: *) => {
```
tbergq: Please create a type

tbergq:
Thanks, looks good now from my point of view. 

Please squash commits.

tbergq:
What happend here? why is it now 134 files changed?

ferrannp:
^ @tbergq it's solved.

app/hotels/src/appRegistry/SingleHotelStandalonePackage.js
```diff
@@ -5,6 +5,7 @@ import * as React from 'react';
 import RootComponent from './RootComponent';
 import SingleHotelStack from '../navigation/singleHotel/SingleHotelStack';
 import type { RoomsConfiguration } from '../singleHotel/AvailableHotelSearchInput';
+import GestureController from '../../../shared/src/GestureController';
```
tbergq: same as above

lukewalczak:
^Fixed

PR #625, opened 2018-05-22
MMB: Add visa required warning

tbergq:


app/MMB/src/scenes/passenger/Passenger.js
```diff
@@ -48,40 +49,43 @@ export class Passenger extends React.Component<Props> {
...
-            value={<Translation passThrough={idNumber} />}
-          />
-          <Column
-            title={<Translation id="mmb.passenger.insurance" />}
-            value={<Translation passThrough={insuranceType} />}
-          />
-        </Row>
-      </SimpleCard>
+      <React.Fragment>
+        <SimpleCard style={styles.card}>
```
mrtnzlml: Did you try it on Android? `SimpleCard` component is with radiuses so I cannot imagine this works well together... 🙂

ios/reactNativeApp.xcodeproj/project.pbxproj
```diff
@@ -480,46 +480,6 @@
 			shellPath = /bin/sh;
 			shellScript = "export NODE_BINARY=node\n../node_modules/react-native/scripts/react-native-xcode.sh";
 		};
-		2711985D20B1EA8100C6292B /* Bundle React Native dependencies */ = {
```
mrtnzlml: Is this intentional?

tbergq: 🤷‍♂️ happens when I build

tbergq: This is ok according to @ferrannp

tbergq:
@mrtnzlml please check again

app/MMB/src/components/Alert.js
```diff
@@ -0,0 +1,68 @@
...
+      return null;
+    case 'success':
+      // TODO
+      return null;
+    default:
+      return null;
+  }
+}
+
+const RED = '#d0021b';
```
mrtnzlml: Should we use new brand colors here or not? https://zpl.io/2GmG6mW

screen shot 2018-05-23 at 09 52 43

tbergq: 🤷‍♂️ are there any designs for these `alerts` in the new design?

mrtnzlml: Only this: https://zpl.io/bzYGpjz

The colors are different though...

tbergq: The color codes in the new `alerts` are the same as before

app/MMB/src/components/Alert.js
```diff
@@ -0,0 +1,68 @@
...
+      <Icon name="warning" size={18} color={RED} />
+    </View>
+    <View>
+      <Text style={[styles.titleText, styles.titleDanger]}>{props.title}</Text>
+      <View style={styles.body}>{props.children}</View>
+    </View>
+  </View>
+);
+
+export default function Alert(props: Props) {
```
mrtnzlml: Is this component necessary? Cannot we just use the "danger alert" directly?

tbergq: The idea was to expose one component, with 3 types, then that component selects the necessary one. 

Basically these components will be the same, with different styling.

PR #627, opened 2018-05-22
Use new styles in all for elements

mrtnzlml:


app/shared/src/forms/TextInput.js
```diff
@@ -2,17 +2,84 @@
...
+  onChangeText?: (text: string) => void,
+  keyboardType?: 'email-address',
+  secureTextEntry?: boolean,
+|};
+
+type State = {|
+  displayPlaceholder: boolean,
+|};
+
+export default class TextInout extends React.Component<Props, State> {
```
mrtnzlml: I rewrote this component so we can use `Translation` in the `placeholder` prop.

tbergq: Do you think you could also rename it to `TextInput`?

mrtnzlml: Yes, this is not intentional... :)

app/shared/src/forms/TextInput.js
```diff
@@ -36,26 +102,17 @@ const styles = StyleSheet.create({
     marginLeft: 10,
     alignSelf: 'center',
   },
+  placeholder: {
+    position: 'absolute',
+    padding: 10,
+    paddingVertical: 15,
+    // this should be the same with 'input' above
+    color: Color.textLight,
```
tbergq: I don't understand this comment, color should be the same as input above? But it is not, it is textLight and input has textDark

app/hotels/src/filter/FilterButton.js
```diff
@@ -12,46 +12,58 @@ type Props = {
...
+    });
+  }
+
+  return (
+    <View style={filterButtonStyles.wrapper}>
+      <Button
+        title={title}
+        onPress={onPress}
+        icon={icon}
+        styles={buttonStyles}
```
mrtnzlml: Refactor of the buttons styles will follow.

app/shared/src/forms/TextInput.js
mrtnzlml: This should be resolved now. Please check again.

PR #637, opened 2018-05-23
[DO NOT MERGE] CircleCI to release iOS framework when merging into master

ferrannp:


scripts/releaseFramework.js
```diff
@@ -10,34 +12,39 @@ const child_process = require('child_process');
...
-          name: version,
-          draft: false,
-          prerelease: false,
-        }),
-      }
-    )
-  ).json();
+  const release = await (await fetch(
+    `https://api.github.com/repos/kiwicom/mobile/releases?access_token=${
+      // $FlowFixMe env.GITHUB_TOKEN will be a string
```
mrtnzlml: Fix it directly please - there is a reason for this Flow error. See for example: https://github.com/facebook/flow/issues/1192

ferrannp: Ok once I have the PR actually working (WIP) I'll do that.

mrtnzlml:
@ferrannp I noticed the build (release job) is green even though it's failing. It would be great to make it red so we know something is wrong... 🙂

ferrannp:
As we already have the script in `master`, let's do this in separated PR.

PR #638, opened 2018-05-23
MMB: Add contact details

tbergq:


app/MMB/src/scenes/passenger/Passenger.js
```diff
@@ -3,30 +3,16 @@
...
-|};
-
-const Column = (props: ColumnProps) => (
-  <View style={styles.column}>
-    <Text style={styles.titleLight}>{props.title}</Text>
-    {props.value}
-  </View>
-);
+import Row from './Row';
+import Column from './Column';
```
mrtnzlml: This is sooo common. Maybe we should extract it to our new package with our basic building blocks? 🙂

tbergq: I guess, but, right now it is just being used by passenger. We could probably make it more generic and move it to shared when needed.

PR #640, opened 2018-05-23
MMB: Make MenuItem onPress optional

tbergq:


app/MMB/src/components/menu/MenuItem.js
```diff
@@ -10,22 +10,27 @@ import {
...
   // icon on the left
   icon?: React.Element<typeof TextIcon>,
 
   // icon on the right (arrow by default)
   actionIcon?: React.Element<typeof TextIcon>,
   description?: React.Element<typeof Translation>,
+
+  // style props
+  titleStyle?: StylePropType,
+  descriptionStyle?: StylePropType,
```
mrtnzlml: Why is it needed? This component is opinionated about styling to prevent style hotfixing...

tbergq: So I should create a new one then? Since my menu item has completely different styling than the existing one

mrtnzlml: > my menu item has completely different styling than the existing one

Then it's completely different component? Can you please share screens for comparison? This PR is about _"MMB: Make MenuItem onPress optional"_ but you are tweaking styles here... 🤔

app/MMB/src/components/menu/MenuItem.js
```diff
@@ -54,37 +59,48 @@ type Props = {|
...
-    </Touchable>
+      </View>
+      {/* Hide icon if menu item is not clickable */}
+      {props.onPress && IconRight}
+    </React.Fragment>
   );
+
+  if (props.onPress) {
+    return (
+      <Touchable onPress={props.onPress} style={styleSheet.wrapper}>
```
mrtnzlml: There is also `disabled` property but it requires dummy `onPress` anyway.

tbergq:
Basically it is this without the arrow
skjermbilde 2018-05-24 kl 14 13 47

vs existing

skjermbilde 2018-05-24 kl 14 14 39

tbergq:
In my opinion, same component with different styling

app/MMB/src/components/menu/MenuItem.js
tbergq: I think we already use similar approach in `SimpleCard`

mrtnzlml:
Feels like different component, yes. Please try to compose components instead of adding props. You can use:

```js
<MenuGroup>
  <MenuItem title={...} desc={...} />
  <MenuItem title={...} desc={...} />
</MenuGroup>
```

Or:

```js
<MenuGroup>
  <DoubleMenuItem title={...} desc={...} />
  <WhateverYouWant />
</MenuGroup>
```

Vs:

```js
<MenuGroup>
  {/* accepts strictly MenuItem components */}
  <MenuItem title={...} desc={...} titleStyle={...} descriptionStyle={...} />
  <MenuItem
    title={...}
    desc={...}
    titleStyle={...}
    descriptionStyle={...}
    someOtherPropBecauseItsStillNotEnough={...}
  />
</MenuGroup>
```

I think the first two examples are OK. The last is highway to hell... :)

tbergq:
Right, I will make a new component

mrtnzlml:
> In my opinion, same component with different styling

You can reuse parts of it, that's fine... 🙂

Issue #642, opened 2018-05-23
RFC: separate apps and packages


mrtnzlml:
I propose the following directory structure:

```
.
├── app/
│   ├── core/        - ???
│   ├── hotels/
│   └── MMB/
└── packages/
    ├── accessibility/
    ├── localization/
    ├── navigation/
    |    ...
    └── shared/
```

Directory `app/` would contain only the independent applications and `packages/` would be the true Yarn workspaces. Currently, we are mixing it together (our apps are not workspaces).

cc @tbergq

tbergq:
Yes, that sounds like a good idea

PR #646, opened 2018-05-24
Split all files into 'app' and 'packages' directories

mrtnzlml:
Closes https://github.com/kiwicom/mobile/issues/642

Our Playground is really pissing me off now.


scripts/activatePlayground.js
```diff
@@ -28,33 +28,40 @@ YellowBox.ignoreWarnings([
...
+  },
+);
+
+Promise.all([appPromise, packagePromise]).then(() => {
+  const playgroundImportsFileContent = [
+    '// @flow',
+    '',
+    playgroundTestFiles
+      .map(
+        // Replacing app from filepath with ../.. which is the path from playgroundList component to 'packages' folder
```
tbergq: This comment does not seem to match what is happening

tbergq:
> Our Playground is really pissing me off now.

What should we do with it?

mrtnzlml:
Invest more time...

tbergq:
> Invest more time...

😄

scripts/activatePlayground.js
mrtnzlml: Fixed. I would appreciate if you could just fix these minor issues and typos if you are trying it locally. You are a merge master - you can do it... 🙂

tbergq: Sure, makes sense, I will do it for the next ones 👍

PR #647, opened 2018-05-24
Localization: Make translation promise cancellable

tbergq:
Sometimes we get warning, can't call set state on unmounted component.
This happens when we leave a screen before all translations promises has resolved.

Making the promise cancellable should remove this warning.

### DIFF
--- a/packages/localization/src/CancellableTranslation.js
+++ b/packages/localization/src/CancellableTranslation.js
@@ -1,0 +1,29 @@
+// @flow
+
+import { NativeModules } from 'react-native';
+
+export type TranslationPromise = {|
+  cancel: () => void,
+  promise: Promise<string>,
+|};
+
+export function cancellableTranslation(nativeKey: string): TranslationPromise {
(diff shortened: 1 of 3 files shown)


app/localization/src/Translation.js
```diff
@@ -61,25 +65,35 @@ export default class Translation extends React.Component<Props, State> {
     this.setTranslatedString();
   };
 
+  componentWillUnmount = () => {
+    if (this.translatePromise !== null) {
+      this.translatePromise.cancel();
+    }
```
mrtnzlml: Did you consider this pattern? https://github.com/kiwicom/mobile/blob/8275d9d0f2eec3d2c70cb0ea78a040dfe07eafa0/app/shared/src/view/AdaptableLayout.js#L55 (I think it's little bit more clean)

tbergq:
@mrtnzlml  please check again

mrtnzlml:
Thanks for fixing it!

PR #651, opened 2018-05-24
MMB: Add other type of menu item

tbergq:
These type of menu items are only (as far as I can see) used in the Passenger detail screen. It makes sense to keep them in the passenger folder.

If needed, we can extract them to the menu items folder later.

### DIFF
--- a/app/MMB/src/scenes/passenger/menu/TitledMenuGroup.js
+++ b/app/MMB/src/scenes/passenger/menu/TitledMenuGroup.js
@@ -1,0 +1,32 @@
+// @flow
+
+import * as React from 'react';
+import { Platform } from 'react-native';
+import { Translation } from '@kiwicom/mobile-localization';
+
+import MenuGroup from './MenuGroup';
+import MenuGroupTitle from '../../../components/menu/MenuGroupTitle';
+
+type Props = {|
+  title: React.Element<typeof Translation>,
+  children: React.Node,
+|};
+
+// This is basically a copy of /components/TitledMenuGroup
+// But it needs a different type of separator, so it uses ad different MenuGroup to wrap its children
+export default function TitledMenuGroup({ title, children }: Props) {
+  let newChildren = Array.isArray(children) ? children : [children];
+
+  if (Platform.OS === 'android') {
+    // we have to copy every children in order to prepend group title for Android
+    newChildren = [<MenuGroupTitle key="android-ftw" title={title} />];
+    React.Children.forEach(children, child => newChildren.push(child));
+  }
+
+  return (
+    <React.Fragment>
+      {Platform.OS === 'ios' && <MenuGroupTitle title={title} />}
+      <MenuGroup>{newChildren}</MenuGroup>
+    </React.Fragment>
+  );
+}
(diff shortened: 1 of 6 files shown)


packages/playground/src/PlaygroundImports.js
```diff
@@ -11,6 +11,7 @@ import '../../../app/hotels/src/allHotels/searchForm/locationPicker/__tests__/Su
 import '../../../app/hotels/src/allHotels/searchForm/locationPicker/__tests__/RecentSearches.test.js';
 import '../../../app/hotels/src/allHotels/__tests__/LoadMoreButton.test.js';
 import '../../../app/MMB/src/scenes/passenger/visa/__tests__/VisaRequired.test.js';
+import '../../../app/MMB/src/scenes/passenger/menu/__tests__/OtherTitledMenuGroup.test.js';
```
mrtnzlml: This seems outdated. Thankfully... :)

tbergq: Yes, that is true 🤕 I will update the playground imports

app/MMB/src/scenes/passenger/menu/TitledMenuGroup.js
```diff
@@ -0,0 +1,32 @@
...
+import MenuGroup from './MenuGroup';
+import MenuGroupTitle from '../../../components/menu/MenuGroupTitle';
+
+type Props = {|
+  title: React.Element<typeof Translation>,
+  children: React.Node,
+|};
+
+// This is basically a copy of /components/TitledMenuGroup
+// But it needs a different type of separator, so it uses ad different MenuGroup to wrap its children
```
mrtnzlml: I don't like these copy-pastes. It's far from being optimal. 😞 I would actually prefer the code before over this. 😞 I think you are working on this, right?

screen shot 2018-05-25 at 09 52 44

I expected something like this:

- RED: titled group is the same (it will have only one element and one responsibility - do the title and card styles)
- BLUE: one new child that will create these multibuttons (double, tripe with one arrow) - you don't have to deal with separators here because there is going to be only one child (in the RED)
- GREEN: new menu item with the new styles (you could decompose `MenuItem` and partially reuse it)

screen shot 2018-05-25 at 09 52 44

Do you want to help with the design here?

tbergq: Well, I am a little bit lost here, If I should not change the existing components, and not copy it's behaviour and not create new one, I don't really know what to do.

mrtnzlml: > I should not change the existing components, and not copy it's behaviour and not create new one

Yes, this is what I just wrote. 🙄 It's fine to modify and reuse the components and I literally wrote I expected new components. I just don't think that the right way how to modify the components is to add new props for every internal element like `titleStyle`, `descStyle`... (see: https://github.com/kiwicom/mobile/pull/640)

Wanna do pair programming for this?

tbergq: > Wanna do pair programming for this?
sure 🙂

PR #657, opened 2018-05-25
Core: Add login through profile tab

tbergq:


app/core/src/screens/LoginScreen.js
```diff
@@ -0,0 +1,53 @@
...
+
+type Props = {|
+  navigation: NavigationType,
+  coordinates: null | {|
+    latitude: number,
+    longitude: number,
+  |},
+  checkin?: string,
+  checkout?: string,
+|};
```
mrtnzlml: Why do we need all this on the login page?

app/core/src/screens/LoginScreen.js
```diff
@@ -0,0 +1,53 @@
...
+        <Login onLogin={this.onLogin} />
+        <Touchable
+          onPress={this.logout}
+          style={{
+            width: '100%',
+            backgroundColor: Color.red.normal,
+            height: 50,
+            justifyContent: 'center',
+            alignItems: 'center',
+          }}
```
mrtnzlml: Don't be lazy with the inline styles here... 😊

tbergq:
@mrtnzlml updated 🙂

README.md
```diff
@@ -61,17 +61,7 @@ yarn ios --simulator="iPad Pro (10.5-inch)"
...
-```
-mutation {
-  login(email: "...@kiwi.com", password: "...") {
-    token
-  }
-}
-```
-
-Then copy the token and paste it into MMB_TOKEN in the `.env` file. Do not forget to restart the packager (`yarn start`).
+You need to login through the profile tab first, then navigate back to Bookings tab.
```
mrtnzlml: Can we remove this completely? I think it's obvious now... :)

PR #659, opened 2018-05-25
MMB: Add contact details, refactor menu group

tbergq:


app/MMB/src/components/menu/MenuGroup.js
```diff
@@ -2,13 +2,12 @@
 
 import * as React from 'react';
 import { View, Platform } from 'react-native';
-import { SimpleCard } from '@kiwicom/mobile-shared';
 
 import { SeparatorFullWidth, SeparatorTrimmed } from '../Separators';
 
 type Props = {|
   children: React.Node,
-  withoutIcons: boolean,
```
mrtnzlml: Did we remove this during our refactoring? It's wrong - the separators are not correct in the "Help" for example (they should be short).

tbergq: Yes you did, I will check the "Help"

tbergq: Added `short` custom separator to help and other menu items

PR #661, opened 2018-05-25
MMB: Fix raw text bug on flight services

### LINKED ISSUE #660: Text not wrapped in text tag somewhere on flightservices
Causes android to crash

tbergq:
closes #660


mrtnzlml:
Could you please write tests for this fix?

PR #664, opened 2018-05-28
Android: have MainApplication extend MultiDexApplication

tbergq:
Without this change, the application will not run on our test device in debug mode 🤷‍♂️


tbergq:
If you want to know more

https://developer.android.com/studio/build/multidex

PR #673, opened 2018-05-30
MMB: Hide flight services for past trips

### LINKED ISSUE #670: Flight services should not be displayed for past trips

tbergq:
closes #670


app/MMB/src/context/BookingDetailContext.js
```diff
@@ -0,0 +1,48 @@
...
+    setIsPastBooking: () => {},
+  },
+};
+
+const { Consumer, Provider: ContextProvider } = React.createContext({
+  ...defaultState,
+});
+
+type Props = {|
+  children: React.Node,
```
mrtnzlml: Can we start to use read-only props and states more? https://flow.org/en/docs/types/interfaces/#interface-property-variance-read-only-and-write-only-

tbergq: Yes, I guess that makes sense 🙂

app/MMB/src/context/__tests__/BookingDetailContext.test.js
```diff
@@ -0,0 +1,32 @@
...
+            />
+          )}
+        </BookingDetailContext.Consumer>
+      </BookingDetailContext.Provider>,
+    );
+
+    const instance = wrapper.root.findByType(ContextConsumer);
+    instance.props.setIsPastBooking(true);
+
+    expect(instance.props.isPastBooking).toBe(true);
```
mrtnzlml: Nice test 👍 - one small detail - maybe try to test even the `false` value right after this because this assumes that the default value is `false` (this test will be pointless if someone changes it to `true`).

tbergq: Maybe test that it is false before I call set state?

app/MMB/src/context/BookingDetailContext.js
tbergq: Added readonly types

app/MMB/src/context/__tests__/BookingDetailContext.test.js
mrtnzlml: Yes, that's also an option.

tbergq: Implemented and pushed

app/MMB/src/menuGroups/__tests__/ServicesMenuGroup.test.js
```diff
@@ -0,0 +1,32 @@
...
+
+  it('renders with past trip', () => {
+    const wrapper = renderer.create(
+      <ServicesMenuGroup
+        openSubmenu={jest.fn()}
+        activeId="123"
+        isPastBooking={true}
+      />,
+    );
+    expect(wrapper.root.findAllByType(MenuItem).length).toBe(1);
```
mrtnzlml: No snapshots - very nice! 👏 Today is your day... 😊

It may be a little bit wonky in the future but I like it anyway.

app/MMB/src/navigation/DetailScreen.js
```diff
@@ -104,6 +105,10 @@ export default class DetailsScreen extends React.Component<Props, State> {
     };
   };
 
+  componentDidMount = () => {
+    this.props.setIsPastBooking(this.props.isPastBooking);
```
mrtnzlml: We could probably add there even the `bookingId` so we don't have to prop-drill it, right? (not in this PR)

app/MMB/src/menuGroups/__tests__/ServicesMenuGroup.test.js
tbergq: I do agree that snapshots are useless, trying to find the better ways to test with react-test-renderer 🙂

app/MMB/src/navigation/DetailScreen.js
tbergq: exactly, that is an option 🙂

app/MMB/src/context/BookingDetailContext.js
mrtnzlml: Please try to do it everywhere in the future.

app/MMB/src/navigation/DetailScreen.js
```diff
@@ -166,3 +171,21 @@ export default class DetailsScreen extends React.Component<Props, State> {
...
+type Props = {|
+  isPastBooking: boolean,
+  bookingId: string,
+  navigation: NavigationType,
+|};
+
+export default function DetailScreenWithContext(props: Props) {
+  return (
+    <BookingDetailContext.Provider>
+      <BookingDetailContext.Consumer>
```
mrtnzlml: This looks odd. Don't be afraid to use the `Consumer` directly in the component...

tbergq: Then how would I inject the function needed to set the property?

PR #678, opened 2018-05-30
MMB: Add booking id to context

tbergq:


app/MMB/src/components/header/Header.js
```diff
@@ -57,6 +58,14 @@ export default class Header extends React.Component<Props> {
   );
 }
 
+export default function HeaderWithContext() {
+  return (
+    <BookingDetailContext.Consumer>
+      {({ bookingId }) => <Header bookingId={bookingId} />}
+    </BookingDetailContext.Consumer>
```
mrtnzlml: What about just this?

```js
<BookingDetailContext.Consumer>
  {({ bookingId }) => <PrivateApiRenderer ... />}
</BookingDetailContext.Consumer>
```

app/MMB/src/context/__tests__/BookingDetailContext.test.js
```diff
@@ -10,18 +10,22 @@ class ContextConsumer extends React.Component<{}> {
 }
 
 describe('BookingDetailContext', () => {
-  it('injects the isPastBooking prop to consumers', () => {
+  it('injects the props to consumers', () => {
     const wrapper = renderer.create(
-      <BookingDetailContext.Provider isPastBooking={false}>
+      <BookingDetailContext.Provider isPastBooking={false} bookingId="1234">
```
mrtnzlml: The original API is a little different:

```
<Provider value={/* context value here */}>
```

But we don't have to follow it...

tbergq: ^^This is already what we do?

Or maybe I just don't understand what you mean?

mrtnzlml: See: https://reactjs.org/docs/context.html#provider

I think value there is the whole content of our context. Therefore it would be:

```js
<BookingDetailContext.Provider value={{
  isPastBooking: false,
  bookingId: "1234"
}}>
```

tbergq: But we are wrapping it in a class

```
class Provider extends React.Component<Props, State> {
  constructor(props: Props) {
    super(props);

    this.state = {
      isPastBooking: props.isPastBooking,
      bookingId: props.bookingId,
    };
  }

  render = () => (
    <ContextProvider value={this.state}>{this.props.children}</ContextProvider>
  );
}
```

tbergq: Sure, We could do as you suggest

```
<BookingDetailContext.Provider value={{
  isPastBooking: false,
  bookingId: "1234"
}}>
```
and then 
```
class Provider extends React.Component<Props, State> {
  constructor(props: Props) {
    super(props);

    this.state = {
      ...props.value
    };
  }

  render = () => (
    <ContextProvider value={this.state}>{this.props.children}</ContextProvider>
  );
}
```

mrtnzlml: I don't even think we need state here... 🙂

PR #680, opened 2018-05-30
iOS navigation connection with real app

lukewalczak:


tbergq:
Don't really know what is going on here, So I will wait for review from @grabbou also before I merge

grabbou:
LGTM, this is work we have done together to expose interface for mobile app to integrate with React Native. Previous approach was good, but there were certain limitations we have to overcome.

PR #681, opened 2018-05-30
MMB: Disconnect VisaInformation from relay

tbergq:
I need this exact component for showing visa information on the `MainMenu`, but then it is for all passengers and not one. I want to remove relay from the `VisaInformation` to make it reusable.


mrtnzlml:
> I need this exact component for showing visa information on the MainMenu, but then it is for all passengers and not one.

How is it going to work with the countries? Or is it without this information in the menu? Just warnings?

tbergq:
>Passengers details: all documents are filled in. Passengers will be shown in the same order in which they appear in the API. There are also checked in bags and VISA info (visa is not required/visa may be required/visa is required)

This is the specification for passenger details on the detail screen. I just noticed that the existing app does not show countries on the detail screen. 

So in case we need to not show the countries, then I the component will actually be slightly different 🤔

mrtnzlml:
@tbergq Time to decompose? 🙂 You have free hands here so let me know if I should merge this or you are going to adjust it because of this new information.

tbergq:
I need to think a little bit more about the structure. Leave it be until tomorrow.

tbergq:
What do you think about this approach, Now I pass the required and warning component into VisaInformation. This allows me to reuse VisaInformation, with different output for warning/error on main menu.

The other approach would be to just copy/paste the VisaInformation and use other components for warning/error there. 

Also, if you have some 3rd option, it would be great. I am not really happy with any of the 2 above 🙈

mrtnzlml:
First idea:

```js
<Alert />, <Warning /> // generic components

// on the PAX screen
<Alert>
  <Title />
  <OtherThings />
</Alert>

// in the menu
<Alert>
  <Title />
</Alert>
```

You can later wrap it into `VisaInfo` component and connect it to the GraphQL. There are milion ways how to do it. Try something and we'll see... 🙂

mrtnzlml:
What happened here? Are you going a different direction? :)

tbergq:
I though I was, but then it seems I am going back to this direction, so I think it will be reopened

PR #685, opened 2018-05-31
Remove partial errors

mrtnzlml:
We tried for a while and I think it's stupid... :)

Now it throws warnings (in dev mode) instead. I think it's going to be
better to try hard to handle there errors so the user is happy. For
example instead of displaying nothing and showing "we failed" let's
display something like "hey, we are not able to display data here for
some reason".


packages/relay/src/Environment.js
```diff
@@ -46,50 +19,26 @@ export default function createEnvironment(
     networkHeaders.authorization = accessToken;
   }
 
-  const fetchQuery = async (
-    operation,
-    variables,
-  ): Promise<{| data: Object, errors?: $ReadOnlyArray<Object> |}> => {
+  const fetchQuery = async (operation, variables): Promise<FetcherResponse> => {
     // TODO: fetch persisted queries instead (based on operation.id)
```
tbergq: Are we ever going to to this, or is it time to remove this comment?

mrtnzlml: I hope so, yes - I am still waiting for it... 🙂

PR #686, opened 2018-05-31
MMB: Make BookingDetailContext work on mobile

tbergq:


app/MMB/src/navigation/DetailScreen.js
```diff
@@ -141,25 +140,29 @@ export default class DetailsScreen extends React.Component<Props, State> {
...
 }
+
+type Props = {|
+  +navigation: NavigationType,
+|};
+
+export default function DetailScreenWithContext(props: Props) {
+  return (
+    <BookingDetailContext.Consumer>
+      {({ bookingId }) => <DetailsScreen {...props} bookingId={bookingId} />}
```
mrtnzlml: Is it necessary? I see it's only in `openMenuOnMobile` so maybe we don't need it here at all and we can consume it directly where needed? :)

app/MMB/src/scenes/FlightServices.js
```diff
@@ -91,6 +92,16 @@ export default class FlightServices extends React.Component<Props> {
   );
 }
 
+type Props = {|
+  +navigation: NavigationType,
+|};
+
+export default function FlightServicesWithContext(props: Props) {
+  <BookingDetailContext.Consumer>
+    {({ bookingId }) => <FlightServices {...props} bookingId={bookingId} />}
```
mrtnzlml: Similar here - just looks like a special kind of prop-drilling. Navigation prop-drilling... :)

app/MMB/src/scenes/list/cityImage/CityImage.js
```diff
@@ -71,8 +75,24 @@ class CityImage extends React.Component<Props> {
   );
 }
 
+type Props = {|
+  image: ImageType,
+  arrival: ArrivalType,
+  departure: DepartureType,
+  type: 'RETURN' | 'ONE_WAY' | 'MULTICITY',
+  navigation: NavigationType,
+|};
```
mrtnzlml: Also could be read-only.

PR #689, opened 2018-05-31
MMB: Detach VisaInformation from relay

tbergq:
Basically same as #681


app/MMB/src/scenes/passenger/Passenger.js
```diff
@@ -64,7 +71,10 @@ export const Passenger = ({ passenger }: Props) => {
             value={<Translation passThrough={insuranceType} />}
           />
         </Row>
-        <VisaInformation visa={passenger} />
+        <VisaInformation requiredIn={requiredIn} warningIn={warningIn}>
+          <VisaRequired countries={requiredIn} />
+          <VisaWarning countries={warningIn} />
+        </VisaInformation>
```
mrtnzlml: I am slightly confused by reading this. Why `VisaInformation` needs the same parameters as the children components. Doesn't it look odd? 🙂

tbergq: Yes, I totally agree 🙈 

Though I want to be able to reuse the logic to of when to render nothing, render ok, or its children.

PR #692, opened 2018-06-04
MMB: Add header share button

tbergq:


app/MMB/src/components/HeaderShareButton.js
```diff
@@ -0,0 +1,64 @@
...
+import BookingDetailContext from '../context/BookingDetailContext';
+
+type PropsWithContext = {|
+  +arrivalCityId: string,
+  +arrivalTime: Date,
+|};
+
+class HeaderShareButton extends React.Component<PropsWithContext> {
+  onPress = () => {
+    const lang = GetDeviceLocale().split('-')[0] || 'en';
```
mrtnzlml: The way how we work with device locale is starting to be really ugly... 😞

tbergq: You are right, I suggest, we add one method to the native code, to get just language, or native pass it as a prop, and we add it to context.

PR #694, opened 2018-06-04
MMB: add "Lounge" whitelabel scene

mrtnzlml:


packages/shared/src/WebView.js
```diff
@@ -21,12 +21,16 @@ export default class WebView extends React.Component<Props> {
...
-  render = () => (
-    <NativeWebView
-      bounces={false}
-      startInLoadingState={false} // it feels faster because we don't have to wait until the page is fully loaded
-      renderError={this.renderError}
-      {...this.props}
-    />
-  );
+  render = () => {
+    console.warn(this.props.source);
```
mrtnzlml: I added this on purpose. Not sure if it's a good idea though...

app/MMB/src/scenes/tripServices/TripServices.js
```diff
@@ -0,0 +1,144 @@
...
+              ) {
+                booking(id: $bookingId) {
+                  availableWhitelabeledServices {
+                    ...LoungeMenuItem
+                  }
+                }
+              }
+            `}
+            variables={{
+              departureTime: new Date('2019-01-12'), // FIXME
```
mrtnzlml: UPS - I need to fix this! 😄

app/MMB/src/components/CountryFlag.js
```diff
@@ -0,0 +1,37 @@
+// @flow strict
+
+import * as React from 'react';
+import { graphql, createFragmentContainer } from '@kiwicom/mobile-relay';
+import { StyleSheet, NetworkImage } from '@kiwicom/mobile-shared';
+
+import type { CountryFlag as CountryFlagType } from './__generated__/CountryFlag.graphql';
+
+type Props = {|
+  data: CountryFlagType,
```
tbergq: could this be ready only type

app/MMB/src/scenes/tripServices/LoungeMenuItem.js
```diff
@@ -0,0 +1,147 @@
...
+
+  openLoungeLink = (whitelabelURL: string) => {
+    this.setState(
+      {
+        popupVisible: false,
+      },
+      () =>
+        this.props.onOpenWebview(
+          whitelabelURL +
+            (whitelabelURL.indexOf('?') === -1 ? '?' : '&') +
```
tbergq: how about using whitelabelURL.includes('?')

packages/shared/src/WebView.js
tbergq: I am not a very big fan of this yellow noisy warnings. How about a console.log instead, less obtrusive?

mrtnzlml: I will remove it completely.

app/MMB/src/scenes/tripServices/TripServices.js
```diff
@@ -0,0 +1,144 @@
...
+          title={<Translation id="mmb.trip_services.local_services.parking" />}
+          onPress={VoidAction}
+          icon={<TextIcon code="&#xe03e;" />}
+        />
+        <MenuItem
+          title={
+            <Translation id="mmb.trip_services.local_services.transportation" />
+          }
+          onPress={VoidAction}
+          icon={<TextIcon code="<" />} // FIXME: icon?
```
tbergq: Should this also be fixed, or is it for another PR?

mrtnzlml: No, this thing is that I don't know what icons should I use here. I will remove it as well though and I will consider it to be fixed.

mrtnzlml:
@tbergq Round 2...

PR #695, opened 2018-06-04
MMB: add "Parking" whitelabel scene

mrtnzlml:


mrtnzlml:
Note - this is needed: https://github.com/kiwicom/graphql/pull/186

PR #672, opened 2018-05-29
Android RN distribution + back button fix

ferrannp:
### Android
- [x] No need for any of the Gitlab repos used before `react-native-hotels` and `react-native-native-modules`. No need to deploy to NPM anymore
- [x] Create playground app that can be used with the packager too (`yarn start`)

### JS
- [x] Fix native back button: Remove `WithBackButtonListener` HOC in favour of native module solution (like iOS) + both platforms together.
- [x] Fix back button for Toolbar when going back to native screen

You can also check the README: https://github.com/kiwicom/mobile/blob/android-rn-distribution/RNAndroidPlayground/README.md

### Android
* More automation for distribution (pushing SNAPSHOT(s))
* Take versions (like `react-native`, `react-native-maps`, etc) from `package.json`
* Make CircleCI ensures that `rnkiwimobile` and `RNAndroidPlayground` are building correctly

### JS
* Write a test that ensures exact versions for the dependencies we use

### iOS
* When this PR is merged, iOS needs to implement the toolbar back button function. Probably here https://github.com/kiwicom/mobile/pull/698.

PR #698, opened 2018-06-04
Create a new module with the code of react-native-native-modules dependencies

lukewalczak:


tbergq:
Please also squash your commits 🙂

ios/RNKiwiMobile/RNKiwiViewController.m
```diff
@@ -4,10 +4,10 @@
 #import "RNKiwiGestureController.h"
 
 #import <RNNavigator/RNNavigationModule.h>
-#import <RNLogging/RNLoggingModule.h>
-#import <RNTranslationManager/RNTranslationManager.h>
-#import <RNCurrencyManager/RNCurrencyManager.h>
-#import <RNDeviceInfo/RNDeviceInfo.h>
+#import <RNModules/RNLoggingModule.h>
+#import <RNModules/RNTranslationManager.h>
+#import <RNModules/RNCurrencyManager.h>
```
grabbou: Why this import here?

tbergq:
> is it possible to turn on "Squash and merge" feature on this repository

It is turned on actually, but we prefer that you do it your selves, and give the commit a meaningful commit message and not just the one github gives you.

mrtnzlml:
More importantly, it's not obvious whether the result should be really just one commit or let's say 3 for example...

grabbou:
Okay, that makes sense :) Thanks for explanation

ios/Podfile
```diff
@@ -35,14 +31,25 @@ def react_native_pods()
...
 target 'RNPlayground' do
   react_native_pods()
+  react_native_modules()
 end
 
 target 'RNKiwiMobile' do
   react_native_pods()
+  react_native_modules()
 
   pod 'RNNavigator', :path => '../node_modules/@kiwicom/react-native-native-modules'
```
ferrannp: We need to remove this one, we will use same approach as https://github.com/kiwicom/mobile/pull/672.

app/hotels/src/appRegistry/HotelsStandalonePackage.js
```diff
@@ -7,7 +7,7 @@ import RootComponent from './RootComponent';
 import HotelsStack from '../navigation/NavigationStack';
 import type { Coordinates } from '../CoordinatesType';
 
-type Props = {|
+type Props = {
```
mrtnzlml: Why did you change this? Is it necessary? If yes, I am OK with that but otherwise it doesn't make sense.

app/hotels/src/appRegistry/SingleHotelStandalonePackage.js
```diff
@@ -7,7 +7,7 @@ import RootComponent from './RootComponent';
 import SingleHotelStack from '../navigation/singleHotel/SingleHotelStack';
 import type { RoomsConfiguration } from '../singleHotel/AvailableHotelSearchInput';
 
-type Props = {|
+type Props = {
```
mrtnzlml: The same here. We want to use exact types with props as much as possible...

app/hotels/src/appRegistry/HotelsStandalonePackage.js
ferrannp: Yeah it is necessary for now or I would have to use a Flow comment. The `WithNativeNavigation` is injecting `onBackClicked` prop but in your non-native app, you are also using that wrapper that injects `onBackClicked` so it is a bit tricky to make back button works in this case for both not-native and native.

Flow does not seem to understand it:

```js
<WrappedComponent
  onBackClicked={
    GestureController.isNativeGestureModule ? this.onBackClicked : null
  }
  onNavigationStateChange={this.onNavigationStateChange}
  {...this.props}
/>
```

For the native app, it will use the `GestureController`. For the wrapper like `HotelsPackageWrapper`, it will be passing `onBackClicked` which overrides the `null` one with `{...this.props}`. Makes sense?

PR #703, opened 2018-06-05
MMB: Tickets screen, add E-Ticket component

tbergq:


app/MMB/src/scenes/tickets/TicketButton.js
```diff
@@ -0,0 +1,44 @@
...
+} from '@kiwicom/mobile-shared';
+import type { TranslationType } from '@kiwicom/mobile-localization';
+
+type Props = {|
+  disabled: boolean,
+  text: TranslationType,
+  onPress: () => void,
+|};
+
+export default function TicketButton({ disabled, text, onPress }: Props) {
```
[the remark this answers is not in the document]
mrtnzlml: Yes, please refactor the button. I think this is going to be a better way since there will be more places where we'll need disabled button...

tbergq:
@mrtnzlml 👀

packages/shared/src/buttons/Button.js
```diff
@@ -8,9 +8,10 @@ import StyleSheet from '../PlatformStyleSheet';
 import type { StylePropType } from '../../index';
 
 type Props = {|
-  children: React.Node,
-  onPress?: () => void,
-  style?: StylePropType,
+  +children: React.Node,
+  +onPress?: () => void,
+  +style?: StylePropType,
+  +disabled?: boolean,
```
mrtnzlml: We should refactor this (not now). What about making `onPress` mandatory instead?

tbergq: Definitely, It does not make sense to have a button with onPress

mrtnzlml: > It does not make sense to have a button with onPress

Please elaborate... 😄

tbergq: I really should stop drinking while working 🍺 🤦‍♂️ 

It should have been without onPress

PR #704, opened 2018-06-05
Shared: Make Button onPress required

tbergq:


app/core/src/components/authentication/EmailLoginForm.js
```diff
@@ -69,14 +69,21 @@ export default class EmailLoginForm extends React.Component<Props, State> {
...
+
+      <TextButton
+        onPress={this.handleFormSubmit}
+        disabled={this.state.loading}
+        title={
+          <Translation
+            id={
+              this.state.loading
+                ? 'core.authentication.logging_in'
+                : 'core.authentication.login' // TODO: Does not seem to toggle when loading changes
```
mrtnzlml: I noticed - something is broken. It's not important now but do you have any idea what's wrong?

PR #709, opened 2018-06-05
Fixing build.gradle for Linux to run android app

RobinCsl:
This is probably more informative than really to be merged. It allowed me to run `yarn android` without error.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -97,7 +97,16 @@ def getPassword(String currentUser, String keyChain) {
    def stdout = new ByteArrayOutputStream()
    def stderr = new ByteArrayOutputStream()
    exec {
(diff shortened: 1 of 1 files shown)


mrtnzlml:
@RobinCsl So we should not merge it right? 🙂 I think it's going to be really easier to say _"hey, we support only macOS at this moment"_ because we require every developer to be able to tweak both iOS and Android versions.

tbergq:
We could maybe add a note to the redame section. 

‘if you for some reason want to run this on Linux, the do ..., however we just support Mac at this point’

mrtnzlml:
In any case, can I close this PR? We can link it somewhere if necessary... 🙂

tbergq:
Yes, let's close this.

PR #712, opened 2018-06-06
Shared: split 'AdaptableLayout' into multiple files

mrtnzlml:


tbergq:
Should we maybe to this with other Contexts as well?

mrtnzlml:
Not necessary. I would do it only when needed. This component was pretty big already so I wanted to split it.

tbergq:
Makes sense 🙂

PR #713, opened 2018-06-06
MMB: refactor trip overview layout

mrtnzlml:
The whole header must be clickable on mobile but there must be just one
button on tablet. I also connected the trip overview to the GraphQL.
Note: no design yet.


app/MMB/src/components/header/Header.js
```diff
@@ -29,11 +33,75 @@ const Header = (props: Props) => {
...
       </View>
     </React.Fragment>
   );
-};
+}
+
+class Header extends React.Component<
+  {|
+    +data: HeaderType,
+    +openSubmenu: (activeId: string, menuId: string) => void,
```
tbergq: Can you define a type Props instead?

app/MMB/src/scenes/tripOverview/TripOverview.js
```diff
@@ -0,0 +1,24 @@
+// @flow strict
+
+import * as React from 'react';
+import { graphql, createFragmentContainer } from '@kiwicom/mobile-relay';
+import { Translation } from '@kiwicom/mobile-localization';
+
+import type { TripOverview as TripOverviewType } from './__generated__/TripOverview.graphql';
+
+type Props = {|
+  data: TripOverviewType,
```
tbergq: Could be read only type

app/MMB/src/scenes/tripOverview/TripOverviewTablet.js
```diff
@@ -0,0 +1,31 @@
+// @flow strict
+
+import * as React from 'react';
+import { graphql, PrivateApiRenderer } from '@kiwicom/mobile-relay';
+
+import TripOverview from './TripOverview';
+import BookingDetailContext from '../../context/BookingDetailContext';
+
+function TripOverviewTablet(rendererProps) {
```
tbergq: missing flow types

mrtnzlml: Seems like Flow is clever enough to understand what's going on here (it would complain otherwise). Fixed... 🙂

PR #715, opened 2018-06-06
MMB: add flight overview timeline

mrtnzlml:
Just a simple skeleton so I can expand on this idea.

### DIFF
--- a/app/MMB/src/scenes/tripOverview/Timeline.js
+++ b/app/MMB/src/scenes/tripOverview/Timeline.js
@@ -1,0 +1,66 @@
+// @flow strict
+
+import * as React from 'react';
+import { View, FlatList } from 'react-native';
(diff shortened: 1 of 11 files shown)


app/MMB/src/scenes/tripOverview/Timeline.js
```diff
@@ -0,0 +1,63 @@
...
+
+    if (!isLast) {
+      lastItemStyle.borderColor = Color.ink.light;
+      lastItemStyle.borderStartWidth = lineWidth;
+    }
+
+    return (
+      <View style={styles.rowContainer}>
+        <View style={[styles.item, lastItemStyle]}>
+          <View style={{ marginBottom: 20 }}>{item}</View>
```
tbergq: Is it necessary to use inline style here?

mrtnzlml: Moved... 🙄

PR #716, opened 2018-06-06
Fix react native linking

lukewalczak:


android/app/build.gradle
```diff
@@ -169,6 +169,7 @@ android {
 }
 
 dependencies {
+    compile project(':react-native-share')
```
tbergq: can you change to implementation, compile is deprecated

package.json
```diff
@@ -24,7 +24,8 @@
   },
   "dependencies": {
     "react": "^16.3.2",
-    "react-native": "^0.55.4"
+    "react-native": "^0.55.4",
+    "react-native-share": "^1.0.27"
```
mrtnzlml: @tbergq Wrong file but not sure where you need it exactly. Shared or MMB?

android/app/src/main/java/com/reactnativeapp/MainApplication.java
```diff
@@ -34,6 +35,7 @@ public boolean getUseDeveloperSupport() {
         protected List<ReactPackage> getPackages() {
             return Arrays.asList(
                     new MainReactPackage(),
+            new RNSharePackage(),
```
tbergq: Please indent this to align with the rest, `OCD` tick

package.json
tbergq: Right, for now it will be in MMB

tbergq: To be specific, this should be added to `app/MMB/package.json`

tbergq:
LGDM, thanks 😄

PR #719, opened 2018-06-07
MMB: Share e-ticket-pdf

tbergq:


android/app/src/main/AndroidManifest.xml
```diff
@@ -1,38 +1,25 @@
...
-      android:name=".MainApplication"
-      android:allowBackup="true"
-      android:label="@string/app_name"
-      android:icon="@mipmap/ic_launcher"
-      android:theme="@style/AppTheme"
-      tools:replace="android:allowBackup">
+    <application android:name=".MainApplication" android:allowBackup="true" android:label="@string/app_name" android:icon="@mipmap/ic_launcher" android:theme="@style/AppTheme" tools:replace="android:allowBackup">
+        <provider android:name="android.support.v4.content.FileProvider" android:authorities="${applicationId}.provider" android:grantUriPermissions="true" android:exported="false">
+            <meta-data android:name="android.support.FILE_PROVIDER_PATHS" android:resource="@xml/filepaths" />
+        </provider>
```
tbergq: The provider tag is the only addition, the rest is reformatting document

app/MMB/src/components/Invoice.js
```diff
@@ -26,7 +26,8 @@ export default class Invoice extends React.Component<{||}> {
       <PdfViewer
         source={{
           uri,
-          cache: true, // no expiration - store as long as possible
+          cache: true,
+          expiration: Infinity, // no expiration - store as long as possible
         }}
```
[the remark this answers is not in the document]
mrtnzlml: Awesome. So intuitive. Could you please add these two options into `PdfViewer` directly? Let's assume for now that every PDF is going to be cached forever by default. Does it make sense? I don't want to repeat this everywhere because this is not the only place where we use it...

app/MMB/src/context/BookingDetailContext.js
```diff
@@ -16,8 +16,10 @@ const defaultState = {
   arrivalCityId: '',
   arrivalTime: new Date(),
   departureTime: new Date(),
+  eTicketPath: null,
   actions: {
     setBookingDetail: () => {},
+    setETicketPath: () => {},
```
mrtnzlml: Do we need to expand context with all these things? Do not create Redux mess from it, please. This doesn't sound like a context thing.

tbergq: The component tree will be like:

     ETicketPdfScreen
   |                   |
navigationHeader .     EticketPdf
   |                   |
ShareButton            PdfViewer

The only way to share this url from the Eticket-pdf-viewer with the share button in the navigation header is through a Context. I could of course create a separate context for this if you think that is better?

mrtnzlml: OK, leave it. It's just strange that we are not in charge of the path now. I would expect we know in advance where are the PDFs going to be stored (?).

tbergq: We could do that, but now it is stored cache (tmp folder ish), we could downlaod it ourselves and probably control the path

mrtnzlml: Only if you are super motivated to make your best here... 🙂 We can leave it like this for now.

tbergq:
@mrtnzlml please check again

I added default props to pdfview component, so you can still override cache and expiration if you want to. default is true and Infinity

app/MMB/src/context/BookingDetailContext.js
tbergq: > We can leave it like this for now

This usually means forever 🙈 

I will have a look at it just after this is merged 👍

mrtnzlml: > This usually means forever 🙈

Damn, this is so true... 😄

OK, thank you very much.

PR #721, opened 2018-06-07
MMB: remove header image

mrtnzlml:


mrtnzlml:
> I wish they could have shared with us that this was to be removed 🙁

🙁

PR #723, opened 2018-06-07
MMB: add basic design for the trip overview timeline

mrtnzlml:


app/MMB/src/scenes/tripOverview/TimelineTitle.js
```diff
@@ -0,0 +1,80 @@
...
+function TimelineTitle(props: Props) {
+  const data = props.data;
+  const localTime = idx(data, _ => _.localTime);
+  const cityName = idx(data, _ => _.airport.city.name);
+  const iataCode = idx(data, _ => _.airport.locationId);
+
+  return (
+    <View style={styleSheet.row}>
+      <View style={styleSheet.dateTime}>
+        {localTime && (
```
tbergq: This is not necessarily `android-safe` should localTime ever be 0, or anything that would be interpreted as text && false, it could crash android.

mrtnzlml: Can you please elaborate more on the Android safeness here? Isn't output of `idx` call always `null`, `undefined` or the expected value (string date in this case)?

How to write it better?

tbergq: If localTime would return '' (not sure if it is a possible value from graphql) it results in crash on android. 

having a fallback value idx(...) || '', then {localTime !== '' && ...
or {Boolean(localTime) && ...

would prevent this

mrtnzlml: > not sure if it is a possible value from graphql

I hope not. The output type is `DateTime` and it should fail otherwise. Anyway - fixed... :)

PR #725, opened 2018-06-07
MMB: Download and store e-ticket pdfs

tbergq:


ios/reactNativeApp/Info.plist
```diff
@@ -1,81 +1,81 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
+<dict>
```
tbergq: I didn't do anything with this file, seems like Xcode insisted on reformatting it 🤷‍♂️

packages/pdf/index.js
```diff
@@ -2,4 +2,6 @@
 
 import PdfViewer from './src/PdfViewer';
 
+export { default as PdfDownloader } from './src/PdfDownloader';
```
tbergq: Open to suggestions for better naming 🤔

app/MMB/src/scenes/tickets/eTicketPdf/ETicketPdf.js
```diff
@@ -3,33 +3,28 @@
...
-          errorMessage={<Translation id="mmb.tickets.not_available" />}
-        />
-      );
-    }
-
-    return <PdfViewer uri={ticketUrl} onLoad={this.onPdfLoaded} />;
-  };
-}
+  return (
+    <PdfDownloader fileName={`/eTickets/${bookingId}.pdf`} url={ticketUrl} />
```
mrtnzlml: I would expect here only the `PdfViewer` here because this is what we want. This component should care of downloading these PDFs. Also, I am not sure about the filename change. Shouldn't we keep the original PDF name?

packages/pdf/src/PdfDownloader.js
```diff
@@ -0,0 +1,46 @@
...
+  };
+
+  render = () => {
+    const { url } = this.state;
+
+    if (url === null) {
+      return <FullPageLoading />;
+    }
+
+    return <PdfViewer uri={url} cache={false} />;
```
mrtnzlml: How does it work exactly? Will it load the file from the filesystem?

app/MMB/src/scenes/tickets/eTicketPdf/ETicketPdf.js
tbergq: My idea was to do the logic about showing loader while downloading, keeping the url contained in one file, instead of having to do this

```

tbergq: Also about keeping the original name, I think it makes sense to store it by `/tickets/<bid>.pdf` how else would we easily find the file, if we not want to put it in a context?

The existing name seems kind of random

packages/pdf/src/PdfDownloader.js
tbergq: Pdf viewer is able to load from assets folder, http(s) adresses or local file system (maybe even base64 string) depends on what you pass it

app/MMB/src/scenes/tickets/eTicketPdf/ETicketPdf.js
mrtnzlml: I will remind you this invention after we'll start generating multiple e-tickets for one BID... :)

tbergq: Are there multiple e-tickes for one BID?

mrtnzlml: No, not yet.

tbergq: Will it ever be? Will it ever make sense?

If you have any better means of storing it, which makes it easy to find, I am down for implementing it

mrtnzlml: I don't know. Think about the future. It may happen. Can we fetch the names from the API where needed?

tbergq: We would need to parse the url, get the name, we could probably do this on graphql

mrtnzlml: OK, forget it for now.

PR #726, opened 2018-06-07
First draft for timeline; accurate styles yet to be applied

RobinCsl:
This is what it looks like on Android.

screenshot_1528381034

### DIFF
--- a/app/MMB/src/scenes/timeline/DaySeparator.js
+++ b/app/MMB/src/scenes/timeline/DaySeparator.js
@@ -1,0 +1,52 @@
+// @flow
+
+import * as React from 'react';
+import { StyleSheet, Text, Color } from '@kiwicom/mobile-shared';
+import {
+  Translation,
+  DateFormatter,
+  DateUtils,
+} from '@kiwicom/mobile-localization';
+
+import TimelineRow from './TimelineRow';
+
+type Props = {|
+  +date: Date,
+|};
+
+const DaySeparator = (props: Props) => {
+  const { date } = props;
+  let translatedDay = null;
+  if (DateUtils.diffInDays(date, new Date()) === 0) {
+    translatedDay = (
+      <Translation id="mmb.booking_timeline.today" textTransform="uppercase" />
+    );
+  } else {
+    translatedDay = (
+      <Translation
+        passThrough={DateFormatter(new Date(date)).formatToDate()}
+        textTransform="uppercase"
+      />
+    );
+  }
+  return (
+    <TimelineRow
+      rowStyle={styles.container}
+      rightColumn={<Text style={styles.date}>{translatedDay}</Text>}
+    />
+  );
+};
+
+const styles = StyleSheet.create({
+  container: {
+    paddingTop: 10,
+    paddingBottom: 10,
+    backgroundColor: Color.grey.$300,
+  },
+  date: {
+    color: Color.ink.light,
+    fontWeight: 'bold',
+  },
+});
+
+export default DaySeparator;
--- a/app/MMB/src/scenes/timeline/Timeline.js
+++ b/app/MMB/src/scenes/timeline/Timeline.js
@@ -1,0 +1,80 @@
+// @flow
+
+import * as React from 'react';
+import { ScrollView } from 'react-native';
+import idx from 'idx';
+import { PrivateApiRenderer, graphql } from '@kiwicom/mobile-relay';
+
+import type {
+  TimelineQueryResponse,
+  TimelineEvent as TimelineEventType,
+} from './__generated__/TimelineQuery.graphql';
+import BookingDetailContext from '../../context/BookingDetailContext';
+import BookedFlightTimelineEvent from './events/BookedFlightTimelineEvent';
+import AirportArrivalTimelineEvent from './events/AirportArrivalTimelineEvent';
+import DaySeparator from './DaySeparator';
+
+// @TODO Display day banners only once for each day
+function getValidTimelineEvent(data: TimelineEventType, index) {
+  if (data && data.__typename) {
+    switch (data.__typename) {
+      case 'BookedFlightTimelineEvent':
+        return (
+          <React.Fragment key={'TimelineEvent-' + data.__typename + index}>
+            {data && data.timestamp && <DaySeparator date={data.timestamp} />}
+            <BookedFlightTimelineEvent data={data} />
+          </React.Fragment>
+        );
+      case 'AirportArrivalTimelineEvent':
+        return (
+          <React.Fragment key={'TimelineEvent-' + data.__typename + index}>
+            {data && data.timestamp && <DaySeparator date={data.timestamp} />}
+            <AirportArrivalTimelineEvent data={data} />
+          </React.Fragment>
+        );
+      default:
+        return null;
+    }
+  }
+  return null;
+}
+
+export default class Timeline extends React.Component<{}> {
+  renderInner = (renderProps: TimelineQueryResponse) => {
+    const events = idx(renderProps, _ => _.bookingTimeline.events);
+    const children =
+      events &&
+      events.map((event, index) => getValidTimelineEvent(event, index));
+
+    return <ScrollView>{children}</ScrollView>;
+  };
+
+  render = () => (
+    <BookingDetailContext.Consumer>
+      {({ bookingId }) => (
+        <PrivateApiRenderer
+          query={graphql`
+            query TimelineQuery($id: ID!) {
+              bookingTimeline(id: $id) {
+                events {
+                  __typename
+                  timestamp
+                  ... on BookedFlightTimelineEvent {
+                    ...BookedFlightTimelineEvent
+                  }
+                  ... on AirportArrivalTimelineEvent {
+                    ...AirportArrivalTimelineEvent
+                  }
+                }
+              }
+            }
+          `}
+          variables={{
+            id: bookingId,
+          }}
+          render={this.renderInner}
+        />
+      )}
+    </BookingDetailContext.Consumer>
+  );
+}
--- a/app/MMB/src/scenes/timeline/TimelineEvent.js
+++ b/app/MMB/src/scenes/timeline/TimelineEvent.js
@@ -1,0 +1,53 @@
+// @flow
+
+import * as React from 'react';
+import { View } from 'react-native';
+import { StyleSheet, Color } from '@kiwicom/mobile-shared';
+
+import TimelineRow from './TimelineRow';
+
+type Props = {|
+  +dateLocation: ?React.Node,
+  +iconVertLines: React.Node,
+  +mainContent: React.Node,
+  +isPastEvent?: boolean,
+|};
+
+export default function TimelineEvent(props: Props) {
+  let rowStyle = {};
+  if (props.isPastEvent) {
+    rowStyle = styles.isPastEvent;
+  }
+  return (
+    <TimelineRow
+      leftColumn={
+        <View style={styles.row}>
+          <View style={styles.dateLocation}>{props.dateLocation}</View>
+          <View style={styles.iconVertLines}>{props.iconVertLines}</View>
+        </View>
+      }
+      rightColumn={<View style={styles.mainContent}>{props.mainContent}</View>}
+      rowStyle={rowStyle}
+    />
+  );
+}
+
+const styles = StyleSheet.create({
+  row: {
+    flexDirection: 'row',
+  },
+  isPastEvent: {
+    backgroundColor: Color.grey.$50,
+  },
+  dateLocation: {
+    flex: 1,
+  },
+  iconVertLines: {
+    flex: 1,
+  },
+  mainContent: {
+    flex: 1,
+    justifyContent: 'center',
+    alignItems: 'flex-start',
+  },
+});
--- a/app/MMB/src/scenes/timeline/TimelineEventDateLocation.js
+++ b/app/MMB/src/scenes/timeline/TimelineEventDateLocation.js
@@ -1,0 +1,29 @@
+// @flow
+
+import * as React from 'react';
+import { View } from 'react-native';
+import { StyleSheet, Text } from '@kiwicom/mobile-shared';
+import { Translation } from '@kiwicom/mobile-localization';
+
+type Props = {|
+  +time: string,
+|};
+
+export default function TimelineEventDateLocation(props: Props) {
+  return (
+    <View style={styles.container}>
+      <Text style={styles.text}>
+        <Translation passThrough={props.time} />
+      </Text>
+    </View>
+  );
+}
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    justifyContent: 'center',
+    alignItems: 'flex-end',
+  },
+  text: { fontSize: 10 },
+});
--- a/app/MMB/src/scenes/timeline/TimelineRow.js
+++ b/app/MMB/src/scenes/timeline/TimelineRow.js
@@ -1,0 +1,36 @@
+// @flow
+
+import * as React from 'react';
+import { View } from 'react-native';
+import { StyleSheet, Color, type StylePropType } from '@kiwicom/mobile-shared';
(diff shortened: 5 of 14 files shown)


app/MMB/src/scenes/timeline/Events.js
```diff
@@ -0,0 +1,39 @@
...
+
+import * as React from 'react';
+import { DateFormatter, DateUtils } from '@kiwicom/mobile-localization';
+import { View } from 'react-native';
+
+import Event, { type Props as EventType } from './event/Event';
+import Day from './event/Day';
+
+type Props = {|
+  events: Array<EventType>,
```
tbergq: Our prop type should be readonly `+events:`

app/MMB/src/scenes/timeline/Events.js
```diff
@@ -0,0 +1,39 @@
...
+          <View key={event.description + index}>
+            {event.first && <Day label={dayLabel} />}
+            <Event
+              icon={event.icon}
+              description={event.description}
+              comment={event.comment}
+              dateTime={event.dateTime}
+              location={event.location}
+              first={event.first}
+            />
```
tbergq: Maybe you could pass the event object instead.
`<Event event={event} /> `

app/MMB/src/scenes/timeline/Timeline.js
```diff
@@ -0,0 +1,65 @@
...
+import { DateUtils } from '@kiwicom/mobile-localization';
+
+import Events from './Events';
+
+type Props = {|
+  +navigation: NavigationType,
+|};
+
+export default class Timeline extends React.Component<Props> {
+  navigate = (key: RouteNamesType) => {
```
tbergq: Is this function needed in this class?

app/MMB/src/scenes/timeline/Timeline.js
```diff
@@ -0,0 +1,65 @@
...
+      },
+    ];
+
+    const sortedEvents = events.sort((a, b) => a.dateTime - b.dateTime);
+
+    const enrichedEvents = sortedEvents.map((event, index) => ({
+      ...event,
+      first:
+        index === 0 ||
+        DateUtils.diffInDays(events[index - 1].dateTime, event.dateTime) !== 0,
```
tbergq: Why do we need to add this first bool here?

app/MMB/src/scenes/timeline/Events.js
mrtnzlml: It's odd to import `Prop` type from the component just to prop-drill the props. We should compose the components instead. Why not to expect `Event` components here directly?

app/MMB/src/scenes/timeline/Events.js
mrtnzlml: > Maybe you could pass the event object instead.

It does not sound like a good idea.

app/MMB/src/scenes/timeline/Timeline.js
```diff
@@ -0,0 +1,65 @@
...
+export default class Timeline extends React.Component<Props> {
+  navigate = (key: RouteNamesType) => {
+    this.props.navigation.navigate({
+      routeName: key,
+      key: `key-${key}`,
+    });
+  };
+
+  render = () => {
+    const events = [
```
mrtnzlml: I think this is a bad way to build applications. Shouldn't you create the API first? You made up this data structure but you are already throwing your work away because this is not going to be the same in API...

app/MMB/src/scenes/timeline/Timeline.js
mrtnzlml: And again - this is completely pointless. Create API first and your application will be clear. No sorting here, no magic boolean constants in your client...

app/MMB/src/scenes/timeline/event/Day.js
```diff
@@ -0,0 +1,38 @@
...
+    borderBottomColor: Color.grey.$400,
+  },
+  day: {
+    color: Color.ink.light,
+    fontSize: 14,
+  },
+});
+
+type Props = {|
+  label: string,
```
mrtnzlml: No strings - translation types.

app/MMB/src/scenes/timeline/event/DescriptionComment.js
```diff
@@ -0,0 +1,45 @@
+// @flow
+
+import * as React from 'react';
+import { View } from 'react-native';
+import { Translation } from '@kiwicom/mobile-localization';
+import { Text, StyleSheet, Color } from '@kiwicom/mobile-shared';
+
+export type DescriptionCommentType = {|
+  description: string,
+  comment: string,
```
mrtnzlml: Please always accept translation types instead of strings. Only this way we can be sure everything is translated properly.

app/MMB/src/scenes/timeline/event/Event.js
```diff
@@ -0,0 +1,43 @@
...
+import DescriptionComment, {
+  type DescriptionCommentType,
+} from './DescriptionComment';
+
+export type Props = {|
+  ...IconVerticalLineType,
+  ...TimeLocationType,
+  ...DescriptionCommentType,
+  first: boolean,
+|};
```
mrtnzlml: This is a really ugly type. Do not reuse `Props` please. It does not make any sense. One way how to make it better is to compose the components instead.

app/MMB/src/scenes/timeline/event/Event.js
```diff
@@ -0,0 +1,43 @@
...
+    paddingEnd: 5,
+  },
+});
+
+const TOP_OFFSET_FIRST = 30;
+const TOP_OFFSET_NORMAL = 5;
+
+export default function Event(props: Props) {
+  const { dateTime, location, icon, description, comment, first } = props;
+  const topOffset = first ? TOP_OFFSET_FIRST : TOP_OFFSET_NORMAL;
```
mrtnzlml: This is one of the props that are most probably very wrong. Did you see the video yesterday? 🙂 Are we going to create `last` because last is going to be different? And `odd`, `even`, `special`, ...

It does not sound right.

app/MMB/src/scenes/timeline/event/IconVerticalLine.js
```diff
@@ -0,0 +1,74 @@
+// @flow
+
+import * as React from 'react';
+import { View } from 'react-native';
+import { Icon, StyleSheet, Color } from '@kiwicom/mobile-shared';
+
+type IconType = {|
+  name: string,
+  type: 'classic' | 'special' | 'ongoing',
+|};
```
mrtnzlml: Just accept `Icon` component here like we do everywhere.

app/MMB/src/scenes/timeline/event/IconVerticalLine.js
```diff
@@ -0,0 +1,74 @@
...
+    alignItems: 'center',
+    width: 30,
+    marginStart: 5,
+    marginEnd: 10,
+  },
+  topLine: {
+    width: 1,
+    backgroundColor: 'grey',
+  },
+  bottomLine: { width: 1, flex: 1, backgroundColor: 'grey' },
```
mrtnzlml: Please, use colors from the shared package. These borders and backgrounds (or brand colors) are changing from time to time.

app/MMB/src/scenes/timeline/event/IconVerticalLine.js
```diff
@@ -0,0 +1,74 @@
...
+    <View style={styles.iconVerticalLine}>
+      <View style={[styles.topLine, { height: topOffset }]} />
+      <Icon
+        name={icon.name}
+        size={18}
+        style={[
+          styles.icon,
+          icon.type === 'classic' && styles.classic,
+          icon.type === 'special' && styles.special,
+          icon.type === 'ongoing' && styles.ongoing,
```
mrtnzlml: See? This does not scale well. Just accept icon here and be happy... 🙂

app/MMB/src/scenes/timeline/DaySeparator.js
```diff
@@ -0,0 +1,54 @@
...
+  container: {
+    paddingTop: 10,
+    paddingBottom: 10,
+    backgroundColor: Color.grey.$300,
+  },
+  date: {
+    color: Color.ink.light,
+    fontFamily: 'CircularPro-Book',
+    fontSize: 14,
+    fontWeight: 'bold',
```
mrtnzlml: You should not change the `fontFamily` or `fontSize` manually here. These are default values so just omit it, please...

app/MMB/src/scenes/timeline/Timeline.js
```diff
@@ -0,0 +1,84 @@
...
+// @TODO Display day banners only once for each day
+function getValidTimelineEvent(data, index) {
+  if (data && data.__typename) {
+    switch (data.__typename) {
+      case 'BookedFlightTimelineEvent':
+        return (
+          <React.Fragment>
+            {data && data.timestamp && <DaySeparator date={data.timestamp} />}
+            <BookedFlightTimelineEvent
+              key={data.__typename + index}
```
mrtnzlml: It would be probably better to return `id` from the API.

app/MMB/src/scenes/timeline/TimelineEventDateLocation.js
```diff
@@ -0,0 +1,29 @@
...
+  );
+}
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    justifyContent: 'center',
+    alignItems: 'flex-end',
+  },
+  text: { fontFamily: 'CircularPro-Book', fontSize: 10 },
```
mrtnzlml: Please, do not copy-paste `fontFamily` everywhere.

app/MMB/src/scenes/timeline/events/AirportArrivalTimelineEvent.js
```diff
@@ -0,0 +1,49 @@
...
+import { Text, StyleSheet } from '@kiwicom/mobile-shared';
+import { Translation, DateFormatter } from '@kiwicom/mobile-localization';
+import idx from 'idx';
+
+import TimelineEvent from '../TimelineEvent';
+import Icon from '../TimelineEventIcon';
+import DateLocation from '../TimelineEventDateLocation';
+
+type Props = {|
+  +data: ?Object,
```
mrtnzlml: This doesn't look right - you should return Relay fragment container from this component and fetch what's necessary here (and use generated Flow type for this component)...

app/MMB/src/scenes/timeline/events/AirportArrivalTimelineEvent.js
```diff
@@ -0,0 +1,49 @@
...
+            }}
+          />
+        </Text>
+      }
+    />
+  );
+};
+
+const styles = StyleSheet.create({
+  mainContent: { fontFamily: 'CircularPro-Book', fontSize: 14 },
```
mrtnzlml: These are the default values. Please remove it.

app/MMB/src/scenes/timeline/events/BookedFlightTimelineEvent.js
```diff
@@ -0,0 +1,53 @@
...
+import idx from 'idx';
+import { Text, StyleSheet } from '@kiwicom/mobile-shared';
+import { Translation, DateFormatter } from '@kiwicom/mobile-localization';
+
+import TimelineEvent from '../TimelineEvent';
+import Icon from '../TimelineEventIcon';
+import DateLocation from '../TimelineEventDateLocation';
+
+type Props = {|
+  +data: ?Object,
```
mrtnzlml: The same - fetch what you need to render this component here...

app/MMB/src/scenes/timeline/events/BookedFlightTimelineEvent.js
```diff
@@ -0,0 +1,53 @@
...
+    time = DateFormatter(new Date(timestamp)).formatToTime();
+  }
+  let dateLocation = null;
+  if (time) {
+    dateLocation = <DateLocation time={time} />;
+  }
+
+  const isPastEvent = true;
+  return (
+    <React.Fragment>
```
mrtnzlml: Seems like this `React.Fragment` is not necessary here (?).

app/MMB/src/scenes/timeline/Timeline.js
mrtnzlml: You could create one there... 🙂

app/MMB/src/scenes/timeline/TimelineEvent.js
```diff
@@ -0,0 +1,53 @@
...
+import * as React from 'react';
+import { View } from 'react-native';
+import { StyleSheet, Color, Text } from '@kiwicom/mobile-shared';
+
+import TimelineRow from './TimelineRow';
+
+type Props = {|
+  +dateLocation: ?React.Element<typeof View>,
+  +icon: React.Element<typeof View>,
+  +mainContent: React.Element<typeof View | typeof Text>,
```
mrtnzlml: Technically this code is correct. But I believe it's better not to create these restrictions. Just `React.Node` is enough - do not overcomplicate it... :)

It's clever to do it for translations for example where you want to be sure not to forget. But in this case, you can really accept "any" element. Maybe the `icon` should be type of `Icon` (?). But `React.Node` should be OK for most of the cases.

app/MMB/src/scenes/timeline/TimelineRow.js
```diff
@@ -0,0 +1,36 @@
+// @flow
+
+import * as React from 'react';
+import { View } from 'react-native';
+import { StyleSheet, Color } from '@kiwicom/mobile-shared';
+
+type Props = {|
+  +leftColumn?: React.Element<typeof View>,
+  +rightColumn?: React.Element<typeof View>,
+  +rowStyle?: Object,
```
mrtnzlml: I think there is a type for it in the shared package.

app/MMB/src/scenes/timeline/events/AirportArrivalTimelineEvent.js
```diff
@@ -0,0 +1,62 @@
...
+  );
+};
+
+// export default AirportArrivalTimelineEvent;
+
+export default createFragmentContainer(
+  AirportArrivalTimelineEvent,
+  graphql`
+    fragment AirportArrivalTimelineEvent on AirportArrivalTimelineEvent {
+      __typename
```
mrtnzlml: The `__typename` is probably not needed here since you already know what's the type.

app/MMB/src/scenes/timeline/TimelineEvent.js
RobinCsl: I understand, this is indeed much simpler :)
I renamed the `icon` props as it is slightly misleading: it is not just an icon, it is an icon surrounded by vertical lines.

app/MMB/src/scenes/timeline/TimelineRow.js
RobinCsl: I had to modify the eslint config to support `import` and `import type` for the same file.

app/MMB/src/scenes/timeline/events/AirportArrivalTimelineEvent.js
RobinCsl: Removed it here and also for the other event file `BookedFlightTimelineEvent.js`

app/MMB/src/scenes/timeline/TimelineRow.js
mrtnzlml: What? Why? We do it everywhere without eslint changes:

```js
import { Component, type ComponentType } from '../..'
```

RobinCsl: Oh, right. I'll try again like this then. I think I had tried that though.

RobinCsl: Well I probably did not try that hard enough... It works. I will revert `eslintrc.js`.

mrtnzlml:
Integrated into master, thanks.

PR #728, opened 2018-06-08
Hotels: Fix go back crash

tbergq:


app/hotels/src/navigation/singleHotel/SingleHotelNavigationScreen.js
```diff
@@ -17,11 +18,11 @@ type Props = {
...
 };
 
+const noop = () => {};
+
 class SingleHotelNavigationScreen extends React.Component<Props> {
   static navigationOptions = (props: Props) => {
-    function goBack() {
-      props.navigation.state.params.goBack();
-    }
+    const goBack = idx(props.navigation.state.params, _ => _.goBack) || noop;
```
mrtnzlml: When this happens actually? Also, is it possible to test it, please?

tbergq: It is one of these things that apparently does not happen in our react-native-repo, but happens when they integrate it to the native app. 

I don't really know how we could be able to test this.

My best guess is when it runs in native code, that it tries to evaluate the params.goBack before the function is called, while in our dev repo it does not evaluate it before the function is called.

mrtnzlml: WTF? 😄

I mean 😢

tbergq: 🤷‍♂️ 

Because on first render (happens before componentDidMount), params.goBack will be undefined. Then after componentDidMount, it will be defined. I think no one is fast enough to click the back button between first render and componentDidMount is being called

PR #733, opened 2018-06-11
CODEOWNERS: add @ferrannp as a code owner of Android and iOS code

mrtnzlml:
@ferrannp Are you OK with this?


ferrannp:
Can you add also `RNAndroidPlayground/` directory?

mrtnzlml:
@ferrannp Where is this directory? I cannot find it. Do you plan to create a new directory in the root folder?

mrtnzlml:
Merging now so you can work freely but please let me know about the `RNAndroidPlayground` so I can update it...

PR #736, opened 2018-06-12
Hotels: reuse double-column layout in the hotels package

mrtnzlml:


packages/shared/src/view/Layout.js
```diff
@@ -1,48 +1,22 @@
...
-};
-
-/**
- * This is base Layout component with default width. It can expand up to
- * portrait mode width. This means that the content is centered in
- * landscape mode.
- */
-export default function Layout({ children }: Props) {
-  const defaultStyle = createStyles();
+export default function Layout(props: Props) {
```
RobinCsl: Is there a reason not to use destructuring here?

mrtnzlml: Is there a reason to use destructing here? :)

RobinCsl: Fair enough. I just prefer destructuring :smile:

mrtnzlml:
> When in landscape mode on Nexus 5X, I can see the double column layout. However, there is no map being displayed. I don't know if this has been implemented yet, though. Otherwise, looks good to me.

Mobile landscapes are irrelevant for us (not officially supported). You probably do not have Google Maps API key in your repo but you should see the map on the 🍎 tablet...

packages/shared/src/view/Layout.js
mrtnzlml: I see - that's very subjective and it should not be part of the code review... :)

RobinCsl: You modified it because you preferred not to destructure. So I am just asking why you make the change as it seems to be arbitrary.

mrtnzlml: Aha, I actually created this component from scratch but GitHub thinks I just modified it because I used the same filename with different component (I actually renamed Layout to LayoutSingleColumn...). That's why.

PR #745, opened 2018-06-18
Relay: implement observable network with offline cache

mrtnzlml:


packages/relay/src/Environment.js
```diff
@@ -19,26 +60,52 @@ export default function createEnvironment(accessToken: string = '') {
...
+            store.publish(new RecordSource(JSON.parse(content)));
+            const operationSelector = createOperationSelector(
+              operation,
+              variables,
+            );
+            if (store.check(operationSelector.root)) {
+              // we have all data in the store to fulfill this query so let's
+              // load it from the memory first and call the API after that
+              // this will make the UI feel really fast
+              observer.next(store.lookup(operationSelector.root));
```
RobinCsl: So, here you don't use `observer.complete()` because you will still use the observer in the `finally` clause. Correct?

packages/relay/src/Environment.js
```diff
@@ -1,13 +1,54 @@
...
+      query: operation.text, // TODO: fetch persisted queries instead (based on operation.id)
+      variables,
+    }),
+  })
+    .then(fetchResponse => fetchResponse.json())
+    .then(jsonResponse => {
+      if (jsonResponse.errors) {
+        jsonResponse.errors.forEach(error => console.warn(error));
+      }
+      observer.next(jsonResponse);
```
RobinCsl: If I understand correctly, this observer will send `jsonResponse` to the store and it will be stored there.

packages/relay/src/Environment.js
```diff
@@ -1,13 +1,54 @@
...
+      observer.complete();
+    })
+    .then(() => {
+      AsyncStorage.setItem(
+        ASYNC_STORE_KEY,
+        JSON.stringify(store.getSource()),
```
RobinCsl: And here, you're setting the key `ASYNC_STORE_KEY` with a whole dump of the store.

packages/relay/src/Environment.js
```diff
@@ -1,13 +1,54 @@
...
+  createOperationSelector,
+} from 'relay-runtime';
+import { AsyncStorage } from 'react-native';
 
 type FetcherResponse = {|
   +data: Object,
   +errors?: $ReadOnlyArray<Object>,
 |};
 
+const ASYNC_STORE_KEY = '@OfflineStore:key';
```
RobinCsl: Just out of curiosity: is there a reason to add `:key` to `ASYNC_STORE_KEY`?

packages/relay/index.js
```diff
@@ -53,7 +53,7 @@ export type RelayPaginationProp = {|
 |};
 
 type RefetchOptions = {|
-  +force?: boolean,
+  +force: true, // we always have to force the refetch since we use offline cache (TODO: wrap it)
```
RobinCsl: So because you use offline cache, you want to display first data from the cache (unless `cacheConfig.force` is true), and then fetch data from the network, and for it to update, this is necessary. Correct?

app/profile/index.js
```diff
@@ -14,6 +14,7 @@ export default class LoginScreen extends React.Component<{||}> {
   };
 
   onLogout = () => {
+    // TODO: clean the Relay offline cache (?)
```
RobinCsl: Would not this be: 
`AsyncStore.removeItem('@OfflineStore:key');`?

app/MMB/src/scenes/passenger/PassengerDetailContainer.js
```diff
@@ -11,7 +11,7 @@ type Props = {||};
 
 export default class PassengerDetailContainer extends React.Component<Props> {
   renderInner = (renderProps: PassengerDetailContainerQuery) => {
-    return <PassengerDetail booking={renderProps.booking} />;
+    return <PassengerDetail booking={renderProps.node} />;
```
RobinCsl: Was the query already updated before?

mrtnzlml: Yes, this was a mistake I just fixed it... 🙂

app/profile/index.js
mrtnzlml: Yes. But I don't want to implement it now since this part of the application is not ready yet. We will have to wrap the `AsyncStore` and clean the cache here. Let's ignore this `profile` part for now (it's only for us).

packages/relay/index.js
mrtnzlml: It's only because of UX let's say. Currently, we try to load data from cache first and after that load them from API and only update the components if necessary (new data).

But this looks weird when refreshing - the loader stops spinning very fast because we read it first from the cache (and that's fast). Therefore I wanted to really explicitly say "re-fetch" and this way we don't even touch the cache while refreshing (and the spinner is really snipping until we re-fetch it from the network). We should wrap it later so it's not necessary to write it everywhere...

packages/relay/src/Environment.js
mrtnzlml: No reason - could be whatever you want. I took it from [the docs](https://facebook.github.io/react-native/docs/asyncstorage.html).

packages/relay/src/Environment.js
mrtnzlml: No, this is like `return` from the fetch (it's more like a `yield` here). This way I can return multiple values (first from the cache and later from the network). Components will be re-rendered only if needed.

packages/relay/src/Environment.js
mrtnzlml: Yes, the whole `store` snapshot will be stored to the `AsyncStorage` so I can read it with the next request and use it immediately without waiting for the network. That's the biggest improvement and UI should feel more responsive now (not long loading spinners)...

packages/relay/src/Environment.js
mrtnzlml: Yes. The observer wouldn't accept `next` after calling `complete`. I will call `complete` only when I am sure I will not call `next` (after network response basically).

PR #746, opened 2018-06-18
Circle CI to build Android dependencies and rnkiwimobile

### LINKED ISSUE #740: More automation for Android integration
- [x] ~~Investigate if we can get rid of intermediate snapshots https://github.com/kiwicom/mobile/blob/master/RNAndroidPlayground/rnkiwimobile/build.gradle#L48-L50 and take them from `node_modules` now that is in the same directory~~ <-- we won't do it
- [x] Create a script to create SNAPSHOT(s) of `rnkiwimobile` and other dependencies
- [x] Remove versions from dependencies.gradle and rely on package.json
- [x] Make CircleCI to create SNAPSHOT(s) of `rnkiwimobile` and other dependencies

ferrannp:
Fixes #740.

- Generates dependencies SNAPSHOTS(s) taking version from package.json if they do not exist
- Generates JS and `rnkiwmobile` library taking version from package.json if it does note xist

### DIFF
--- a/scripts/buildAndroidSnapshots.js
+++ b/scripts/buildAndroidSnapshots.js
@@ -1,0 +1,84 @@
+/* @flow */
+
+/* eslint-disable no-console */
+
+const urlExists = require('url-exists');
+const child_process = require('child_process');
+
+const exec = (command, options) =>
+  child_process.execSync(command, {
+    stdio: 'inherit',
+    ...options,
+  });
+
+const DEFAULT_URL =
+  'http://trinerdis.cz:8000/repository/snapshots/com/trinerdis/skypicker/';
+
+if (!process.env.ANDROID_DEPLOYMENT_PASSWORD) {
+  throw Error('You need to pass ANDROID_DEPLOYMENT_PASSWORD.');
+}
+
+// Read native dependencies versions from package.json (maps, vector icons...)
+const buildPackage = require('../.build/package.json');
+
+// Get versions and remove ^ if necessary
+const getDependencyVersion = packageName => {
+  return buildPackage.dependencies[packageName].replace('^', '');
+};
+
+// Get rnkiwimobile version
+const RNKiwiMobileVersion = buildPackage.version;
+
+const deployToTrinerdis = (packageName, manualVersion, buildJS = false) => {
+  const version = manualVersion
+    ? manualVersion
+    : getDependencyVersion(packageName);
+
+  return new Promise((resolve, reject) =>
+    urlExists(
+      `${DEFAULT_URL}${packageName}/${version}-SNAPSHOT/`,
+      (err, exists) => {
+        if (err) {
+          reject(err);
+          return;
+        }
+        if (!exists) {
+          if (buildJS) {
+            console.log('Generating Android JS build...');
+            child_process.execSync('scripts/buildAndroidJavascript.sh', {
+              stdio: 'inherit',
+            });
+          }
+          console.log(`Deploying ${packageName}/${version}-SNAPSHOT`);
+          exec(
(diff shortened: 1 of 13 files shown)


scripts/buildAndroidSnapshots.js
```diff
@@ -0,0 +1,84 @@
...
+
+const DEFAULT_URL =
+  'http://trinerdis.cz:8000/repository/snapshots/com/trinerdis/skypicker/';
+
+if (!process.env.ANDROID_DEPLOYMENT_PASSWORD) {
+  throw Error('You need to pass ANDROID_DEPLOYMENT_PASSWORD.');
+}
+
+// Read native dependencies versions from package.json (maps, vector icons...)
+const buildPackage = require('../.build/package.json');
```
mrtnzlml: This file is going to be eventually removed. Is there a better solution? (maybe just the `packing.json` from the root)

scripts/buildAndroidSnapshots.js
```diff
@@ -0,0 +1,84 @@
...
+// Get rnkiwimobile version
+const RNKiwiMobileVersion = buildPackage.version;
+
+const deployToTrinerdis = (packageName, manualVersion, buildJS = false) => {
+  const version = manualVersion
+    ? manualVersion
+    : getDependencyVersion(packageName);
+
+  return new Promise((resolve, reject) =>
+    urlExists(
```
mrtnzlml: Just a question - why is this necessary?

scripts/buildAndroidSnapshots.js
```diff
@@ -0,0 +1,84 @@
...
+    ),
+  );
+};
+
+(() => {
+  console.log(`Start building Android SNAPSHOT(s)...`);
+  Promise.all([
+    deployToTrinerdis('react-native-maps'),
+    deployToTrinerdis('react-native-vector-icons'),
+    deployToTrinerdis('react-native-tooltips'),
```
mrtnzlml: Ah. Uh. So we didn't solve the pain with native deps... 😞

scripts/buildAndroidSnapshots.js
ferrannp: Mmmm, if we cannot use that file, we will need to check several ones because you have dependencies spread. Like `react-native` in the root but others are in shared. Want me to change that?

scripts/buildAndroidSnapshots.js
ferrannp: So I talk this with @Blackhex and we would collapse trinerdis if we deploy every time we merge so I am checking if that url exists and just deploying when we change versions.

scripts/buildAndroidSnapshots.js
ferrannp: Also talked with @Blackhex and we need those for Android as they don't really exist out there deployed. We need the SNAPSHOTS(s) in order to avoid build time. Updating a native dependency should not require any work, adding a new one yes.

scripts/buildAndroidSnapshots.js
mrtnzlml: We went from this:

1. add native dep
2. add native dep to the package.json

To this:

1. add native dep
2. add native dep to the package.json
3. add native dep to the buildAndroidSnapshots.js

Instead of this:

1. add native dep

That's what I don't like about it - it's error prone.

ferrannp: Mmm that is not entirely true because when you added a native dep, someone had to go here https://gitlab.skypicker.com/mobile/react-native-hotels and it + create the Snapshot process.

Now that we have it though all in one repo, maybe we can think of way to reduce steps.

PR #747, opened 2018-06-19
Adding LeaveForAirportTimelineEvent

RobinCsl:


app/MMB/src/scenes/timeline/Timeline.js
```diff
@@ -32,6 +33,13 @@ function getValidTimelineEvent(data: TimelineEventType, index) {
             <AirportArrivalTimelineEvent data={data} />
           </React.Fragment>
         );
+      case 'LeaveForAirportTimelineEvent':
+        return (
+          <React.Fragment key={'TimelineEvent-' + data.__typename + index}>
+            {data && data.timestamp && <DaySeparator date={data.timestamp} />}
+            <LeaveForAirportTimelineEvent data={data} />
+          </React.Fragment>
```
mrtnzlml: This is clearly quite repetitive. Let's stop here and try to think how to improve it, please...
 🙂

```js
// this is basically still the same
<React.Fragment key={'TimelineEvent-' + data.__typename + index}>
  {data && data.timestamp && <DaySeparator date={data.timestamp} />}
  <Component data={data} />
</React.Fragment>
```

Moreover, I don't think you should add the `DaySeparator` everywhere. This should be done automatically. So maybe it's still just WIP? (but even MVP can be DRY)

RobinCsl: Okay, I was planning to deal with the DaySeparator later in a different branch, but I guess I'll do it in this one so that next PR's are nicer.

mrtnzlml: No, it's fine - I am OK with this explanation. I just wanted to stop it but since you are going to fix it then we are good to go.

app/MMB/src/scenes/timeline/events/LeaveForAirportTimelineEvent.js
```diff
@@ -0,0 +1,61 @@
...
+import type { LeaveForAirportTimelineEvent as LeaveForAirportTimelineEventType } from './__generated__/LeaveForAirportTimelineEvent.graphql';
+import TimelineEvent from '../TimelineEvent';
+import TimelineEventIcon from '../TimelineEventIcon';
+import DateLocation from '../TimelineEventDateLocation';
+
+type Props = {|
+  +data: LeaveForAirportTimelineEventType,
+|};
+
+const LeaveForAirportTimelineEvent = (props: Props) => {
```
mrtnzlml: For the future reference: internals of this component looks also very similar to other events. Maybe you could reuse some parts?

app/MMB/src/scenes/timeline/events/LeaveForAirportTimelineEvent.js
```diff
@@ -0,0 +1,61 @@
...
+    <TimelineEvent
+      isPastEvent={isPastEvent}
+      dateLocation={dateLocation}
+      iconVertLines={
+        <TimelineEventIcon name="directions-walk" isPastEvent={isPastEvent} />
+      }
+      mainContent={
+        <TranslationFragment>
+          <Translation id="mmb.booking_timeline.event.leave_for_airport.title" />
+          <Text style={{ fontSize: 13, color: Color.ink.light }}>
```
mrtnzlml: Please try to avoid inline styles. This object is created every time and it may cause problems with `PureComponent` later (`Text` is one of the adepts for `PureComponent`)...

app/MMB/src/scenes/timeline/events/LeaveForAirportTimelineEvent.js
RobinCsl: Yes, you're right. I will abstract this a bit more. Also, I realised that for now, I have timestamps for all TimelineEvents. So this should probably be dealt with in `TimelineEvent.js`. And that means I can remove some Relay Fragment containers as they are then useless. (I think in the current state, it is only justified to use it for LeaveForAirport, as `location` field is specific to this event). Thoughts?

app/MMB/src/scenes/timeline/events/LeaveForAirportTimelineEvent.js
RobinCsl: I will use StyleSheet. I had thought about it and then forgot. Sorry about that.

app/MMB/src/scenes/timeline/events/LeaveForAirportTimelineEvent.js
mrtnzlml: Yes, definitely. You don't need to work with the data from API here. You only need to tweak here styles, icons and translations and this moment.

app/MMB/src/scenes/timeline/events/LeaveForAirportTimelineEvent.js
```diff
@@ -0,0 +1,36 @@
...
+  +data: TimelineEventType,
+|};
+
+const LeaveForAirportTimelineEvent = (props: Props) => {
+  return (
+    <TimelineEvent
+      data={props.data}
+      iconVertLines={<TimelineEventIcon name="directions-walk" />}
+      mainContent={
+        <TranslationFragment>
```
mrtnzlml: Why do you use `TranslationFragment` here?

RobinCsl: I thought it was necessary if you had more than one `Translation` component in the children or further down the tree. I will replace it with `React.Fragment` then.

app/MMB/src/scenes/timeline/events/BookedFlightTimelineEvent.js
```diff
@@ -1,42 +1,21 @@
 // @flow
 
 import * as React from 'react';
-import idx from 'idx';
-import { graphql, createFragmentContainer } from '@kiwicom/mobile-relay';
-import { Translation, DateFormatter } from '@kiwicom/mobile-localization';
+import { Translation } from '@kiwicom/mobile-localization';
 
-import type { BookedFlightTimelineEvent as BookedFlightTimelineEventType } from './__generated__/BookedFlightTimelineEvent.graphql';
+import type { TimelineEvent as TimelineEventType } from '../__generated__/TimelineQuery.graphql';
```
mrtnzlml: Are you sure `TimelineEvent` type exists? I wasn't able to find it. Also, you can see from this diff the prop-drilling (`data` are not used - only passed down). That looks weird since we have fragments and every component can fetch what's needed in that component. (not fetching for the other components)

app/MMB/src/scenes/timeline/events/LeaveForAirportTimelineEvent.js
mrtnzlml: It's necessary only when you require the type of the translation. That's not the case for the `mainContent` property.

app/MMB/src/scenes/timeline/events/BookedFlightTimelineEvent.js
RobinCsl: Oh right, so I should use a Relay Fragment Container on the `TimelineEvent` component, since that's where I handle the timestamp. And remove it from there. That makes sense. Thank you!

PR #749, opened 2018-06-20
Display Day banners when necessary in the timeline

RobinCsl:


packages/localization/src/DateUtils.js
```diff
@@ -25,4 +25,16 @@ DateUtils.diffInDays = (higherDate: Date, lowerDate: Date) => {
   return Math.ceil(diffMs / (1000 * 3600 * 24));
 };
 
+/**
+ *
+ * Returns whether firstDate and secondDate are the same day in local time
+ *
+ */
```
mrtnzlml: This comment is not only quite big but probably not necessary? 😃

RobinCsl: Agreed, I had a more complicated function before, so that's why I had thought about describing it. I will remove it :smile:

packages/localization/src/DateUtils.js
```diff
@@ -25,4 +25,10 @@ DateUtils.diffInDays = (higherDate: Date, lowerDate: Date) => {
   return Math.ceil(diffMs / (1000 * 3600 * 24));
 };
 
+DateUtils.isSameDay = (firstDate: Date, secondDate: Date): boolean => {
+  const localizedFirstDate = firstDate.toLocaleDateString();
+  const localizedSecondDate = secondDate.toLocaleDateString();
+  return localizedFirstDate === localizedSecondDate;
```
mrtnzlml: Why don't you just compare it? I don't see a point in comparing localized strings here.

mrtnzlml: Why do you need `toLocaleDateString` for it?

RobinCsl: The way I understand it is that getDate() or getMonth() will just extract naively from the UTCZ timestamp. So for example:
2018-12-23T23:00:00.000Z is 2018-12-24T01:00:00.000+02:00

mrtnzlml: Why don't you try it? 🙂

```
> new Date('2018-12-23T23:00:00.000Z').toLocaleDateString('cs')
'2018-12-24'
```

RobinCsl: I am confused by your comment. That's what my `isSameDay` function does, comparing the output of `toLocaleDateString`.

mrtnzlml: Aha, I didn't notice it's called `isSameDay` and not `isSameDate`...

PR #750, opened 2018-06-21
MMB: prepare insurance overview screen

mrtnzlml:


app/MMB/src/scenes/tripServices/insurance/InsuranceOverview.js
```diff
@@ -0,0 +1,44 @@
...
+import DestinationImage from './DestinationImage';
+import TripInfo from '../../../components/header/TripInfo';
+import BookingDetailContext from '../../../context/BookingDetailContext';
+import type { InsuranceOverviewQueryResponse } from './__generated__/InsuranceOverviewQuery.graphql';
+
+function InsuranceOverview(response: InsuranceOverviewQueryResponse) {
+  return (
+    <LayoutSingleColumn>
+      <DestinationImage data={response.node} />
+      <TripInfo data={response.node} />
```
RobinCsl: So this is just a dummy view for the time being? It does not seem strictly related to insurance.

app/MMB/src/scenes/tripServices/insurance/InsuranceOverview.js
```diff
@@ -0,0 +1,44 @@
+// @flow strict
+
+import * as React from 'react';
+import { LayoutSingleColumn } from '@kiwicom/mobile-shared';
```
RobinCsl: Not strictly related to this PR, but I am not sure in my own Timeline view I use this `LayoutSingleColumn`. Should I?

mrtnzlml: You should probably just align your container to the center.

app/MMB/src/scenes/tripServices/insurance/InsuranceOverview.js
mrtnzlml: I am not sure if I understand. This component is reused from the header because this is what we need according to the design.

RobinCsl: Oh right, sorry.

PR #754, opened 2018-06-21
MMB insurance: prepare missing screens (no design yet)

mrtnzlml:


app/MMB/src/scenes/tripServices/insurance/InsuranceOverviewScene.js
```diff
@@ -0,0 +1,135 @@
...
+                onPress={this.goToTheInsuranceSelection}
+              />
+              <MenuItem
+                title={
+                  <Translation passThrough="TODO: list of PAX (links to the insurance selection)" />
+                }
+                onPress={this.goToTheInsuranceSelection}
+              />
+            </TitledMenuGroup>
+            <View style={{ padding: 10 }}>
```
RobinCsl: Do you want your style to be inline here?

app/MMB/src/scenes/tripServices/insurance/InsuranceOverviewScene.js
```diff
@@ -0,0 +1,135 @@
...
+              <TextButton
+                title={
+                  <Translation id="mmb.trip_services.order.process_to_refund" />
+                }
+                onPress={this.goToTheInsuranceRefund}
+              />
+            </View>
+            <View style={{ padding: 10 }}>
```
RobinCsl: Idem

app/MMB/src/scenes/tripServices/insurance/InsuranceOverviewScene.js
mrtnzlml: You are right it should be done differently. However, this is all WIP so can we skip it here, please? 🙂

PR #755, opened 2018-06-21
Android: Update distribution docs

ferrannp:


RNAndroidPlayground/README.md
```diff
@@ -1,29 +1,81 @@
 # Android distribution
 
-## Deploy rnkiwimobile
+## Deploy rnkiwimobile and dependencies
 
-1. Generate JS/assets files from the root of this repo:
+GitlabCI will take care of this anytime we merge something to `master`. It will skip it if the version(s) has not changed in the `package.json`.
```
mrtnzlml: This is not very clear. We have to deploy the app itself every time (like we do with iOS). Only the deps are deployed after they'll change, right?

ferrannp: Nop. In this case, it deploys the app (`rnkiwimobile`) when you change version in `package.json` so it's not like iOS frameworks.

mrtnzlml: So no continuous delivery as we wanted then? Do I understand correctly?

ferrannp: Not for every merge in `master`. Several reasons:

- As we talked with @Blackhex you would deploy way to many `SNAPSHOT(s)` to trinerdis. Probably you have a space limitation?. Unless there is a way to delete / replace.
- Also, now we deploy every commit to `master` an `iOS` framework. How many of them are being ignored (not use by native time)? Probably most of them.

I think it make sense to do a full release every time you bump the version https://github.com/kiwicom/mobile/blob/master/.build/package.json#L3.

If we want real continuous delivery for only JS changes. We might be better looking at things like CodePush.

WDYT?

mrtnzlml: Why do we have a continuous delivery for iOS if it doesn't make sense? 🙂

ferrannp: I guess we have it because it is possible (we use Github space anyway). So iOS team could be copying the framework everytime if they want. If we want CodePush, I guess we will adapt that. But you know what I mean here right?

ferrannp: @Blackhex is there a way in Trinerdis we can delete and add new version of `rnkiwimobile` without bloating it?

mrtnzlml: Sure. Just trying to figure out where did our goal change from _"let's release it as often as possible so they can use it immediately"_ to _"let's do it only for iOS"_... :)

We are off-topic now. You can merge the PR if it's not WIP anymore.

Blackhex: I am not sure if I understand the issue hidden in previous comments completely or if you understand each other but it should be completely fine if new `rnkiwimobile` package is deployed everytime someone pushes to master. Only the packages of external dependencies (like e.g. `react-native-maps`) do not need to be updated for each push to `master` if the version of the dependency does not change, hence there should be some kind version check of already deployed package (for the dependency).

I'll check if there is some way how to clean repository from outdated packages to prevent reaching space limitations, though.

mrtnzlml: ^^ this makes much more sense to me :)

Blackhex: Apparently it is https://cwiki.apache.org/confluence/display/ARCHIVA/Repository+Purge, I'll try it tomorrow...

PR #760, opened 2018-06-25
Profile: Close keyboard when submitting login form

tbergq:
Basically when debugging on device, after login, I had to shake device and reload to be able to navigate to mmb tab. Keyboard was on top of the tabs, and could not be closed.

### DIFF
--- a/app/profile/index.js
+++ b/app/profile/index.js
@@ -1,32 +1,66 @@
 // @flow
 
 import * as React from 'react';
-import { AsyncStorage, Alert } from 'react-native';
-import { SimpleCard, LayoutSingleColumn } from '@kiwicom/mobile-shared';
+import { AsyncStorage, View } from 'react-native';
+import {
+  SimpleCard,
+  LayoutSingleColumn,
+  DismissKeyboardView,
+  StyleSheet,
+} from '@kiwicom/mobile-shared';
 
 import Login from './components/authentication/Login';
 import Logout from './components/authentication/Logout';
 
-export default class LoginScreen extends React.Component<{||}> {
+type State = {|
+  isLoggedIn: boolean,
+|};
+
+export default class LoginScreen extends React.Component<{||}, State> {
+  state = {
+    isLoggedIn: false,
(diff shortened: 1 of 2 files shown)


app/profile/index.js
```diff
@@ -2,15 +2,32 @@
...
+|};
+
+export default class LoginScreen extends React.Component<{||}, State> {
+  state = {
+    isLoginSuccess: false,
+  };
+
   onLogin = (token: string) => {
     AsyncStorage.setItem('mobile:MMB-Token', token);
-    Alert.alert('Successfully logged in');
```
tbergq: Alert made the keyboard open again 🙈

app/profile/index.js
```diff
@@ -21,12 +38,17 @@ export default class LoginScreen extends React.Component<{||}> {
...
-      </LayoutSingleColumn>
+      <DismissKeyboardView>
+        <LayoutSingleColumn>
+          <SimpleCard>
+            <Login onLogin={this.onLogin} />
+            <Logout onLogout={this.onLogout} />
+          </SimpleCard>
+          {this.state.isLoginSuccess && (
+            <Translation passThrough="Successfully logged in" />
+          )}
```
mrtnzlml: What about displaying login for if logged in and logout button otherwise? It would make more sense to me...

tbergq:
@mrtnzlml please check again

app/profile/index.js
```diff
@@ -1,32 +1,66 @@
...
+          </LayoutSingleColumn>
+        </DismissKeyboardView>
+      </View>
     );
   };
 }
+
+const styles = StyleSheet.create({
+  container: {
+    marginTop: 60,
```
mrtnzlml: Why is the navigation header missing? Wouldn't be better to add it properly to the navigation? :)

tbergq: Not sure why it is missing. I would say it is not important now. I just need to be able to click the logout button on an iPhone X, hence the margin.

PR #762, opened 2018-06-25
Adding Departure and Arrival

RobinCsl:
And correcting icons style, now that I can run it on iOS

### DIFF
--- a/app/MMB/src/scenes/timeline/MainContentNote.js
+++ b/app/MMB/src/scenes/timeline/MainContentNote.js
@@ -1,0 +1,25 @@
+// @flow
+
+import * as React from 'react';
+import { Text, Color, StyleSheet } from '@kiwicom/mobile-shared';
+import { type TranslationType } from '@kiwicom/mobile-localization';
+
+type Props = {|
+  +children: ?TranslationType,
+|};
+
+const Note = (props: Props): React.Node => {
+  if (props.children) {
+    return <Text style={styles.note}>{props.children}</Text>;
+  }
+  return null;
+};
+
+export default Note;
+
+const styles = StyleSheet.create({
+  note: {
+    fontSize: 13,
+    color: Color.ink.light,
+  },
+});
--- a/app/MMB/src/scenes/timeline/Timeline.js
+++ b/app/MMB/src/scenes/timeline/Timeline.js
@@ -11,6 +11,8 @@ import BookingDetailContext from '../../context/BookingDetailContext';
 import BookedFlightTimelineEvent from './events/BookedFlightTimelineEvent';
 import AirportArrivalTimelineEvent from './events/AirportArrivalTimelineEvent';
 import DownloadInvoiceTimelineEvent from './events/DownloadInvoiceTimelineEvent';
+import DepartureTimelineEvent from './events/DepartureTimelineEvent';
+import ArrivalTimelineEvent from './events/ArrivalTimelineEvent';
 import LeaveForAirportTimelineEvent from './events/LeaveForAirportTimelineEvent';
 import BoardingTimelineEvent from './events/BoardingTimelineEvent';
 import DaySeparator from './DaySeparator';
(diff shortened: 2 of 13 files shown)


app/MMB/src/scenes/timeline/events/DepartureTimelineEvent.js
```diff
@@ -0,0 +1,124 @@
...
+
+import type { DepartureTimelineEvent as DeparturelineEventType } from './__generated__/DepartureTimelineEvent.graphql';
+import TimelineEvent from '../TimelineEvent';
+import TimelineEventIcon from '../TimelineEventIcon';
+
+type Props = {|
+  +data: DeparturelineEventType,
+|};
+
+function renderNote(duration: number): React.Element<typeof Text> {
```
tbergq: This looks like it could be a separate component in a seperate file

app/MMB/src/scenes/timeline/events/DepartureTimelineEvent.js
```diff
@@ -0,0 +1,124 @@
...
+        <Translation
+          id="mmb.booking_timeline.event.departure.note_hours_minutes"
+          values={{ hours, minutes }}
+        />
+      );
+    }
+  }
+
+  return <Text style={styles.note}>{note}</Text>;
+}
```
tbergq: Is there some better way of doing this, it is a lot of nested else and ifs.

app/MMB/src/scenes/timeline/events/DepartureTimelineEvent.js
```diff
@@ -0,0 +1,124 @@
...
+
+const DepartureTimelineEvent = (props: Props) => {
+  const timestamp = idx(props, _ => _.data.timestamp);
+  const destination = idx(props, _ => _.data.location.airport.city.name);
+  const duration = idx(props, _ => _.data.duration);
+
+  const note = duration ? renderNote(duration) : null;
```
tbergq: this logic could live in the component, if it is not passed any duration, it could return null

app/MMB/src/scenes/timeline/events/DepartureTimelineEvent.js
RobinCsl: I think this note is really specific to this timeline event. 
What could be abstracted in a different file would be this: 
```
const Note = (note) => <Text style={styles.note}>{note}</Text>
```
as it is used in several timeline events.

RobinCsl:
Updated the PR:
- Abstracted <Text style={styles.note} />  into a component, and used it where necessary
- Improved the logic on determining the translation id in DownloadInvoice and Departure components.

Please have a second look @tbergq, thank you :)

app/MMB/src/scenes/timeline/MainContentNote.js
```diff
@@ -0,0 +1,25 @@
...
+
+import * as React from 'react';
+import { Text, Color, StyleSheet } from '@kiwicom/mobile-shared';
+import { type TranslationType } from '@kiwicom/mobile-localization';
+
+type Props = {|
+  +children: ?TranslationType,
+|};
+
+const Note = (props: Props): ?React.Node => {
```
tbergq: React.Node includes null, so you don't need the ?React.Node

RobinCsl: Ok, I will remove it.

RobinCsl: Done

app/MMB/src/scenes/timeline/events/DepartureTimelineEvent.js
```diff
@@ -0,0 +1,88 @@
...
+import Note from '../MainContentNote';
+import type { DepartureTimelineEvent as DeparturelineEventType } from './__generated__/DepartureTimelineEvent.graphql';
+import TimelineEvent from '../TimelineEvent';
+import TimelineEventIcon from '../TimelineEventIcon';
+
+type Props = {|
+  +data: DeparturelineEventType,
+|};
+
+function renderNote(duration: ?number): ?React.Element<typeof Translation> {
```
tbergq: I still think this should be a component.

app/MMB/src/scenes/timeline/events/DepartureTimelineEvent.js
```diff
@@ -0,0 +1,88 @@
...
+      iconVertLines={<TimelineEventIcon name="flight-takeoff" />}
+      mainContent={
+        <React.Fragment>
+          <Translation
+            id="mmb.booking_timeline.event.departure.title"
+            values={{
+              destination,
+            }}
+          />
+          <Note>{renderNote(duration)}</Note>
```
tbergq:
```

PR #764, opened 2018-06-26
Hotels: Fix isDateBeforeToday bug

tbergq:


app/hotels/src/search/SearchQueryHelpers.js
```diff
@@ -40,7 +40,7 @@ export const getSearchQueryParams = (
 
 export const isDateBeforeToday = (date: Date) => {
   const today = new Date();
-  return date < today && date.getUTCDay() !== today.getUTCDay();
```
tbergq: getUTCDay() returns the number of the day, i.e. tuesday is always 2, so for past date that was same week day, it would return wrong result.

tbergq:
Ok, thanks for pointing this out. I will try to solve it differently.

tbergq:
I added some new functions to date utils, isBeforeDate & isAfterDate. 

I also found another issue, that if dates passed from native code to AllHotels would be null or undefined, we should really set a date, so we do not search with invalid dates. 

please have a second look @RobinCsl

PR #766, opened 2018-06-26
Hotels: Filter stripe refactor + fix scroll to start

tbergq:


app/hotels/src/filter/Filters.js
```diff
@@ -0,0 +1,27 @@
...
+type Props = {|
+  +children: React.Node,
+|};
+
+export default function Filters(props: Props) {
+  const activeFilters = [];
+  const inActiveFilters = [];
+
+  React.Children.forEach(props.children, child => {
+    if (child.props.isActive) {
```
mrtnzlml: Nice! Finally, someone figured out how to refactor it... 🙂

app/hotels/src/filter/Filters.js
```diff
@@ -0,0 +1,27 @@
...
+
+import * as React from 'react';
+
+type Props = {|
+  +children: React.Node,
+|};
+
+export default function Filters(props: Props) {
+  const activeFilters = [];
+  const inActiveFilters = [];
```
RobinCsl: If I understand correctly, this should read _inactive_ and not _in active_, right? In that case, I would replace `inActive` by `inactive`.

tbergq: you are of course right, i will change it to lowercase `a`

tbergq:
@RobinCsl grammar fixed

PR #767, opened 2018-06-26
Added Boarding event; fixed TimelineEventIcon to display Icon or Text…

RobinCsl:
…Icon

Note: There probably will be some conflicts with my previous PR as I had to modify some of the same files. I will fix them after the first PR is landed.

### DIFF
--- a/app/MMB/src/scenes/timeline/TimelineEventIcon.js
+++ b/app/MMB/src/scenes/timeline/TimelineEventIcon.js
@@ -2,25 +2,21 @@
 
 import * as React from 'react';
 import { View } from 'react-native';
-import { StyleSheet, Color, Icon } from '@kiwicom/mobile-shared';
-
-const size = 17;
+import { StyleSheet, Color, Icon, TextIcon } from '@kiwicom/mobile-shared';
 
 export type Props = {|
-  +name: string,
+  +icon: React.Element<typeof TextIcon | typeof Icon>,
   +isFirst?: boolean,
   +isPastEvent?: boolean,
 |};
 
 export default function TimelineEventIcon(props: Props) {
+  const icon = React.cloneElement(props.icon, {
+    style: [props.icon.props.style, styles.icon],
+  });
   const pastEventIcon = (
-    <Icon
-      name="check"
-      size={size}
-      style={[styles.icon, styles.pastEventIcon]}
-    />
+    <TextIcon code="V" style={[styles.icon, styles.pastEventIcon]} />
   );
-  const icon = <Icon name={props.name} size={size} style={styles.icon} />;
   return (
     <View style={styles.container}>
       <View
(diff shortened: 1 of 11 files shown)


app/MMB/src/scenes/timeline/TimelineEventIcon.js
```diff
@@ -2,25 +2,27 @@
...
 import * as React from 'react';
 import { View } from 'react-native';
-import { StyleSheet, Color, Icon } from '@kiwicom/mobile-shared';
+import { StyleSheet, Color, Icon, TextIcon } from '@kiwicom/mobile-shared';
 
 const size = 17;
 
 export type Props = {|
-  +name: string,
+  +type: 'Icon' | 'TextIcon',
```
tbergq: I think we should rather pass in the Icon component here.

RobinCsl: Ok, I will refactor this.

app/MMB/src/scenes/timeline/events/BoardingTimelineEvent.js
```diff
@@ -0,0 +1,46 @@
...
+  BoardingTimelineEvent,
+  graphql`
+    fragment BoardingTimelineEvent on BoardingTimelineEvent {
+      timestamp
+    }
+  `,
+);
+
+const styles = StyleSheet.create({
+  note: { fontSize: 13, color: Color.ink.light },
```
tbergq: I should probably be an eslint-rule, that is much more readable

```
note: { 
fontSize: 13, 
color: Color.ink.light,
},
```

RobinCsl: We can add 
```
"object-property-newline": ERROR
```
to `.eslintrc.js` to fix this. I agree with you.

tbergq: Yes, please do so

RobinCsl: I'll make a different PR for this, it is modifying too many files after I add this rule.

tbergq: Right, ok, then I would say leave it for now. Not that important. 

Some times it makes sense to write it on one line I guess.

RobinCsl:
I have refactored the way to pass icons. 

**Note**: This PR needs the previous one to be merged before, and the resulting conflicts to be fixed, before it can be merged.

app/MMB/src/scenes/timeline/events/BoardingTimelineEvent.js
RobinCsl: The notes are taken care of in the other PR https://github.com/kiwicom/mobile/pull/762

PR #769, opened 2018-06-26
Adding 'object-property-newline' to eslint config

RobinCsl:


tbergq:
I think sometimes this rule does not make sense, so we should not merge it.

PR #774, opened 2018-06-26
Hotels: Improve children age control

### LINKED ISSUE #773: Hotels: Show less than a year instead of 0
[Trello](https://trello.com/c/YbLi5kW5/274-nice-to-have-when-i-select-less-than-year-would-be-nice-to-display-less-than-year-and-not-zero-because-then-i-dont-know-what-is)

tbergq:
closes #773


packages/localization/src/vocabularies/Hotels.js
```diff
@@ -41,12 +41,15 @@ export default {
   'hotels_search.filter.stars_popup.title': 'Hotel stars',
   'hotels_search.filter.stars_popup.unrated': 'Unrated',
   'hotels_search.guests.adults': 'Adults:',
+  'hotels_search.guests.age_control.less_than_one': 'Less than year',
```
RobinCsl: __Less than a year__ is probably more correct

PR #778, opened 2018-06-27
Shared: Datepicker set UTC date

### LINKED ISSUE #777: Hotels: Datepicker gives wrong date
I pick today, and it shows yesterday

tbergq:
closes #777

### DIFF
--- a/packages/shared/src/forms/datePicker/__tests__/DatePicker.test.ios.js
+++ b/packages/shared/src/forms/datePicker/__tests__/DatePicker.test.ios.js
@@ -25,7 +25,7 @@ describe('DatePicker ios', () => {
   it('handles on change with date object', () => {
     const wrapper = getComponent();
     const instance = wrapper.getInstance();
-    const date = new Date(2);
+    const date = new Date(Date.UTC(2018, 1, 1));
 
     instance.onDateChange(date);
     expect(instance.state.date).toEqual(date);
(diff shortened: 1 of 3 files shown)


packages/shared/src/forms/datePicker/__tests__/DatePicker.test.ios.js
```diff
@@ -34,9 +34,18 @@ describe('DatePicker ios', () => {
...
+    const date = new Date(Date.UTC(2018, 2, 2));
 
     instance.onDateChange(date.toISOString());
     expect(instance.state.date).toEqual(date);
   });
+
+  it('handles UTC', () => {
+    const wrapper = getComponent();
+    const instance = wrapper.getInstance();
+    const date = new Date('Wed Jun 27 2018 00:00:00 GMT+0200 (CEST)');
```
RobinCsl: This is actually: `2018-06-26T22:00:00.000Z`, so that's why the test fails

tbergq: I know, but it needs to be set to 27th from `instance.onDateChange(date)` 

When I pick 27th of June from the datepicker, I expect to see 27th of June in the form, and not 26th.

This test passed locally, but there is another timezone on our CI. So I need to test some more..

tbergq:
We probably need to rethink the entire DateFormatter & DateUtils

PR #779, opened 2018-06-27
Adding DownloadETicket Event

RobinCsl:


app/MMB/src/components/ETicket.js
```diff
@@ -0,0 +1,51 @@
...
+
+    if (!uri) {
+      return (
+        <GeneralError
+          errorMessage={<Translation id="mmb.tickets.not_available" />}
+        />
+      );
+    }
+
+    return <PdfViewer uri={uri} />;
```
tbergq: We already have a quite similar component called `ETicketPdf.js`

This takes the uri, and stores the pdf, so you don't need to download it again each time you open it. Maybe you could re-use that component?

RobinCsl: I am using it by navigating to the same route you use in your other component. Hope that's how it's meant to be done.

tbergq: That's looks good to me 😄

Issue #738, opened 2018-06-12
GetDeviceLocale.js may return unexpected values


mrtnzlml:
For example this: `"es_SA@calendar=gregorian"`. We do not expect this and we should refactor it ASAP...

tbergq:
Does this happen for iOS and/or Android?

mrtnzlml:
I think this is from my iOS device.

tbergq:
Interesting, I just tested from my device, and I get, en-GB as expected.

@ferrannp you have some experience with this?

mrtnzlml:
@tbergq It's not that hard to set it - just use for example Arabic language for the device in Europe (very common in Germany for example)...

tbergq:
That gave me ar-ES

PR #780, opened 2018-06-27
Avoid appending calendar in device info locale

lukewalczak:
Related to: #738 

Currently `GetDeviceLocale` always will return object:
```
{
  Format: {
    Underscored: 'en_US',
    Dashed: 'en-US',
  },
  Language: 'en',
  Territory: 'US',
},
```
where `Format.Underscored` and `Format.Dashed` are a combination of `Language` and `Territory`. In that way we are able to avoid appending unnecessary `calendar=gregorian`.

### DIFF
--- a/app/MMB/src/navigation/DetailShareButton.js
+++ b/app/MMB/src/navigation/DetailShareButton.js
@@ -2,7 +2,7 @@
 
 import * as React from 'react';
 import { Share, Platform } from 'react-native';
-import { DateFormatter, GetDeviceLocale } from '@kiwicom/mobile-localization';
+import { DateFormatter, DeviceInfo } from '@kiwicom/mobile-localization';
 
 import BookingDetailContext from '../context/BookingDetailContext';
 import ShareButton from '../components/ShareButton';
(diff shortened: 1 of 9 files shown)


tbergq:
Could we rather expose more functions?

something like `getLanguage()` , `getTerritory()` , `getLocaleUnderscored()` , `getLocaleDashed()`

lukewalczak:
Sure!

app/MMB/src/navigation/DetailShareButton.js
```diff
@@ -14,7 +14,7 @@ type PropsWithContext = {|
 
 class DetailShareButton extends React.Component<PropsWithContext> {
   onPress = () => {
-    const lang = GetDeviceLocale().split('-')[0] || 'en';
+    const lang = GetDeviceLocale.getLanguage() || 'en';
```
tbergq: Before this file was just about DeviceLocale, but now it is about more. Maybe it would be better to rename it to DeviceInfo

`DeviceInfo.getLanguage()` makes more sense than `GetDeviceLocale.getLanguage()`

lukewalczak: Agree, now when functions are called `get`, the prefix in `DeviceInfo` is unnecessary.

PR #783, opened 2018-06-27
Top and bottom line display depending on the position of the event in…

RobinCsl:
… the timeline

### DIFF
--- a/app/MMB/src/scenes/timeline/Timeline.js
+++ b/app/MMB/src/scenes/timeline/Timeline.js
@@ -4,7 +4,7 @@ import * as React from 'react';
 import { ScrollView } from 'react-native';
 import idx from 'idx';
 import { PrivateApiRenderer, graphql } from '@kiwicom/mobile-relay';
-import { DateUtils } from '@kiwicom/mobile-localization';
+import { DateFormatter } from '@kiwicom/mobile-localization';
 
 import type { TimelineQueryResponse } from './__generated__/TimelineQuery.graphql';
 import BookingDetailContext from '../../context/BookingDetailContext';
@@ -20,13 +20,33 @@ import BoardingTimelineEvent from './events/BoardingTimelineEvent';
 import TransportFromAirportTimelineEvent from './events/TransportFromAirportTimelineEvent';
 import DaySeparator from './DaySeparator';
 import DownloadETicketTimelineEvent from './events/DownloadETicketTimelineEvent';
+import TimelineEventIconContext from '../../context/TimelineEventIconContext';
 
 export const TimelineSubmenuItems = {
   ...TimelineInvoiceSubmenuItems,
 };
 
-function getValidTimelineEvent(data) {
-  if (data && data.__typename) {
+type Events = $PropertyType<
+  $NonMaybeType<$PropertyType<TimelineQueryResponse, 'bookingTimeline'>>,
+  'events',
+>;
+
+type Event = $ElementType<$NonMaybeType<Events>, number>;
+
+type TimelineEvent = ?React.Element<
+  | typeof BookedFlightTimelineEvent
+  | typeof AirportArrivalTimelineEvent
+  | typeof DownloadInvoiceTimelineEvent
+  | typeof DepartureTimelineEvent
+  | typeof ArrivalTimelineEvent
+  | typeof LeaveForAirportTimelineEvent
+  | typeof BoardingTimelineEvent
+  | typeof TransportFromAirportTimelineEvent
+  | typeof DownloadETicketTimelineEvent,
+>;
+
+function getValidTimelineEvent(data: Event): TimelineEvent {
+  if (data) {
     switch (data.__typename) {
       case 'BookedFlightTimelineEvent':
         return <BookedFlightTimelineEvent data={data} />;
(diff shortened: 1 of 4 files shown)


app/MMB/src/scenes/timeline/Timeline.js
```diff
@@ -53,35 +54,53 @@ function getValidTimelineEvent(data) {
...
+    const prevTimestamp = idx(prevRenderedEvent, _ => _.props.data.timestamp);
+    if (timestamp && prevTimestamp) {
+      const eventTimestamp = new Date(timestamp);
+      const prevEventTimestamp = new Date(prevTimestamp);
+      if (!DateUtils.isSameDay(eventTimestamp, prevEventTimestamp)) {
+        return <DaySeparator date={eventTimestamp} />;
+      }
     }
   }
   return null;
```
tbergq: Could you please try to refactor this function a bit? The level of nested if/else makes it hard to read and understand.

RobinCsl: Ok, will do.

app/MMB/src/scenes/timeline/Timeline.js
```diff
@@ -53,35 +54,53 @@ function getValidTimelineEvent(data) {
...
-          daySeparator = renderDaySeparator(event, events[index - 1]);
-        }
-        return (
-          <React.Fragment key={'TimelineEvent-' + event.__typename + index}>
-            {daySeparator}
-            {getValidTimelineEvent(event)}
-          </React.Fragment>
+    const renderedEvents = events
+      .map(event => getValidTimelineEvent(event))
+      .filter(renderedEvent => renderedEvent);
```
tbergq: Why is this needed?

app/MMB/src/scenes/timeline/TimelineEventIcon.js
```diff
@@ -4,9 +4,13 @@ import * as React from 'react';
 import { View } from 'react-native';
 import { StyleSheet, Color, Icon, TextIcon } from '@kiwicom/mobile-shared';
 
+export const TimelineEventIconContext = React.createContext({
+  displayTopLine: true,
+  displayBottomLine: true,
+});
+
```
tbergq: What is this context used for? 
You set two booleans, why can it not be variables?

app/MMB/src/scenes/timeline/Timeline.js
RobinCsl: There could be some additional events given by the GraphQL server which would not correspond to any implemented event component.
There could be situations where that adds an additional day banner when no component is being rendered, as `getValidTimelineEvent` would return `null`; also, since the first and last events need to display the vertical bars in a different way, the length of the non-null components must be taken into account, not the length of the events.
Does that make sense?

app/MMB/src/scenes/timeline/TimelineEventIcon.js
RobinCsl: This is to indicate default values. But that context is set depending on the index of the rendered event components.

app/MMB/src/scenes/timeline/Timeline.js
tbergq: Yes

tbergq: Maybe you could add a comment explaining it

app/MMB/src/scenes/timeline/Timeline.js
RobinCsl: Done

RobinCsl:
We will need to revisit DateUtils and DateFormatter.
This is the current state of the timeline for one journey:
simulator screen shot - iphone 6 - 2018-06-27 at 18 23 36

And this is a screenshot from the currently deployed app:
img_0274

We can see the dates and times are off.

tbergq:
> We will need to revisit DateUtils and DateFormatter.

I think our DateFormatter and DateUtils are wrong. 

If I fly from Barcelona - Lima with return flight, and we format the flight time in UTC it will give me the wrong time. I am interested in seeing the actual time the flight takes off in that timezone.

@mrtnzlml please correct me if I am wrong.

app/MMB/src/scenes/timeline/Timeline.js
```diff
@@ -53,35 +54,47 @@ function getValidTimelineEvent(data) {
...
+            timestamp &&
+            DateUtils.isSameDay(
+              new Date(currentTimestamp),
+              new Date(timestamp),
+            ),
         );
+        if (index === sameDays.indexOf(true)) {
+          daySeparator = <DaySeparator date={new Date(currentTimestamp)} />;
+        }
       }
```
tbergq: It seems like you map every event to get the type, then loop to remove nulls, then loop to get all timestamps. Then loop over the events, and for each event loop the timestamps.

Would it be possible here to use reduce, loop once and return an object.

```
events.reduce((acc, curr) => {
const event = getValidTimelineEvent(curr);
if (!event) return acc;
// attach curr to acc and return
}, {})
```

having the structure like {'2018-06-01': { events: [...] }, '2018-06-02': { events : [...]}

Then loop over the keys in the object and between each key add the day separator.

Do you think this makes sense?

RobinCsl: Yes, that's a much cleaner way of doing it. I will give it a try. Thank you!

mrtnzlml:
@tbergq I don't know man. Are you sure you are not doing something dodgy there? We should always get the date in the UTC format (at least from our GraphQL):

```json
{
  "data": {
    "allFlights": {
      "edges": [
        {
          "node": {
            "arrival": {
              "localTime": "2018-08-10T00:40:00.000Z",
              "time": "2018-08-10T05:40:00.000Z"
            },
            "departure": {
              "localTime": "2018-08-08T23:55:00.000Z",
              "time": "2018-08-08T21:55:00.000Z"
            }
          }
        },
...
```

It's defined by the `DateTime` type in GraphQL:

> A date-time string at UTC, such as 2007-12-03T10:15:30Z, compliant with the date-time format outlined in section 5.6 of the RFC 3339 profile of the ISO 8601 standard for representation of dates and times using the Gregorian calendar.

So this is why it's so important to call the Intl with UTC timezone (so it will stay in UTC and it won't be converted into local timezone):

```
Intl.DateTimeFormat('en', {timeZone: 'UTC', year: 'numeric', month: 'numeric', day: 'numeric', hour: 'numeric', minute: 'numeric'}).format(new Date('2018-08-10T00:40:00.000Z'))
> "8/10/2018, 12:40 AM"   (correct)

BUT

Intl.DateTimeFormat('en', {year: 'numeric', month: 'numeric', day: 'numeric', hour: 'numeric', minute: 'numeric'}).format(new Date('2018-08-10T00:40:00.000Z'))
> "8/10/2018, 2:40 AM"   (WTF? It's not true...)
```

Maybe I don't understand what it's wrong exactly. Can you please elaborate it more with some example?

PR #785, opened 2018-06-27
Displaying location below date possible; highlighted date and locatio…

RobinCsl:
…n for AirportArrival and Arrival events

### DIFF
--- a/app/MMB/src/scenes/timeline/TimelineEventDateLocation.js
+++ b/app/MMB/src/scenes/timeline/TimelineEventDateLocation.js
@@ -2,22 +2,48 @@
 
 import * as React from 'react';
 import { View } from 'react-native';
-import { StyleSheet, Text } from '@kiwicom/mobile-shared';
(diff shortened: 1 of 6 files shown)


app/MMB/src/scenes/timeline/TimelineEventDateLocation.js
```diff
@@ -2,23 +2,51 @@
 
 import * as React from 'react';
 import { View } from 'react-native';
-import { StyleSheet, Text } from '@kiwicom/mobile-shared';
+import { StyleSheet, Text, Color } from '@kiwicom/mobile-shared';
 import { Translation } from '@kiwicom/mobile-localization';
 
+export const TimelineEventDateLocationContext = React.createContext({
+  highlightText: false,
+});
```
tbergq: Can we create contexts in a separate files? If I wanted to find this context comes from, this is not the first place I would look.

RobinCsl: I remedied to that and used the aptly named `context` folder :)

PR #796, opened 2018-06-29
Hotels: Using UTC dates for search form

tbergq:


mrtnzlml:
@tbergq Would be great to have a tests for it since you are fixing a bug here. Maybe you'll decide to use some library for it later but it should still work the same...

tbergq:
Yes, you are right, I will add some test for it

tbergq:
@mrtnzlml 🚀

packages/localization/src/DateFormatter.js
```diff
@@ -46,7 +47,7 @@ function pad(number) {
  * formatter. Every function is localized and you should never call formatting
  * on the raw Date object.
  */
-function DateFormatter(rawDate: Date = new Date()) {
+function DateFormatter(rawDate: Date = DateUtils.getUTCNow()) {
```
mrtnzlml: I still believe this method should not accept the `Date` type... :)

tbergq: Yes, maybe for a later refactor.

PR #808, opened 2018-07-03
MMB: Add download boarding pass button

tbergq:


app/MMB/src/scenes/tickets/boardingPasses/DownloadButton.js
```diff
@@ -0,0 +1,62 @@
...
+    backgroundColor: Color.brand,
+    height: 44,
+    borderRadius: 4,
+    marginTop: 15,
+  },
+  buttonText: {
+    color: Color.white,
+  },
+  disabled: {
+    backgroundColor: '#b8c1cc',
```
RobinCsl: I think we should take care of the colors file to make sure we have all of them, right?

tbergq: Yes, I think that makes sense, I can add a disabled color.

PR #811, opened 2018-07-03
[DO NOT MERGE] - iOS: Integrate MMB into RNNativePlayground

lukewalczak:
- [x] - refactor to apply more RN Apps
- [x] - integrate MMB
- [x] - prepare native `Log in` to get `AccessToken` and pass it to `Bookings`
- [x] - fix navigation to work properly
- [x] - prepare standalone package for `MMB`


app/core/src/screens/MMBPackageWrapper.js
```diff
@@ -42,7 +47,8 @@ export default class MMBPackageWrapper extends React.Component<
 
   componentDidMount = () => {
     this.fetchToken();
-    this.props.navigation.setParams({ fetchToken: this.fetchToken });
+    this.props.navigation &&
+      this.props.navigation.setParams({ fetchToken: this.fetchToken });
```
grabbou: Let's clear this file and create an appropriate "StandalonePackage" file

grabbou: Should be `appRegistry` folder with root component and standalone packages.

tbergq: this file is not supposed to be a part of the native application. This is just a rapper to pass in to `ManageMyBookingPackage` what native should pass into the application

app/MMB/src/appRegistry/MMBStandalonePackage.js
```diff
@@ -0,0 +1,74 @@
...
+import NavigationStack from '../../src/navigation/NavigationStack';
+import RootComponent from './RootComponent';
+
+type Props = {|
+  dimensions: DimensionType,
+  onNavigationStateChange: () => void,
+  accessToken: string,
+  currency: string,
+|};
+
```
tbergq: please use read-only prop types

app/MMB/src/appRegistry/MMBStandalonePackage.js
```diff
@@ -0,0 +1,74 @@
...
+  };
+
+  componentDidMount = () => {
+    this.fetchToken();
+  };
+
+  fetchToken = async () => {
+    const accessToken = await AsyncStorage.getItem('mobile:MMB-Token');
+    this.setState({ accessToken });
+  };
```
tbergq: The token should be passed in as a prop from native code

app/MMB/src/appRegistry/MMBStandalonePackage.js
```diff
@@ -0,0 +1,74 @@
...
+
+  renderInnerComponent = () => {
+    if (!this.state.accessToken && !this.props.accessToken) {
+      return (
+        <View style={styles.loginMessageContainer}>
+          <Translation passThrough="You are not logged in, go to profile and log in" />
+        </View>
+      );
+    }
```
tbergq: This will be the responsibility of the native app to show this message.

app/MMB/src/appRegistry/MMBStandalonePackage.js
```diff
@@ -0,0 +1,74 @@
...
+    );
+  };
+
+  render = () => {
+    return (
+      <RootComponent
+        dimensions={this.props.dimensions}
+        accessToken={this.state.accessToken || this.props.accessToken}
+      >
+        {this.renderInnerComponent()}
```
tbergq: And then this would just be 
```
 <NavigationStack
   onNavigationStateChange={this.props.onNavigationStateChange}
 />
```

app/MMB/src/appRegistry/RootComponent.js
```diff
@@ -0,0 +1,27 @@
...
+import { AuthContext } from '@kiwicom/mobile-relay';
+import { Dimensions, type DimensionType } from '@kiwicom/mobile-shared';
+
+import BookingDetailContext from '../context/BookingDetailContext';
+
+type Props = {|
+  children: React.Node,
+  dimensions: DimensionType,
+  accessToken: string,
+|};
```
tbergq: read-only props please

packages/relay/src/AuthContext.js
```diff
@@ -37,6 +37,13 @@ class Provider extends React.Component<Props, State> {
     };
   }
 
+  static getDerivedStateFromProps(props: Props, state: State) {
+    return {
+      ...state,
+      accessToken: props.accessToken,
+    };
+  }
```
tbergq: Why is this needed, will the prop being passed from native ever change?

packages/shared/src/WithNativeNavigation.js
```diff
@@ -9,7 +9,7 @@ type NavigationState = {|
 
 type InjectedProps = {|
   onNavigationStateChange: () => void,
-  onBackClicked: () => void,
+  onBackClicked?: () => void,
```
ferrannp: Why is this optional now? What has changed?

ferrannp:
I see this code:

```objc
- (NSURL *)jsCodeLocation {
  // TODO: Change with hotels bundle
  return [NSURL URLWithString:@"http://localhost:8081/index.bundle?platform=ios&dev=true"];
}

- (NSURL *)jsCodeLocation {
  // TODO: Change with bookings bundle
  return [NSURL URLWithString:@"http://localhost:8081/index.bundle?platform=ios&dev=true"];
}
```

Does this mean we want to have two JS bundles instead of one ? Will we init two bridges? We will have two copies of React Native in each JS bundle? Even if we have two difference JS, we are adding all native dependencies (even the ones hotels do not need). How we handle that?

I am trying to thing if having more than one JS bundle is really the way to go. WDYT @grabbou ?

packages/shared/src/WithNativeNavigation.js
lukewalczak: Reverted. Added `FlowFixMe` till `onBackClicked` will not be implmented in Android app.

grabbou:
I think this is a question that Kiwi team has to answer themselves. I would myself go with a single bundle, because the additional KBs of Javascript are going to vanish when combined with native code along the way. One important trade-off of multiple bundles to understand is that separate bridges mean separate memory. That makes it hard to maintain user session or any other in-memory kind of data between multiple bridges. Keeping things simple usually delivers best performance.

We can always keep a single bundle and in the future, think about code splitting.

ferrannp:
> If this feature is in the release bundle during it's development, it will increase the final APK size which is quite sensitive parameter for Android users

Yeah I agree on this. I would also not merge right now PR that adds extra JS code and extra native dependencies to the release apk if the code is not used.

Anyway this PR is for iOS and we decided to keep it on hold till the apps need to use MMB feature.

On the other hand, I also agree that branching/QA here would be useful. Just create the version library for Android or the framework for iOS and test it in the real app though (not our pure RN example). CC @tbergq

tbergq:
> Anyway this PR is for iOS and we decided to keep it on hold till the apps need to use MMB feature.

So this means no increased size until it is released. 

For now, I see no need for branching since this workflow just creates more work. Maybe some time in the future it will make sense to change the workflow.

lukewalczak:
Closed (obsolete): The structure was changed along with the PR #1083.

PR #814, opened 2018-07-03
MMB: Add boarding pass for one way booking

tbergq:


app/MMB/src/scenes/tickets/boardingPasses/BoardingPassOneWay.js
```diff
@@ -0,0 +1,42 @@
...
+type Props = {|
+  +data: BookingType,
+|};
+
+const BoardingPassOneWay = (props: Props) => (
+  <FlightSegments
+    data={idx(props.data, _ => _.trip)}
+    icon={<TextIcon code="&#xe079;" style={styles.outboundIcon} />}
+    iconTitle={
+      <Translation id="mmb.boarding_passes.boarding_pass_return.departure" />
```
RobinCsl: Should there be a special `id` for OneWay? Or a more generic for all of the types?

tbergq: you are right, good catch

tbergq:
@RobinCsl updated

PR #827, opened 2018-07-06
Localization: Improve diffInDaysFunction

### LINKED ISSUE #765: Improve diffInDays function
see #764

tbergq:
closes #765


packages/localization/src/DateUtils.js
```diff
@@ -44,7 +44,11 @@ function DateUtils(rawDate: Date = getUTCNow()) {
  */
 DateUtils.diffInDays = (higherDate: Date, lowerDate: Date) => {
   const diffMs = higherDate - lowerDate;
-  return Math.ceil(diffMs / (1000 * 3600 * 24));
+  const oneDayInMs = 1000 * 60 * 60 * 24;
```
RobinCsl: Since it is a constant, I think it should be all caps.

packages/localization/src/DateUtils.js
```diff
@@ -44,7 +44,11 @@ function DateUtils(rawDate: Date = getUTCNow()) {
...
+
+  // Math.round (negativeNumber / anyNumber) === -0
+  // -0 is still 0 days, adding 0 will turn -0 into +0
+  return Math.round(diffMs / oneDayInMs) + 0;
```
RobinCsl: By rounding the result, `Math.round(0.5) === 1` is true. So I am not sure we should even round the result. It could make sense that diffInDays can return a non-integer value; it would just be a fraction of a day. Does that make sense?

tbergq: Ok, I will add changes as we agreed on slack

tbergq:
@RobinCsl 🚀

PR #833, opened 2018-07-09
MMB: Add fill travel document detail screen

tbergq:


packages/navigation/types/Navigation.js
```diff
@@ -40,7 +40,8 @@ type MMBStackNavigatorNames =
   | 'mmb.tickets.boarding_pass'
   | 'mmb.timeline.invoice'
   | 'mmb.explore'
-  | 'DetailScreen';
+  | 'DetailScreen'
+  | 'TravelDocumentScreen';
```
RobinCsl: Just out of curiosity: why do we have a different naming convention here?

tbergq: Not really sure 🤷‍♂️

RobinCsl: I believe it would be better to stick to one, to avoid confusions in the future. What do you think?

tbergq: Well, this dotted syntax is different somehow. They are actual screens on mobile, but not on ios

RobinCsl: Ok then. If you think that it makes sense to have this distinction, I'll approve it.

PR #836, opened 2018-07-09
Scrolling into view future event closest to today

### LINKED ISSUE #812: [Booking-Timeline] Open the timeline at relevant position

RobinCsl:
Fixes #812

### DIFF
--- a/app/MMB/src/components/ScrollIntoView.js
+++ b/app/MMB/src/components/ScrollIntoView.js
@@ -1,0 +1,60 @@
+// @flow
+
+import * as React from 'react';
+import { View, ScrollView } from 'react-native';
+
+type Props = {|
+  +shouldScrollIntoView: boolean,
+  +scrollToY?: (y: number) => void,
+  +children: React.Node,
+|};
+
+type OnLayoutEvent = {|
+  +nativeEvent: {
+    +layout: { +x: number, +y: number, +width: number, +height: number },
+  },
+|};
+
+export class ScrollIntoView extends React.Component<Props> {
+  onLayout = (e: OnLayoutEvent) => {
+    this.props.scrollToY && this.props.scrollToY(e.nativeEvent.layout.y);
+  };
+  render() {
+    return (
+      <View onLayout={this.props.shouldScrollIntoView && this.onLayout}>
+        {this.props.children}
+      </View>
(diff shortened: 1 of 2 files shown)


app/MMB/src/components/ScrollIntoView.js
```diff
@@ -0,0 +1,59 @@
...
+
+type OnLayoutEvent = {|
+  nativeEvent: {
+    layout: { x: number, y: number, width: number, height: number },
+  },
+|};
+
+export class ScrollIntoView extends React.Component<Props> {
+  static defaultProps = {
+    scrollToY: () => {},
```
tbergq: this prop is mandatory, why do we have defaultProp?

app/MMB/src/components/ScrollIntoView.js
```diff
@@ -0,0 +1,59 @@
...
+type Props = {|
+  +shouldScrollIntoView: boolean,
+  +scrollToY: (y: number) => void,
+  +children: React.Node,
+|};
+
+type OnLayoutEvent = {|
+  nativeEvent: {
+    layout: { x: number, y: number, width: number, height: number },
+  },
```
tbergq: could also be read-only

app/MMB/src/components/ScrollIntoView.js
```diff
@@ -0,0 +1,59 @@
...
+  render() {
+    return <View onLayout={this.onLayout}>{this.props.children}</View>;
+  }
+}
+
+type CustomScrollViewProps = {|
+  +children: React.Node,
+|};
+
+export class CustomScrollView extends React.Component<CustomScrollViewProps> {
```
tbergq: Custom in what way? This name does not tell me what this component offers

app/MMB/src/components/ScrollIntoView.js
RobinCsl: I get the following Flow error
```
Cannot create ScrollIntoView element because property scrollToY is missing in props [1] but exists in
Props [2].
```
if it is not present. This is because this prop is injected.

tbergq: Then I would set the prop to not be mandatory, since you actually create the component without it.

PR #841, opened 2018-07-10
MMB: Add passenger menu group to fill missing information

tbergq:


app/MMB/src/scenes/travelDocument/menuItem/PassengerMenuItem.js
```diff
@@ -0,0 +1,69 @@
...
+class PassengerMenuItem extends React.Component<Props> {
+  onPress = () => console.warn('TODO');
+
+  render = () => {
+    const date = idx(this.props.data, _ => _.travelDocument.expiration) || null;
+    const idNumber =
+      idx(this.props.data, _ => _.travelDocument.idNumber) || null;
+
+    return (
+      <Touchable onPress={this.onPress} disabled={idNumber !== null}>
```
RobinCsl: I don't think it should be disabled; if you look on the current app, it is possible to access the screen of passengers with filled information.

tbergq: Right, but it is actually a bug, I will send you a link on slack

PR #845, opened 2018-07-10
Adding missing animations on TripOverview in Booking Details

RobinCsl:
and removing visual feedback on Touchable

### DIFF
--- a/packages/shared/src/icons/TextIcon.js
+++ b/packages/shared/src/icons/TextIcon.js
@@ -30,19 +30,20 @@ const styleSheet = StyleSheet.create({
  *
  * We currently support only one font family: "spfont".
  */
-function TextIcon(props: Props) {
-  return (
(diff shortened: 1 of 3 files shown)


packages/shared/src/icons/TextIcon.js
```diff
@@ -30,19 +30,20 @@ const styleSheet = StyleSheet.create({
...
+      fontSize: 20,
+    },
+  };
+  render() {
+    return (
+      <Text style={[styleSheet.icon, this.props.style]}>
+        <Translation passThrough={this.props.code} />
+      </Text>
+    );
+  }
```
tbergq: Why is this change needed?

RobinCsl: This is to be able to use `Animated.createAnimatedComponent`; it was complaining that it only works with classes.

PR #846, opened 2018-07-10
MMB: Add boarding pass texts for future bookings

tbergq:


app/MMB/src/components/MissingInformation.js
```diff
@@ -16,14 +16,26 @@ import {
...
+  setIsMissingDocumentId: (isMissingDocumentId: boolean) => void,
 |};
 
-export class MissingInformation extends React.Component<Props> {
+export class MissingInformation extends React.Component<PropsWithContext> {
+  componentDidMount = () => {
+    this.props.setIsMissingDocumentId(this.isPassengerMissingId());
+  };
+
+  isPassengerMissingId = () => {
```
RobinCsl: Maybe, is**Some**PassengerMissingId since it's related to passengers, and not just a single passenger.

tbergq: 🚀

PR #851, opened 2018-07-11
MMB: Add form for filling in travel document data

tbergq:
TODO: Submitting and validation

### DIFF
--- a/app/MMB/src/scenes/travelDocument/menuItem/PassengerMenuItem.js
+++ b/app/MMB/src/scenes/travelDocument/menuItem/PassengerMenuItem.js
@@ -23,6 +23,13 @@ class PassengerMenuItem extends React.Component<Props> {
     this.props.navigation.navigate({
       routeName: 'TravelDocumentModalScreen',
       key: 'key-TravelDocumentModalScreen',
+      params: {
+        title: idx(this.props.data, _ => _.title) || '',
+        fullName: idx(this.props.data, _ => _.fullName) || '',
+        idNumber: idx(this.props.data, _ => _.travelDocument.idNumber) || null,
+        expiryDate:
+          idx(this.props.data, _ => _.travelDocument.expiration) || null,
+      },
     });
   };
 
@@ -55,6 +62,7 @@ export default createFragmentContainer(
   graphql`
     fragment PassengerMenuItem on Passenger {
       fullName
+      title
       travelDocument {
         idNumber
         expiration
--- a/packages/shared/src/forms/datePicker/DatePicker.ios.js
+++ b/packages/shared/src/forms/datePicker/DatePicker.ios.js
@@ -2,7 +2,11 @@
 
 import * as React from 'react';
 import { DatePickerIOS } from 'react-native';
-import { Translation, DeviceInfo } from '@kiwicom/mobile-localization';
+import {
+  Translation,
+  DeviceInfo,
+  DateUtils,
+} from '@kiwicom/mobile-localization';
 
 import DatePickerButton from './DatePickerButton';
 import BarPopup from '../../popup/BarPopup';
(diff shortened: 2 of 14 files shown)


app/MMB/src/scenes/travelDocument/menuItem/PassengerMenuItem.js
```diff
@@ -23,6 +23,15 @@ class PassengerMenuItem extends React.Component<Props> {
     this.props.navigation.navigate({
       routeName: 'TravelDocumentModalScreen',
       key: 'key-TravelDocumentModalScreen',
+      params: {
+        fullName: `${idx(this.props.data, _ => _.title) || ''}. ${idx(
```
RobinCsl: This is pedantic, but I think whether there should be a full stop ( . ) after the title should be decided by the locale currently used: https://english.stackexchange.com/questions/374899/which-is-correct-dr-or-dr

packages/shared/src/forms/datePicker/DatePicker.ios.js
```diff
@@ -22,7 +26,7 @@ export default class DatePicker extends React.Component<Props, State> {
   static getDerivedStateFromProps = ({ date }: Props) => ({ date });
 
   componentDidMount = () => {
-    this.setState({ date: this.props.date });
+    this.setState({ date: this.props.date || DateUtils.getUTCToday() });
```
RobinCsl: Why not set `DateUtils.getUTCToday()` as default value for prop `date`?

packages/shared/src/forms/datePicker/DatePickerButton.js
```diff
@@ -10,19 +14,26 @@ import Text from '../../Text';
...
       <View style={styles.date}>
         {props.iconComponent}
-        <Text style={styles.dateText}>
-          <Translation passThrough={DateFormatter(props.date).formatToDate()} />
-        </Text>
+        {props.date !== null && (
+          <Text style={styles.dateText}>
+            <Translation
+              passThrough={DateFormatter(props.date)[
+                props.formatFunction || 'formatToDate'
```
RobinCsl: Same question here: is it not cleaner to use default value for prop `formatFunction`?

app/MMB/src/scenes/travelDocument/menuItem/PassengerMenuItem.js
tbergq: We could maybe pass the entre title through translations 🤔

packages/shared/src/forms/datePicker/DatePicker.ios.js
tbergq: Because if it is called like `<DatePicker date={null} />` it has a value null, and default value just takes place of the prop is missing (`<DatePicker />`).

The the app crashes when you open the datepicker and date = null.

RobinCsl: Didn't think about that. Makes sense!

packages/shared/src/forms/datePicker/DatePickerButton.js
tbergq: Definitely, here I wanted to use `defaultProps`, however flow does not like it 🤷‍♂️

app/MMB/src/scenes/travelDocument/menuItem/PassengerMenuItem.js
RobinCsl: Maybe. Who is the translations expert here?

tbergq: I mean, it makes sense, now we just show what is passed through the graphql. But, if I put the existing app on Spanish, it shows me `Sr.` which is the spanish translation for `Mr.`

tbergq:
@RobinCsl 🚀

PR #852, opened 2018-07-11
iOS: Prepare two options to present hotels view

lukewalczak:


packages/shared/src/HeaderButton.js
```diff
@@ -35,7 +33,7 @@ export default class HeaderRightButton extends React.PureComponent<
   };
 
   render = () => {
-    const { onPress } = this.props;
+    const { onPress, left } = this.props;
```
ferrannp: Instead of a prop I would have `HeaderButton.Left` and `HeaderButton.Right`. Something similar to:

```js
HeaderButton.Right = ({ style, ...rest }) => ({
  <HeaderButton {...rest} style={[props.style, styles.right]} />
})
```

packages/shared/src/HeaderButton.js
```diff
@@ -72,4 +76,14 @@ const styles = StyleSheet.create({
       margin: 16,
     },
   },
+  innerLeftContainer: {
+    ios: {
+      marginStart: 10,
```
ferrannp: This is too much duplication, seems shorter to have a specific style. And then:

```js
style={[styles.innerContainer, styles.right]}
```

Where `right` contain the specifics overrides.

packages/shared/src/HeaderButton.js
```diff
@@ -10,16 +10,14 @@ type Props = {|
...
 
 type State = {|
   reference: React.ElementRef<typeof Touchable> | null,
 |};
 
-export default class HeaderRightButton extends React.PureComponent<
-  Props,
-  State,
-> {
+export default class HeaderButton extends React.PureComponent<Props, State> {
```
ferrannp: This component is a good candidate now for testing (even if Snapshots) for `.Left` and `.Right`.

app/hotels/src/navigation/allHotels/AllHotelsNavigationScreen.js
```diff
@@ -107,3 +123,16 @@ export default class AllHotelsNavigationScreen extends React.Component<Props> {
...
+const styles = StyleSheet.create({
+  closeButton: {
+    color: Color.brand,
+    fontWeight: '600',
+    android: {
+      fontSize: 18,
+    },
+    ios: {
+      fontSize: 17,
+    },
```
ferrannp: Let's make it shorter `fontSize: Platform.OS === 'android' ? 18 : 17`.

app/hotels/src/navigation/allHotels/AllHotelsNavigationScreen.js
```diff
@@ -39,12 +48,19 @@ export default class AllHotelsNavigationScreen extends React.Component<Props> {
...
+            <Text style={styles.closeButton}>
+              <Translation id="hotels.navigation.close.button" />
+            </Text>
+          </HeaderButton>
+        ) : (
+          <HeaderBackButton
+            tintColor={Color.brand}
+            onPress={props.onBackClicked}
+          />
+        ),
```
ferrannp: We will need `HOC` in the future for all that static stuff (MMB will have it too).

packages/shared/src/HeaderButton.js
lukewalczak: Done.

app/hotels/src/navigation/allHotels/AllHotelsNavigationScreen.js
lukewalczak: Maybe open an issue for that?

ferrannp: Yeah let's postpone it till we need it.

PR #859, opened 2018-07-12
Uniform interpretation of the booking status between flight list and …

RobinCsl:
…booking detail

### DIFF
--- a/app/MMB/src/scenes/list/cityImage/StatusBadge.js
+++ b/app/MMB/src/scenes/list/cityImage/StatusBadge.js
@@ -16,42 +16,87 @@ const StatusBadge = (props: Props) => {
   const status = idx(props.data, _ => _.status) || '';
   const isPastBooking = idx(props.data, _ => _.isPastBooking);
 
+  let renderStatus;
   switch (status) {
-    case 'REFUNDED':
-      return (
-        <AdaptableBadge
-          translation={<Translation id="mmb.booking_state.refunded" />}
-          style={styles.refundedBadge}
-          textStyle={styles.badgeText}
-        />
-      );
-    case 'CONFIRMED':
-      return (
-        <AdaptableBadge
-          translation={
-            <Translation
-              id={
-                isPastBooking
-                  ? 'mmb.booking_state.confirmed.travelled'
-                  : 'mmb.booking_state.confirmed'
-              }
-            />
-          }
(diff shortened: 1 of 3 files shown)


app/MMB/src/scenes/list/cityImage/StatusBadge.js
```diff
@@ -17,11 +17,19 @@ const StatusBadge = (props: Props) => {
   const isPastBooking = idx(props.data, _ => _.isPastBooking);
 
   switch (status) {
-    case 'REFUNDED':
+    case 'CANCELLED':
```
ferrannp: Overall I think this file can be much, much smaller. What do you think of an object like:

```js
const statuses = {
   // Map object
   CANCELLED: {
     translation: 'your-value-here'
     style: 'your-value-here'
   }
  // Other status
  ...
}
```

`textStyle` is the same for all. Then you can do.

```js
switch(status) {
 return (
    <AdaptableBadge
        translation={<Translation id={statuses[STATUS].translation} />}
        style={statuses[STATUS].style}
        textStyle={styles.badgeText}
    />
 )
}
```

So you only write the `AdaptableBadge` component once.

app/MMB/src/scenes/list/cityImage/StatusBadge.js
```diff
@@ -66,13 +104,36 @@ export default createFragmentContainer(
 
 const styles = StyleSheet.create({
   confirmBadge: {
-    backgroundColor: 'rgba(51, 189, 27, 0.6)',
+    backgroundColor: Color.green.normal,
+    opacity: 0.7,
```
ferrannp: We need to reuse style. If they all use `opacity: 0.7`, then we can have:

```js
style={[styles.badge, styles.someColor]}
```

Where:

```js
badge {
  opacity: 0.7,
}
```

Also you have a lot of styles that are exactly the same. Instead of per badge, you can have, `styles.black`, `styled.red` and so on... Then, from the other commend I made:

```js
const statuses = {
   // Map object
   CANCELLED: {
     translation: 'your-value-here'
     style: [styles.badge, styles.red]
   }
  // Other status
  ...
}
```

RobinCsl:
@ferrannp PR updated 🚀

ferrannp:
We need to use the `switch` because of the flow type `%future added value` added by `yarn relay` command.

PR #866, opened 2018-07-13
Adding EnterDetails event

RobinCsl:


app/MMB/src/scenes/timeline/events/EnterDetailsTimelineEvent.js
```diff
@@ -0,0 +1,80 @@
...
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    width: '100%',
+  },
+  button: {
+    flex: 1,
+    marginTop: 10,
+    marginBottom: 10,
```
tbergq: use `marginVertical` instead

PR #869, opened 2018-07-16
chore: Update to react-navigation 2.x.x.

### LINKED ISSUE #861: Update to react-navigation 2.x.x
If we want to take advantage of new react-navigation, options, performance and more we really need to bump from version 1.

ferrannp:
Closes #861.

Got two warnings (one ignored):

- https://reactnavigation.org/docs/en/common-mistakes.html <-- Ignored due our architecture with package and stack (to include in native app). We can review it later

- ~`react-navigation-props-mapper` gives a warning if used in stateless components (I removed it from MMB for now <-- I do not see is used there?). Question, can we remove it entirely? I do not see a huge benefit on using this dependency (we are using `navigate.state.params` in different places anyway). CC @tbergq~ Solved.

### DIFF
--- a/app/MMB/src/navigation/ListScreen.js
+++ b/app/MMB/src/navigation/ListScreen.js
@@ -11,7 +11,7 @@ export default function ListScreen() {
 }
 
 ListScreen.navigationOptions = () => ({
-  title: (
+  headerTitle: (
     <HeaderTitle>
       <Translation id="mmb.my_bookings.title.bookings" />
     </HeaderTitle>
--- a/app/MMB/src/navigation/NavigationStack.js
+++ b/app/MMB/src/navigation/NavigationStack.js
@@ -4,7 +4,6 @@ import {
   StackNavigator,
   StackNavigatorOptions,
 } from '@kiwicom/mobile-navigation';
-import { withMappedNavigationAndConfigProps as withMappedProps } from 'react-navigation-props-mapper';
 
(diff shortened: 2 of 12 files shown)


tbergq:
> I removed it from MMB for now <-- I do not see is used there

I think there is one screen that uses it, `TravelDocumentModalScreen`

>  I do not see a huge benefit on using this dependency

The main benefit, from my point of view, is that you can declare the prop types for the component, instead of passed props being sent as cryptic `navigation.state.params`

> we are using navigate.state.params in different places anyway

I think this is just in `static navigationOptions` when props are being passed from component to static method. 

> Question, can we remove it entirely?

If the pros are better than the cons(which I guess they are here) I am fine with removing it.

ferrannp:
Ok, so the problem is that we were calling:

```js
screen: MainStack,
```

And `MainStack` is not a component (screen) but another navigator. I solved that.

ferrannp:
@tbergq nice catch. Kinda breaking change. I fixed it. I also created https://github.com/kiwicom/mobile/issues/889.

mrtnzlml:
> I do not see a huge benefit on using this dependency...

It was my first and probably the only attempt to hide react-navigation implementation details and fix its not very nice design. My motivation is well described here: https://github.com/react-navigation/react-navigation/issues/935

ferrannp:
Thanks @mrtnzlml. We finally kept it 👍.

PR #891, opened 2018-07-18
Shared: Add cancellable promise to Price component

tbergq:


packages/shared/src/CancellablePromise.js
```diff
@@ -0,0 +1,25 @@
...
+export default (incomingPromise: Promise<*>) => {
+  let resolve;
+  let reject;
+
+  const promise = new Promise((originalResolve, originalReject) => {
+    resolve = originalResolve;
+    reject = originalReject;
+  });
+
+  incomingPromise.then(resolve).then(reject);
```
RobinCsl: I know this is just abstracting the original idea in the other file, but I don't fully understand the logic here. 
What is it useful for?

tbergq: basically, when you try to call setState on an unmounted component, you will get a yellow warning. It is also a memory leak. 

This is the suggested solution: 

https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html

RobinCsl: If I look at the solution, it looks like it should be
```
incomingPromise.then(resolve, reject)
```

tbergq: That would not work. 

You can see #647 and we decieded to go for this solution:

https://blog.bloomca.me/2017/12/04/how-to-cancel-your-promise.html#pure-promises

PR #892, opened 2018-07-18
iOS: block swiping back on modally presented screen

lukewalczak:
With that solution it's necessary to add `isModule: true` to `params` when declaring `modal` screen, eg:
```
    this.props.navigation.navigate({
      routeName: 'XYZ',
      key: 'key-XYZ',
      params: {
        isModal: true,
      },
    });
```

### DIFF
--- a/packages/navigation/types/Navigation.js
+++ b/packages/navigation/types/Navigation.js
@@ -77,3 +77,24 @@ export type Navigation = {
   goBack: (key?: string | null) => void,
   addListener: NavigationListener,
 };
+
+type Routes = Array<{|
+  key: string,
+  index: number,
+  isTransitioning: boolean,
+  routeName: string,
+|}>;
+
+export type Navigator = {
+  state: {
+    nav: {
+      key: string,
+      index: number,
+      isTransitioning: boolean,
+      params?: {|
+        isModal: boolean,
+      |},
+      routes: Routes,
+    },
+  },
+};
--- a/packages/navigation/types/Navigation.js
+++ b/packages/navigation/types/Navigation.js
@@ -78,23 +78,23 @@ export type Navigation = {
   addListener: NavigationListener,
 };
 
-type Routes = Array<{|
-  key: string,
-  index: number,
-  isTransitioning: boolean,
-  routeName: string,
+type Routes = $ReadOnlyArray<{|
+  +key: string,
+  +index: number,
+  +isTransitioning: boolean,
+  +routeName: string,
 |}>;
 
-export type Navigator = {
-  state: {
-    nav: {
-      key: string,
-      index: number,
-      isTransitioning: boolean,
-      params?: {|
+export type Navigator = {|
+  +state: {|
+    +nav: {|
+      +key: string,
+      +index: number,
+      +isTransitioning: boolean,
+      +params?: {|
         isModal: boolean,
       |},
-      routes: Routes,
-    },
-  },
-};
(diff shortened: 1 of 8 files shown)


packages/navigation/types/Navigation.js
```diff
@@ -70,3 +70,24 @@ export type Navigation = {
...
+      key: string,
+      index: number,
+      isTransitioning: boolean,
+      params?: {|
+        isModal: boolean,
+      |},
+      routes: Routes,
+    },
+  },
+};
```
tbergq: could you change this to

```
type Routes = $ReadOnlyArray<{|
  +key: string,
  +index: number,
  +isTransitioning: boolean,
  +routeName: string,
|}>;

export type Navigator = {|
  +state: {|
    +nav: {|
      +key: string,
      +index: number,
      +isTransitioning: boolean,
      +params?: {|
        isModal: boolean,
      |},
      +routes: Routes,
    |},
  |},
|};
```

packages/shared/src/WithNativeNavigation.js
```diff
@@ -10,6 +11,7 @@ type NavigationState = {|
 type InjectedProps = {|
   onNavigationStateChange: () => void,
   onBackClicked: () => void,
+  onNavigator: (ref: React.ElementRef<*>) => void,
```
tbergq: you can probably change all these to read-only as well

packages/shared/src/WithNativeNavigation.js
```diff
@@ -18,16 +20,31 @@ function withNativeNavigation<Props: {}>(
...
   return class WithNativeNavigation extends React.Component<*> {
     currentIndex: number;
+    navigator: NavigatorType;
 
     constructor(props) {
       super(props);
       this.currentIndex = 0;
     }
+
+    nestings = function nestings(nav) {
```
tbergq: could this be arrow function instead?

packages/shared/src/WithNativeNavigation.js
```diff
@@ -18,16 +20,31 @@ function withNativeNavigation<Props: {}>(
...
+
     onNavigationStateChange = (
       previousState: NavigationState,
       currentState: NavigationState,
     ) => {
-      if (currentState.index === 0) {
+      const nav = this.navigator.state.nav;
+
+      if (this.nestings(nav)) {
+        GestureController.disableGestures(moduleName);
```
tbergq: Just so I understand, this line is the magic that disables the swipe in the modal?

lukewalczak: Exactly! It disabling native gestures from the js side.

packages/shared/src/WithNativeNavigation.js
lukewalczak: I would prefer to stick with that approach, I think it's more optimal.

tbergq: > I think it's more optimal

Please elaborate 😄

lukewalczak: https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1

tbergq: I am still not convinced. We have chosen to go for arrow functions in this project, and we should stick to it.

lukewalczak:
Replaced with #894

Issue #898, opened 2018-07-19
Add flow-typed react-navigation instead of re-typing it


ferrannp:
`flow-typed install react-navigation@2.x.x`

PR #899, opened 2018-07-20
[WIP] feat: add flow-typed react-navigation

lukewalczak:
#Fixes: #898

### DIFF
--- a/app/MMB/src/navigation/DetailScreen.js
+++ b/app/MMB/src/navigation/DetailScreen.js
@@ -151,7 +151,6 @@ export default class DetailsScreen extends React.Component<Props, State> {
    * Do proper navigation transition on mobile (narrow) devices.
    */
   openMenuOnMobile = (key: string) => {
-    // $FlowExpectedError: Flow type for this navigation seems to be quite challenging. I am skipping it on purpose.
     this.props.navigation.navigate({
       routeName: key,
       key: `key-${key}`,
--- a/app/hotels/src/allHotels/AllHotels.js
+++ b/app/hotels/src/allHotels/AllHotels.js
@@ -129,7 +129,7 @@ export class AllHotels extends React.Component<PropsWithContext> {
 type Props = {|
   +currency: string,
   +coordinates: Coordinates | null,
-  +openSingleHotel: (searchParams: AvailableHotelSearchInput) => void,
+  +openSingleHotel: (searchParams: AvailableHotelSearchInput) => void | boolean,
   +openLocationPicker: (location: string | null) => void,
   +openGuestsModal: () => void,
   +checkin: ?Date,
--- a/packages/navigation/index.js
+++ b/packages/navigation/index.js
@@ -2,7 +2,7 @@
 
 import * as React from 'react';
 import { Platform, StatusBar } from 'react-native';
-import ReactNavigation from 'react-navigation';
+import { StackNavigator as MainStackNavigator } from 'react-navigation'; // eslint-disable-line no-restricted-imports
 import {
   Color,
   Device,
@@ -115,7 +115,7 @@ export const StackNavigator = (
     +headerMode?: 'none',
   |},
 ) => {
-  return ReactNavigation.StackNavigator(RouteConfigs, StackNavigatorConfig);
+  return MainStackNavigator.StackNavigator(RouteConfigs, StackNavigatorConfig);
 };
 
 export const StackNavigatorOptions = {
--- a/packages/navigation/types/Navigation.js
+++ b/packages/navigation/types/Navigation.js
@@ -1,6 +1,6 @@
 // @flow
 
-type NavigationStateParameters = Object;
+import type { NavigationScreenProp } from 'react-navigation';
 
 // FIXME: this is related to the packages and not navigation
 type CoreStackNavigatorRouteNames =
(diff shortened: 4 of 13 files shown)


packages/navigation/index.js
```diff
@@ -2,7 +2,7 @@
 
 import * as React from 'react';
 import { Platform, StatusBar } from 'react-native';
-import ReactNavigation from 'react-navigation';
+import { StackNavigator as MainStackNavigator } from 'react-navigation'; // eslint-disable-line no-restricted-imports
```
mrtnzlml: Why are you ignoring eslint rules here? I think it would be better to fix it instead.

packages/navigation/types/Navigation.js
```diff
@@ -55,25 +55,4 @@ export type RouteNames =
...
-    params?: NavigationStateParameters,
-  }) => void,
-  state: {
-    params: NavigationStateParameters,
-  },
-  setParams: (newParameters: NavigationStateParameters) => void,
-  goBack: (key?: string | null) => void,
-  addListener: NavigationListener,
-};
+export type Navigation = NavigationScreenProp<*>;
```
mrtnzlml: Please do not use this `*` type. This existential type has been deprecated.

packages/navigation/types/Navigation.js
```diff
@@ -55,25 +55,4 @@ export type RouteNames =
...
-  callback: (payload: Object) => void,
-) => { remove: () => void };
-
-/**
- * @see https://reactnavigation.org/docs/navigators/navigation-prop
- */
-export type Navigation = {
-  navigate: ({
-    routeName: RouteNames,
-    key: string, // should be unique
```
mrtnzlml: We used custom type on purpose so we are forced to use the `key` (and therefore avoid the double opening of each screen on double-tap). This is how I understood it at that time and I am afraid you just removed this check so we are not going to be able to prevent this bug.

app/MMB/src/navigation/DetailScreen.js
```diff
@@ -151,7 +151,6 @@ export default class DetailsScreen extends React.Component<Props, State> {
    * Do proper navigation transition on mobile (narrow) devices.
    */
   openMenuOnMobile = (key: string) => {
-    // $FlowExpectedError: Flow type for this navigation seems to be quite challenging. I am skipping it on purpose.
```
mrtnzlml: It was challenging because we had checked here for `RouteNames` so we couldn't use a different than existing screen name. This check is also gone now and I am not sure if it's a good idea (you are not getting errors even thou you should - see?). But if so then you should remove the `RouteNames` type as well.

packages/navigation/types/Navigation.js
ferrannp: Ok, first it will be good to add a comment saying about the `key` because we as new in the project, we thought you have custom types because the ones from `react-navigation 1.x.x` were not really great (broken all the time).

What do you think of:

```js
type Navigation = NavigationScreenProp<NavigationStateParameters> & { 
 key: string,
}
```

That is the only custom thing you need here, right? About `NavigationStateParameters`...  Right now in your codebase you have it as:

```js
type NavigationStateParameters = Object;
```

Ideally we would have:

```js
type Navigation<S> = NavigationScreenProp<S> & { 
 key: string,
}
```

But that would require components that use that type.

packages/navigation/types/Navigation.js
ferrannp: Should be or `NavigationStateParameters` or dynamic type as I explained in the other comment.

packages/navigation/index.js
ferrannp: I agree in not ignoring the rule. Let's investigate it. Moreover, in `react-navigation 2` they use `createStackNavigator` instead.

app/MMB/src/navigation/DetailScreen.js
ferrannp: Mmm I am not entirely sure why there was a flow error here? We were allowing to navigate to a `routeName` that does not exist? How? Was it working?

app/hotels/src/allHotels/AllHotels.js
```diff
@@ -129,7 +129,7 @@ export class AllHotels extends React.Component<PropsWithContext> {
 type Props = {|
   +currency: string,
   +coordinates: Coordinates | null,
-  +openSingleHotel: (searchParams: AvailableHotelSearchInput) => void,
+  +openSingleHotel: (searchParams: AvailableHotelSearchInput) => void | boolean,
```
mrtnzlml: Why exactly is it necessary to use `void | boolean`? It seems odd to me.

app/MMB/src/navigation/DetailScreen.js
mrtnzlml: > We were allowing to navigate to a routeName that does not exist? How?

Just the Flow type didn't exist because it's dynamic here. Otherwise, these pages exist. I don't remember correctly and I was probably just lazy but it works on every other place thanks to the `RouteNames` type.

packages/navigation/types/Navigation.js
mrtnzlml: > Should be or...

I think Github messed up your comment here? Did you mean _"Should be `<something here>` or"_? 😊

packages/navigation/types/Navigation.js
mrtnzlml: > first it will be good to add a comment saying about the `key`

Add it, please.

About your suggestion - I didn't try it but I get your point. It could work. Not sure if it disallows the string instead of the object though.

app/hotels/src/allHotels/AllHotels.js
mrtnzlml: I do not see the `void` type there.

packages/shared/src/buttons/Button.js
```diff
@@ -9,7 +9,7 @@ import type { StylePropType } from '../../index';
 
 type Props = {|
   +children: React.Node,
-  +onPress: () => void,
+  +onPress: () => *,
```
mrtnzlml: The same - existential type is deprecated.

app/hotels/src/allHotels/AllHotels.js
lukewalczak: I will remove `void`

PR #900, opened 2018-07-20
[Trip Services - Insurance] MoreInfo screen

RobinCsl:


app/MMB/src/scenes/tripServices/insurance/InsuranceTermsPdfScene.js
```diff
@@ -0,0 +1,23 @@
...
+import { PdfViewAndStore } from '@kiwicom/mobile-pdf';
+
+const InsuranceTermsPdfScene = () => {
+  const languageCode = DeviceInfo.getLanguage();
+  let url =
+    'https://s3-eu-west-1.amazonaws.com/skypicker-pdf/insurance/VPPCP-Kiwi.com_ENG-CZE.pdf';
+  if (languageCode === 'cs') {
+    url =
+      'https://s3-eu-west-1.amazonaws.com/skypicker-pdf/insurance/VPPCP-Kiwi.com_CZE-ENG.pdf';
+  }
```
tbergq: this pdf is just available in two languages?

app/MMB/src/scenes/tripServices/insurance/InsuranceTermsPdfScene.js
```diff
@@ -0,0 +1,23 @@
...
+  return (
+    <PdfViewAndStore
+      fileName={`/terms-and-conditions/insurance-${languageCode}.pdf`}
+      url={url}
+    />
```
tbergq: Not sure we want to store this pdf to the users device. We have a component also for viewing, where you can cache it.

app/MMB/src/scenes/tripServices/insurance/MoreInfoScene.js
```diff
@@ -0,0 +1,374 @@
+// @flow
+
+// <Translation passThrough=flow
```
tbergq: can you delete this comment

app/MMB/src/scenes/tripServices/insurance/MoreInfoScene.js
```diff
@@ -0,0 +1,374 @@
...
+  Text,
+  StyleSheet,
+  LinkButton,
+} from '@kiwicom/mobile-shared';
+
+type Props = {|
+  +navigation: NavigationType,
+|};
+
+const MoreInfoScene = (props: Props) => {
```
tbergq: This file is way to big, please extract this component to separate file

app/MMB/src/scenes/tripServices/insurance/MoreInfoScene.js
```diff
@@ -0,0 +1,374 @@
...
+  const rows = [
```
tbergq: This file is way to big, please extract these rows to a separate file

app/MMB/src/scenes/tripServices/insurance/MoreInfoScene.js
```diff
@@ -0,0 +1,374 @@
...
+        passThrough="20%"
+      />,
+      <Translation
+        key="co_insurance_causes_subsequent_services-2-20%"
+        passThrough="20%"
+      />,
+    ],
+  ];
+
+  function onPress() {
```
tbergq: this is just really a hack for not triggering the eslint rule that does not allow arrow functions in jsx. This function will also be re created on each re render

Please rather create a class.

app/MMB/src/scenes/tripServices/insurance/MoreInfoScene.js
```diff
@@ -0,0 +1,374 @@
...
+type RowProps = {|
+  +data: [
+    React.Element<typeof Translation>,
+    React.Element<typeof Translation | typeof TextIcon>,
+    React.Element<typeof Translation | typeof TextIcon>,
+  ],
+  +index: number,
+|};
+
+const Row = (props: RowProps) => {
```
tbergq: Also move this to separate file

app/MMB/src/scenes/tripServices/insurance/MoreInfoScene.js
```diff
@@ -0,0 +1,374 @@
...
+      </View>
+    </ScrollView>
+  );
+};
+
+export default MoreInfoScene;
+
+type RowProps = {|
+  +data: [
+    React.Element<typeof Translation>,
```
tbergq: What are these props in the array? Could you rather pass them as an object with some naming?

app/MMB/src/scenes/tripServices/insurance/MoreInfoScene.js
```diff
@@ -0,0 +1,374 @@
...
+  scrollView: {
+    backgroundColor: Color.white,
+    paddingStart: 6,
+    paddingEnd: 6,
+  },
+  rowWrapper: {
+    flexDirection: 'row',
+    alignItems: 'center',
+    paddingTop: 15,
+    paddingBottom: 15,
```
tbergq: `paddingVertical` is the same as paddingTop & paddingBottom

app/MMB/src/scenes/tripServices/insurance/MoreInfoScene.js
```diff
@@ -0,0 +1,374 @@
...
+      </View>
+    </View>
+  );
+};
+
+const styles = StyleSheet.create({
+  scrollView: {
+    backgroundColor: Color.white,
+    paddingStart: 6,
+    paddingEnd: 6,
```
tbergq: `paddingHorizontal` is the same as paddingStart & paddingEnd

app/MMB/src/scenes/tripServices/insurance/InsuranceTermsPdfScene.js
RobinCsl: That's what I found in the native repository, so I assume the answer is yes. I can ask again to the PM.

app/MMB/src/scenes/tripServices/insurance/InsuranceTermsPdfScene.js
RobinCsl: Ok, I will replace it.

app/MMB/src/scenes/tripServices/insurance/MoreInfoScene.js
RobinCsl: Good catch!

app/MMB/src/scenes/tripServices/insurance/InsuranceTermsPdfScene.js
tbergq: Anyways, maybe it is even better to move these urls to the graphql, if some url changes, or there are added more, we would have to change this on all clients. Better to change on graphql only

RobinCsl: That makes sense, I will do that.

RobinCsl: @tbergq https://github.com/kiwicom/graphql/pull/280

app/MMB/src/scenes/tripServices/insurance/InsuranceTermsPdfScene.js
```diff
@@ -0,0 +1,45 @@
...
+
+import type { InsuranceTermsPdfSceneQueryResponse as Props } from './__generated__/InsuranceTermsPdfSceneQuery.graphql';
+
+const renderPdfViewer = (props: Props) => {
+  const documents = idx(props, _ => _.allDocuments.edges) || [];
+  let document;
+  let uri = '';
+  if (documents.length >= 1) {
+    document = idx(documents[0], _ => _.node);
+  }
```
tbergq: this if is not needed, idx can handle this for you
`const document = idx(documents, _ => _[0].node)`

RobinCsl: Ah, cool!

app/MMB/src/scenes/tripServices/insurance/InsuranceTermsPdfScene.js
```diff
@@ -0,0 +1,42 @@
...
+import type { InsuranceTermsPdfSceneQueryResponse as Props } from './__generated__/InsuranceTermsPdfSceneQuery.graphql';
+
+const renderPdfViewer = (props: Props) => {
+  const documents = idx(props, _ => _.allDocuments.edges) || [];
+  let uri = '';
+  const document = idx(documents, _ => _[0].node);
+  if (document && document.__typename === 'InsuranceTerms') {
+    uri = document.url || '';
+  }
+  return <PdfViewer uri={uri} />;
```
tbergq: also I am a little bit unsure about your approach. If someone adds another document type, it is not sure insurance document will be the first one returned anymore

you should probably filter the list until you find typename InsuranceTerms

app/MMB/src/scenes/tripServices/insurance/moreInfoScene/MoreInfoSceneData.js
```diff
@@ -0,0 +1,174 @@
...
+      <Translation id="mmb.trip_services.insurance.selection.more_info.liability_health" />
+    ),
+    travel_basic: <Translation passThrough="—" />,
+    travel_plus: <Translation passThrough="30 000 €" />,
+  },
+  {
+    title: (
+      <Translation id="mmb.trip_services.insurance.selection.more_info.liability_property" />
+    ),
+    travel_basic: <Translation passThrough="—" />,
```
tbergq: also please use camelCase and not snake_case here

packages/shared/src/buttons/LinkButton.js
```diff
@@ -11,15 +12,17 @@ import Button from './Button';
...
   +title: TranslationType,
   +onPress: Function,
+  +titleStyle?: StylePropType,
 |};
 
 /**
  * Button that looks like a simple link.
  */
 export default function LinkButton(props: Props) {
+  const titleStyle = props.titleStyle || null;
```
tbergq: this line is not needed
`style={[styleSheet.text, titleStyle]}` works fine even if `props.titleStyle` is undefined

app/MMB/src/scenes/tripServices/insurance/InsuranceTermsPdfScene.js
RobinCsl: But since I use 
```
node {
                ... on InsuranceTerms {
                  url
                  __typename
                }
              }
```
there should only be documents of `InsuranceTerms` type in the response. And the version corresponding to the language is chosen by the server too. Maybe I should make the response less smart and let the app decide what it wants to do with the different versions in different languages?

PR #902, opened 2018-07-23
Profile: Allow saving custom token

tbergq:
Currently we only have login with email and password. I need to access bookings on an account with facebook login to test apple-wallet integration, so we need to be able to save a custom token.

### DIFF
--- a/app/profile/components/SaveCustomToken.js
+++ b/app/profile/components/SaveCustomToken.js
@@ -1,0 +1,55 @@
+// @flow strict
+
+import * as React from 'react';
+import {
+  Text,
+  TextInput,
+  Button,
+  StyleSheet,
+  Color,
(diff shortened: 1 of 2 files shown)


app/profile/components/SaveCustomToken.js
```diff
@@ -0,0 +1,55 @@
...
+
+  onPress = () => {
+    this.props.onPress(this.state.token);
+  };
+
+  render = () => {
+    return (
+      <React.Fragment>
+        <TextInput
+          placeholder={<Translation passThrough="Custom token" />}
```
RobinCsl: So since it's only for testing purposes, we don't need any translations as it won't stay in the app forever, right?

tbergq: Exactly

PR #904, opened 2018-07-23
fix: Nested routes object is optional

ferrannp:


packages/shared/src/WithNativeNavigation.js
```diff
@@ -33,9 +33,11 @@ function withNativeNavigation<Props: {}>(
     ) => {
       // Native gesture should only be enabled in the very first screen
       // Also, there should not be modal (like first screen + modal)
+
       const isDisabled =
         currentState.routes.length > 1 ||
-        currentState.routes[0].routes.length > 1;
+        (currentState.routes[0].routes &&
+          currentState.routes[0].routes.length > 1);
```
tbergq: could you use `idx` for this

PR #909, opened 2018-07-24
Navigation: Export withNavigation HOC from navigation package

tbergq:


tbergq:
Now we import `withNavigation` from `react-navigation` and type `NavigationType` from `@kiwicom/mobile-navigation`. It reduces the number of imports we have to do. And also it is what we already do with e.g. exporting `graphql` and `createFragmentContainer` from `mobile-relay` instead.

PR #915, opened 2018-07-24
fix: Back button on Android after going to a nested screen (RN v2)

ferrannp:


packages/shared/src/WithNativeNavigation.js
```diff
@@ -4,8 +4,9 @@ import * as React from 'react';
 import { GestureController } from '@kiwicom/mobile-shared';
 
 type NavigationState = {|
+  +index: number,
   +routes: $ReadOnlyArray<{
-    +routes?: $ReadOnlyArray<mixed>,
+    +index: number,
```
tbergq: even this could be exact type

PR #918, opened 2018-07-25
MMB: Prepare data for apple wallet screen

tbergq:
I considered two approaches to get data for the apple wallet screen. 

A: Use query renderer
• I would have to pass legId and passengerId the new screen
• Query the booking for all legs, loop all legs to find my leg, loop pkpasses to find my passenger
• This does not sound optimal

B: The chosen approach. 
• Prepare data in context and set the selected one when clicking on the passenger
• I would then be able to get all needed data on the apple wallet screen from the context through the selectedSegment. 
• I do have to prop drill legId down to the click handler, but I would have had to do that with approach A also. 

C: Could be an approach I did not consider, so please let me know if you see another way 😄

### DIFF
--- a/app/MMB/src/context/WalletContext.js
+++ b/app/MMB/src/context/WalletContext.js
@@ -1,0 +1,103 @@
+// @flow strict
+
+import * as React from 'react';
+
+const defaultState = {
+  segments: [],
+  selectedSegment: null,
+  actions: {
+    addSegment: () => {},
+    addPkpassData: () => {},
+    setSelectedSegment: () => {},
+  },
+};
+const { Consumer, Provider: ContextProvider } = React.createContext({
+  ...defaultState,
+});
+
+type Props = {|
+  +children: React.Node,
+|};
+
+export type Segment = {|
+  +id: string,
+  +airlineLogoUrl: string,
+  +flightDate: Date | null,
+  +passengerName?: string,
+  +pkpassUrl?: string,
+|};
+
+type State = {|
+  segments: $ReadOnlyArray<Segment>,
+  selectedSegment: Segment | null,
+  +actions: {|
+    +addSegment: (segment: Segment) => void,
+    +setSelectedSegment: (segmentId: string | null) => void,
+    +addPkpassData: (
+      passengerName: string,
+      pkpassUrl: string,
+      segmentId: string,
+    ) => void,
+  |},
+|};
+
+class Provider extends React.Component<Props, State> {
+  constructor(props: Props) {
+    super(props);
+
+    this.state = {
+      segments: [],
+      selectedSegment: null,
+      actions: {
+        addSegment: this.addSegment,
+        addPkpassData: this.addPkpassData,
+        setSelectedSegment: this.setSelectedSegment,
+      },
+    };
+  }
+
+  addSegment = (segment: Segment) => {
+    this.setState(state => ({
+      segments: [...state.segments, segment],
+    }));
+  };
+
+  addPkpassData = (
+    passengerName: string,
+    pkpassUrl: string,
+    segmentId: string,
+  ) => {
+    this.setState(
+      state => ({
+        segments: state.segments.map(segment => {
+          if (segment.id === segmentId) {
+            return {
+              ...segment,
+              passengerName,
+              pkpassUrl,
+            };
+          }
+          return segment;
+        }),
+      }),
+      () => {
+        this.setSelectedSegment(segmentId);
+      },
+    );
+  };
+
+  setSelectedSegment = (segmentId: string | null) => {
+    this.setState(state => ({
+      selectedSegment:
+        segmentId === null
+          ? null
+          : state.segments.find(segment => segment.id === segmentId),
+    }));
+  };
+
+  render = () => (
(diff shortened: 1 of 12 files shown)


app/MMB/src/scenes/tickets/boardingPasses/appleWallet/AppleWalletPassenger.js
```diff
@@ -9,14 +9,18 @@ import { withNavigation } from 'react-navigation';
...
+  ...Props,
+  addPkpassData: (passengerName: string, pkpassUrl: string, id: string) => void,
 |};
 
-class AppleWalletPassenger extends React.Component<Props> {
+class AppleWalletPassenger extends React.Component<PropsWithContext> {
   onPress = () => {
+    const name = idx(this.props.data, _ => _.passenger.fullName) || '';
+    const url = idx(this.props.data, _ => _.url) || '';
+    this.props.addPkpassData(name, url, this.props.legId || '');
```
RobinCsl: This won't do anything if `this.props.legId` is `null` or `undefined`, right? So maybe I would add a conditional instead of some default empty string value because there is no reason to call that function in that case. Similarly, you could add the variable creations after checking for legId.

app/MMB/src/scenes/tickets/boardingPasses/appleWallet/AppleWallet.js
```diff
@@ -12,6 +12,7 @@ import type { AppleWallet as Pkpasses } from './__generated__/AppleWallet.graphq
 
 type Props = {|
   +data: Pkpasses,
+  +legId: ?string,
```
RobinCsl: Shall we start renaming these with `segmentId`? Since you are already using the (correct) word _segment_ for your context.

app/MMB/src/scenes/tickets/boardingPasses/FlightFromTo.js
```diff
@@ -10,57 +10,98 @@ import idx from 'idx';
...
-          <Text style={styles.cityText}>
-            <Translation
-              passThrough={idx(props.data, _ => _.departure.airport.city.name)}
-            />
+export class FlightFromTo extends React.Component<PropsWithContext> {
+  componentDidMount = () => {
+    const id = idx(this.props.data, _ => _.id) || '';
+    const airlineLogoUrl = idx(this.props.data, _ => _.airline.logoUrl) || '';
+    const flightDate = idx(this.props.data, _ => _.departure.localTime) || null;
+    this.props.addSegment({
```
RobinCsl: Maybe `addSegmentToContext` to make it clear that we are doing something outside of this component?

RobinCsl:
Well, interestingly, this is the same kind of problem I am dealing with at the moment: how to pass data correctly when navigating from one screen to another when this deals with one item of a list... 
I think the approach with the context is quite good because it really is about the context and all the information is managed by one state store. 
If we had had some global store, then it could have been slightly easier as we would just need to connect to that store and all the information would be available. I guess people are using the Relay Store as a replacement for Redux. SO maybe it would be possible to do something similar? My 2 cents.

One further comment, more specific to the code: it does not look like you are using `selectedSegment` anywhere yet. Is it by design and will be used later?

app/MMB/src/scenes/tickets/boardingPasses/FlightFromTo.js
tbergq: I think since it comes from props, it should be clear that it is something passed to the component, also since it is passed from the context, ain't it clear enough?

If not, I could change it.

app/MMB/src/scenes/tickets/boardingPasses/appleWallet/AppleWalletPassenger.js
tbergq: yes that makes sense

app/MMB/src/scenes/tickets/boardingPasses/appleWallet/AppleWallet.js
tbergq: Yes, this is already a bit confusing, since we fetch it like `leg.id` but we use it like segmentId later 🤔 

I will rename it.

tbergq:
>  Is it by design and will be used later

Yes, when I navigate to `AppleWalletScreen` I will use the `selectedSegment` that way, I don't need to pass any parameters, or do any queries on that screen.

tbergq:
@RobinCsl 🚀

app/MMB/src/scenes/tickets/boardingPasses/FlightFromTo.js
RobinCsl: I guess you're right, no need to add that.

app/MMB/src/context/WalletContext.js
```diff
@@ -0,0 +1,99 @@
...
+    };
+  }
+
+  addSegment = (segment: Segment) => {
+    this.setState(state => ({
+      segments: [...state.segments, segment],
+    }));
+  };
+
+  addPkpassData = (passengerName: string, pkpassUrl: string, id: string) => {
```
RobinCsl: Maybe I would add a precision that it must be a segment id, so I would replace `id` by `segmentId`.

PR #920, opened 2018-07-25
MMB: Travel information form validations

tbergq:


app/MMB/src/navigation/TravelDocumentModalScreen.js
```diff
@@ -63,24 +76,70 @@ export default class TravelDocumentModalScreen extends React.Component<
...
+  };
+
   onIdNumberChange = (idNumber: string) => {
-    this.setState({ idNumber });
+    this.setState(
+      state => ({
+        idNumber,
+        error: {
+          ...state.error,
+          idNumber: idNumber.length < 5 && idNumber.length !== 0,
```
RobinCsl: I believe it would be clearer to write: 
```
idNumber.length > 0 && idNumber.length < 5
```

RobinCsl: Also, is there a place where we would centralize these form-checking conditions?

app/MMB/src/navigation/TravelDocumentModalScreen.js
```diff
@@ -63,24 +76,70 @@ export default class TravelDocumentModalScreen extends React.Component<
...
+  };
+
+  onSave = () => {
+    // TODO: Send mutation
+    this.props.navigation.goBack();
+  };
+
+  validate = () => {
+    const hasError =
+      this.state.idNumber.length < 5 || this.state.error.expiryDate;
```
RobinCsl: And here you are using a different conditional for the idNumber

RobinCsl: You probably wanted `this.state.error.idNumber`?

app/MMB/src/navigation/TravelDocumentModalScreen.js
```diff
@@ -158,4 +229,13 @@ const styles = StyleSheet.create({
   datePickerContainer: {
     flex: 1,
   },
+  error: {
+    alignSelf: 'flex-end',
+  },
+  noExpiry: {
+    color: '#DFE4EA',
```
RobinCsl: Shall we add this color to the `Color.js`?

app/MMB/src/navigation/__tests__/TravelDocumentModalScreen.test.js
```diff
@@ -0,0 +1,84 @@
...
+  it('should disable header button correctly', () => {
+    const instance = getInstance();
+    jest.resetAllMocks();
+
+    instance.onIdNumberChange('12345');
+    expect(navigation.setParams).toHaveBeenCalledWith({ disabled: true });
+    expect(navigation.setParams).toHaveBeenCalledTimes(1);
+    jest.resetAllMocks();
+
+    instance.onDateChange(new Date());
```
RobinCsl: Maybe we should check that the expiry date of the document is later than the last leg of the journey?

RobinCsl: Hum, I mean, segment 😄

app/MMB/src/navigation/TravelDocumentModalScreen.js
tbergq: yes, thank you, that looks better

packages/shared/src/forms/datePicker/DatePicker.ios.js
```diff
@@ -82,7 +90,7 @@ export default class DatePicker extends React.Component<Props, State> {
             {...rest}
             minimumDate={minDate}
             maximumDate={maxDate}
-            date={this.state.date}
+            date={this.state.date || new Date()}
```
RobinCsl: Same comment; are these default values necessary?

app/MMB/src/navigation/TravelDocumentModalScreen.js
tbergq: Actually not, but it should be commented. If idNumber.length === 0. I do not want to show error message below the input, but I want to disable header button

RobinCsl: That makes sense. 👍

app/MMB/src/navigation/TravelDocumentModalScreen.js
tbergq: 🤷‍♂️ Colors are currently a mess from the designers from my POV. 

This color is for now only used one place. If it should be reused somehow, we find a name for it and add it to `Color.js`

app/MMB/src/navigation/__tests__/TravelDocumentModalScreen.test.js
tbergq: I got it as a TODO somewhere, but yes, you are right

app/MMB/src/navigation/TravelDocumentModalScreen.js
RobinCsl: I disagree a bit: if you are using a new color, then you must be diligent enough to check if it already exists in the codebase, add it to the `Color.js` file and then reflect changes to the other which has probably nothing to do with your current task.
As opposed to just checking for colors in `Color.js`.

packages/shared/src/forms/datePicker/DatePicker.ios.js
tbergq: Yes, if this.state.date is null, then DatePickerIOS will crash the entire app. Which is what happend on my demo on monday 😞 

And yes, it should be `DateUtils.getUTCToday()`

app/MMB/src/navigation/TravelDocumentModalScreen.js
tbergq: If we search for `'#` in our project, we have 77 results in 22 files. 

Probably room for improvements, because a lot of them are just white and black, and color codes I think we already have in colors. 

But ok, I will move this color to the file 😖

tbergq:
@RobinCsl 🚀

app/MMB/src/navigation/TravelDocumentModalScreen.js
RobinCsl: I will do this cleaning right now then.

packages/shared/src/forms/datePicker/DatePicker.ios.js
```diff
@@ -36,7 +40,9 @@ export default class DatePicker extends React.Component<Props, State> {
   };
 
   onSave = () => {
-    this.props.onDateChange(this.state.date);
+    // if this.state.date is null, fallback sent to DatepickerIOS is new Date()
```
RobinCsl: These comments aren't completely accurate anymore since we are using `DateUtils.getUTCToday()`, so maybe just correct this.

PR #921, opened 2018-07-25
Refactor Header buttons to make them consistent on Android and iOS

### LINKED ISSUE #913: Modal close button should be an "X" on Android
Check Location modal and Guests modal.

ferrannp:
Fixes #913.

- [x] Fix Location and Guests modals
- [x] Remove `HeaderButton` duplication from `shared` and `navigation`

Before:

from

After:

to


app/MMB/src/navigation/TravelDocumentModalScreen.js
```diff
@@ -53,19 +54,30 @@ export default class TravelDocumentModalScreen extends React.Component<
     }
     return {
       headerLeft: (
-        <HeaderButton onPress={goBack}>
-          <Translation id="mmb.missing_informaiton.travel_document_modal_screen.close" />
-        </HeaderButton>
+        <HeaderButton.Left onPress={goBack}>
+          {Platform.OS === 'android' ? (
+            <Icon name="close" size={24} color={Color.brand} />
+          ) : (
```
tbergq: Could this somehow be done in the component instead, so we don't have to do this with all header buttons?

app/MMB/src/navigation/TravelDocumentModalScreen.js
```diff
@@ -53,19 +54,30 @@ export default class TravelDocumentModalScreen extends React.Component<
...
       ),
       headerRight: (
-        <HeaderButton onPress={goBack}>
-          <Translation id="mmb.missing_informaiton.travel_document_modal_screen.save" />
-        </HeaderButton>
+        <HeaderButton.Right onPress={goBack}>
+          <HeaderButton.Text>
+            <Translation
+              id="mmb.missing_informaiton.travel_document_modal_screen.save"
+              textTransform={Platform.OS === 'android' ? 'uppercase' : ''}
```
tbergq: also nice to do this in the HeaderButton component

packages/navigation/src/HeaderButton.js
```diff
@@ -19,6 +30,11 @@ type State = {|
   reference: React.ElementRef<typeof Touchable> | null,
 |};
 
+type CloseModalProps = {
+  onPress: () => void,
+  textId?: TranslationKeysType,
```
tbergq: please make exact and read-only

packages/localization/src/transformations/CaseTransform.js
```diff
@@ -1,6 +1,6 @@
 // @flow strict
 
-export type SupportedTransformationsType = 'lowercase' | 'uppercase';
+export type SupportedTransformationsType = 'lowercase' | 'uppercase' | '';
```
tbergq: I don't think we should do this, we have these two types for a reason

packages/navigation/src/HeaderButton.js
```diff
@@ -0,0 +1,164 @@
...
+  }: {|
+    +children: TranslationType,
+    +disabled?: boolean,
+  |}) => (
+    <Text
+      style={[styles.headerButtonText, disabled ? styles.disabledButton : null]}
+    >
+      {React.cloneElement(children, {
+        textTransform: Platform.OS === 'android' ? 'uppercase' : '',
+      })}
```
tbergq:
```
Platform.Select({
ios: children,
android: React.cloneElement(children, ...)
})
```
might be better?

packages/navigation/src/HeaderButton.js
```diff
@@ -0,0 +1,164 @@
...
+    </Text>
+  );
+
+  static CloseModal = ({ onPress, textId }: CloseModalProps) => (
+    <HeaderButton.Left onPress={onPress}>
+      {Platform.OS === 'android' ? (
+        <Icon name="close" size={24} color={Color.brand} />
+      ) : (
+        <HeaderButton.Text>
+          <Translation id={textId || 'shared.button.close'} />
```
tbergq: I think it is better to pass translations than the id, so 

```
{textId || <Translation id="shared.button.close" />
```

packages/navigation/src/HeaderButton.js
```diff
@@ -0,0 +1,164 @@
...
+        onLongPress={this.onLongPress}
+        ref={this.storeReference}
+        disabled={disabled}
+      >
+        <View style={styles.container}>
+          <View style={[styles.innerContainer, style]}>
+            {this.props.children.type === HeaderButton.Text
+              ? // $FlowExpectedError: flow is having a hard time here
+                React.cloneElement(this.props.children, { disabled })
+              : this.props.children}
```
tbergq: I am also having a hard time here, what and why is the text disabled?

ferrannp: So `Button` can be `disabled` but you also need to pass it to `Text` because then it has a special style (disabled color right).

packages/navigation/src/HeaderButton.js
ferrannp: I am not sure I understand this one, `{textId || <Translation id="shared.button.close" />` this won't work because `textId` is something like "hotels.button.cancel` so it needs to be used with `Translation` right?

tbergq: Right, I am sorry, I explained my self badly

```
// If close modal is like:
static CloseModal = ({ onPress, text }: {|+onPress: () => void, +text?: TranslationType |}) {
if (text === undefined) return <Translation id="shared.button.close" />
return text;
}

// Then I can use it like
<CloseModal onPress={() => {}} text={<Translation id="mmb.lol.lol" />} />

```

packages/navigation/src/HeaderButton.js
tbergq: Right, I understand

PR #924, opened 2018-07-26
Hotels: Add ancillaryPurchased logging event when hotel is paid for

tbergq:
closes #924

### DIFF
--- a/packages/shared/src/WebView.js
+++ b/packages/shared/src/WebView.js
@@ -5,13 +5,14 @@ import { WebView as NativeWebView } from 'react-native';
 import { Translation } from '@kiwicom/mobile-localization';
 
 import GeneralError from './errors/GeneralError';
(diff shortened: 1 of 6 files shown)


packages/shared/src/WebView.js
```diff
@@ -5,13 +5,14 @@ import { WebView as NativeWebView } from 'react-native';
 import { Translation } from '@kiwicom/mobile-localization';
 
 import GeneralError from './errors/GeneralError';
+import type { WebViewStateChangeEvent } from '../types/Events';
 
-type Props = {
-  // not exact - may contain additional properties (extend as needed)
```
RobinCsl: This seemed to have been not exact on purpose; has this changed now?

tbergq: We had more components like this, but they are slowly disappearing. It is better to define what we can pass in, and expand when needed. In this case, we were only using source.uri as prop anyways.

PR #925, opened 2018-07-26
MMB: Add passenger name, flight date and info text to wallet screen

tbergq:


app/MMB/src/scenes/appleWallet/Passenger.js
```diff
@@ -0,0 +1,25 @@
...
+  return (
+    <WalletContext.Consumer>
+      {({ selectedSegment }) => {
+        const walletData = selectedSegment || {};
+        return (
+          <WalletText
+            label={
+              <Translation id="mmb.apple_wallet.apple_wallet_scene.passenger" />
+            }
+            text={<Translation passThrough={walletData.passengerName} />}
```
RobinCsl: This will render undefined if selectedSegment is falsy, right?

app/MMB/src/scenes/appleWallet/DateAndTime.js
```diff
@@ -0,0 +1,57 @@
...
+
+import WalletText from './WalletText';
+import WalletContext from '../../context/WalletContext';
+
+export default function DateAndTime() {
+  return (
+    <WalletContext.Consumer>
+      {({ selectedSegment }) => {
+        const walletData = selectedSegment || {};
+        const date = new Date(walletData.flightDate || 0);
```
RobinCsl: Maybe you should not render the date if it is missing, instead of printing 1970.

app/MMB/src/scenes/appleWallet/Passenger.js
tbergq: yes, that is correct, I should fix that

tbergq:
@RobinCsl added some more null/undefined checking 🚀

PR #926, opened 2018-07-26
MMB: Add update passenger mutation

tbergq:


app/MMB/src/navigation/TravelDocumentModalScreen.js
```diff
@@ -132,8 +139,29 @@ export default class TravelDocumentModalScreen extends React.Component<
...
+                ? null
+                : DateFormatter(this.state.expiryDate).formatForMachine(),
+            documentNumber: this.state.idNumber,
+          },
+        ],
+      },
+      () => {
+        this.props.navigation.goBack();
+        this.setState({ isSubmitting: false });
+        this.props.navigation.setParams({ disabled: true });
```
RobinCsl: Should this still be disabled?

tbergq: 🙈

tbergq: 🚀

PR #930, opened 2018-07-27
MMB: Refactor TravelDocumentModalScreen

tbergq:
Basically moving state to a context, and create smaller components

### DIFF
--- a/app/MMB/src/navigation/__tests__/TravelDocumentModalScreen.test.js
+++ b/app/MMB/src/navigation/__tests__/TravelDocumentModalScreen.test.js
@@ -1,84 +1,94 @@
 // @flow strict
 
-import * as React from 'react';
-import renderer from 'react-test-renderer';
-
-import TravelDocumentModalScreen from '../TravelDocumentModalScreen';
-
-jest.mock('../../components/TitleTranslation.js', () => () => null);
-jest.mock('DatePickerIOS');
+import { TravelDocumentModalScreen } from '../TravelDocumentModalScreen';
 
 const navigation = {
   setParams: jest.fn(),
 };
 
-const getInstance = () =>
-  renderer
-    .create(
-      // $FlowExpectedError: All props not needed for this test
-      <TravelDocumentModalScreen expiryDate={null} navigation={navigation} />,
-    )
-    .getInstance();
-
 afterEach(() => {
   jest.resetAllMocks();
 });
 
+const getComponent = props =>
+  // $FlowExpectedError: Passing just props needed to run test
+  new TravelDocumentModalScreen({
+    ...props,
+    navigation,
+  });
+
 describe('TravelDocumentModalScreen', () => {
-  it('should validate onIdNumberChange correctly', () => {
-    const instance = getInstance();
-    expect(instance.state.error.idNumber).toBe(false);
+  it('should call navigation setParams when a idNumber changes', () => {
+    const Component = getComponent({
+      idNumber: 'a',
+    });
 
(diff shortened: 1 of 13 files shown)


app/MMB/src/navigation/__tests__/TravelDocumentModalScreen.test.js
```diff
@@ -1,84 +1,94 @@
...
+  it('should enable save button if idNumber > 5 && expiryDate is set', () => {
+    const Component = getComponent({
+      idNumber: '123456',
+      expiryDate: new Date(),
+    });
+
+    // $FlowExpectedError: Passing just props needed to run test
+    Component.componentDidUpdate({
+      idNumber: '123456',
+      expiryDate: null,
```
RobinCsl: So you are passing a null expiryDate and you want the Save button not to be disabled? I would have believed it should be disabled in that case, since noExpiry is not set to `true`.

tbergq: I get that this is a bit confusing, but it is actually previousProps that are being passed into componentDidUpdate, so you kind of have to read this backwards.

app/MMB/src/scenes/travelDocument/form/TravelDocumentForm.js
```diff
@@ -0,0 +1,49 @@
...
+  +fullName: string,
+  +title: string,
+|};
+
+export default function TravelDocumentForm(props: Props) {
+  return (
+    <React.Fragment>
+      <View style={styles.row}>
+        <TextIcon code="w" style={styles.icon} />
+        <TitleTranslation title={props.title} />
```
RobinCsl: I think this could be improved further, as you don't know that all languages will write `Mr. Bla`; it might very well be `Bla, Mr.`. So I think we should do: 
'title_translation_mr': 'Mr. __name__', and then use
```
<Translation id="title_translation_mr" values={{name: 'John Smith'}} />
```

tbergq: That might be true, but I would say that is the scope of another PR 😄

packages/relay/src/AuthContext.js
```diff
@@ -47,3 +47,24 @@ class Provider extends React.Component<Props, State> {
...
+      {({ accessToken, actions: { setAccessToken } }) => (
+        <Component
+          {...props}
+          accessToken={accessToken}
+          setAccessToken={setAccessToken}
+        />
+      )}
+    </Consumer>
+  );
+  // $FlowExpectedError: We need to pass on the navigationOptions if any, flow does not know about it, but a react component might have it
```
RobinCsl: Maybe here that's also because it is written as a function, and not as a class?

tbergq: No, it is the same in `WithStorage` HOC which is a class.

app/MMB/src/navigation/__tests__/TravelDocumentModalScreen.test.js
RobinCsl: Oh, right. That makes sense.

app/MMB/src/scenes/travelDocument/form/TravelDocumentForm.js
RobinCsl: Absolutely

tbergq: I will create an issue for it

PR #931, opened 2018-07-27
Improving the timeline: day banner is sticky

RobinCsl:
@tbergq This is sort of working, but it is sketchy.
I believe that the timeout waiting times will be dependent on the number of elements of the list... So this is still WIP, but if you could check out the branch and see for yourself the current behaviour and give me your feedback, that would be great. 
Also, if you can see some obvious improvements, do let me know.


app/MMB/src/components/sectionListScrollTo/OnLayout.js
```diff
@@ -0,0 +1,27 @@
+// @flow
+
+import * as React from 'react';
+import { View } from 'react-native';
+import type { OnLayout as OnLayoutType } from '@kiwicom/mobile-shared';
+
+type State = {| height: number |};
+type Props = {| +children: State => React.Node |};
+
+export default class OnLayout extends React.Component<Props, State> {
```
tbergq: We already have this component, it is Exported as Dimensions from the shared package

app/MMB/src/components/sectionListScrollTo/SectionListScrollTo.js
```diff
@@ -0,0 +1,240 @@
...
+    height: 0,
+    rendered: false,
+  }));
+
+const getLayoutInfoLength = (data: Section[]) =>
+  data.reduce((acc, section) => acc + 1 + section.data.length + 1, 0);
+
+export default class SectionListScrollTo extends React.Component<Props> {
+  layoutInfo: ItemLayoutInfo[] = generateDefaultLayoutInfo(
+    getLayoutInfoLength(this.props.data),
```
tbergq: if you are passing the method props, it should probably be a class method instead

app/MMB/src/components/sectionListScrollTo/SectionListScrollTo.js
```diff
@@ -0,0 +1,240 @@
...
+  );
+
+  renderSectionFooter = ({ section }: { section: Section }) => {
+    this.addInfo('footer', 0, section);
+    return null;
+  };
+
+  keyExtractor = (item: React.Node[], index: number) => index;
+
+  componentDidMount() {
```
tbergq: this should be the first method in the class

app/MMB/src/components/sectionListScrollTo/SectionListScrollTo.js
```diff
@@ -0,0 +1,240 @@
...
+    this.verifyAndScroll;
+  }
+
+  componentWillUnmount() {
```
tbergq: lifecycle methods should be the first methods of the class, in the order they are being called.

app/MMB/src/components/sectionListScrollTo/SectionListScrollTo.js
```diff
@@ -0,0 +1,240 @@
...
+    clearInterval(this.verifyAndScroll);
+  }
+
+  render = () => {
+    if (!this.props.data) {
+      return null;
+    }
+    return (
+      <React.Fragment>
```
tbergq: Is this fragment necessary?

app/MMB/src/components/sectionListScrollTo/OnLayout.js
RobinCsl: Dimensions is taking the dimension of the whole window, as opposed to the dimensions of the element it seems.
screen shot 2018-07-31 at 15 01 31

tbergq: ahh, sorry, I misunderstood the usecase

tbergq: Then could you please rename the children prop to render in this case?

RobinCsl: Of course!

app/MMB/src/components/sectionListScrollTo/SectionListScrollTo.js
RobinCsl: No, it's not. Thanks

RobinCsl:
@tbergq  PR improved

PR #934, opened 2018-07-30
MMB: Prepare List screen for simple token usage

tbergq:
Nothing will happen yet when you click on a booking when using simple token. That is up in the next PR.

### DIFF
--- a/app/core/src/screens/MMBPackageWrapper.js
+++ b/app/core/src/screens/MMBPackageWrapper.js
@@ -8,17 +8,22 @@ import { Translation } from '@kiwicom/mobile-localization';
 import { StyleSheet, type DimensionType } from '@kiwicom/mobile-shared';
 
 type Props = {|
-  navigation: NavigationType,
-  dimensions: DimensionType,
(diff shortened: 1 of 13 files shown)


app/core/src/screens/MMBPackageWrapper.js
```diff
@@ -39,26 +44,53 @@ export default class MMBPackageWrapper extends React.Component<
...
-        currency="EUR"
-        accessToken={this.state.token}
-      />
-    );
+    if (this.state.bookingId && this.state.simpleToken) {
+      return (
+        <ManageMyBookingPackage
+          onNavigationStateChange={this.props.onNavigationStateChange}
+          dimensions={this.props.dimensions}
+          currency="EUR"
```
RobinCsl: Currency is hardcoded?

tbergq: This wrapper acts like the native application. It passes the props that the native package will pass to MMB.

It is not a part of the app in production.

PR #948, opened 2018-08-02
Refactor: core navigation

RobinCsl:
A proposition to have a clearer understanding of how the routing is organised. I will refactor other navigation folders in this way if this sounds like a good idea.


tbergq:
This is basically just moving files to other flolders right?

RobinCsl:
That's right, plus moving the navigationOptions closer to the component being rendered, by using `static`.

PR #952, opened 2018-08-02
Hotels: Fix navigation bug on detail screen

tbergq:
@ferrannp , this is basically undoing what you did in commit `fix: Create WithStandaloneScreen HOC.` 27th of July

I guess you did this change for a reason, so please check what I broke with this change


ferrannp:
Sure, this was done because SingleHotel should be able to be shown as a Modal too. I'll be working on this.

PR #955, opened 2018-08-03
fix: Details shown as modal

ferrannp:


tbergq:
Seems to work now, I tested it both in rn and rn-playground.

PR #960, opened 2018-08-06
chore: Update native dependencies

### LINKED ISSUE #944: Upgrade dependencies with native packages

ferrannp:
Closes #944.

- react-native-vectors-icons to 5.0.0


.build/package.json
```diff
@@ -1,12 +1,12 @@
...
   "main": "index.js",
   "dependencies": {
     "react-native": "^0.55.4",
     "react-native-maps": "^0.21.0",
-    "react-native-vector-icons": "^4.6.0",
-    "react-native-tooltips": "^0.0.4"
+    "react-native-vector-icons": "^5.0.0",
+    "react-native-tooltips": "^0.0.6"
   }
 }
```
tbergq: I don't think we need this file anymore. This is the package.json file that was uploaded to npm. So maybe we can delete it?

ferrannp: We need to think a better way then because Android reads from that `package.json` as it has everything needed (like the rnkiwimobile versions uses that `0.0.25`.

Btw, when we merge this, let's see if all the process is automatically updated correctly (new versions and stuff, specially Android).

tbergq: Right, I did not know that android still used it. 

So yes, we need a better way 😄

PR #961, opened 2018-08-06
MMB: Show menu item in explore component

tbergq:


app/MMB/src/components/exploreCity/ExploreText.js
```diff
@@ -17,32 +20,39 @@ const ExploreText = (props: Props) => {
...
+          ) : (
+            <Translation id="mmb.main_menu.explore_city.find_stuff_to_do" />
+          )}
+        </Text>
+      </View>
+      <View style={[styles.padding, styles.buttonContainer]}>
+        <ExploreButton cityName={city || ''} />
+      </View>
+      <SeparatorFullWidth />
+      <ExploreMenuGroup data={airport} />
```
RobinCsl: It seems like this should be concerning the airport of the first segment, e.g. on trip Vienna-Birmingham with segments Vienna-Dublin, Dublin-Birmingham, the explore menu group has button 
`Order Transport to DUB`.

tbergq: You are right, this should be arrival, not the actual leg. I will fix it tomorrow.

tbergq: I am unable to reproduce what you are mentioning. See screenshot. 
image

RobinCsl: And this is what I see in the real app:
image from ios

tbergq:
This feature needs more refactoring, we will merge this PR as is for now.

PR #966, opened 2018-08-07
MMB: Explore component when time to takeoff > 8

tbergq:


app/MMB/src/components/exploreCity/ExploreMulticity.js
```diff
@@ -3,26 +3,16 @@
...
-  const date = new Date(idx(props.data, _ => _.trips[0].departure.time) || 0);
-  const daysLeft = DateUtils.diffInDays(date, DateUtils.getUTCToday());
-  const index = trips.findIndex(
-    trip =>
-      new Date(idx(trip, _ => _.departure.time) || 0) > DateUtils.getUTCToday(),
-  );
-  const trip = idx(props.data, _ => _.trips[daysLeft > 0 ? 0 : index - 1]);
-
-  return <ExploreText data={trip} />;
+  return <ExploreVariant data={idx(props.data, _ => _.trips[0])} />;
```
RobinCsl: Is this temporary? It seems like it should not be just about the first trip.

tbergq: Yes, this is the first step, I will do changes accordingly for the next steps

PR #969, opened 2018-08-09
MMB: Add Prior to booking component for explore on main menu

tbergq:


packages/localization/src/vocabularies/ManageMyBooking.js
```diff
@@ -255,9 +255,21 @@ export default {
...
+  'mmb.main_menu.explore_city.card_content.explore_airport':
+    'Explore __airport__ airport',
   'mmb.main_menu.explore_city.explore_menu_group.order_transport':
     'Order transportation to __iata__',
+  'mmb.main_menu.explore_city.prior_to_departure.flying_soon':
+    "You're flying soon",
+  'mmb.main_menu.explore_city.prior_to_departure.flying_to_destination':
+    'Your flight to __1_destination__ departs at __2_time__.',
+  'mmb.main_menu.explore_city.prior_to_departure.go_to_gate':
+    ' Go to gate __gate__.',
```
RobinCsl: Extra space

packages/localization/src/vocabularies/ManageMyBooking.js
```diff
@@ -255,9 +255,21 @@ export default {
...
+  'mmb.main_menu.explore_city.prior_to_departure.recheck_your_bags':
+    " Don't forget to recheck your bags.",
```
RobinCsl: Same

packages/localization/src/vocabularies/ManageMyBooking.js
```diff
@@ -255,9 +255,21 @@ export default {
...
+  'mmb.main_menu.explore_city.prior_to_departure.recheck_your_bags_and_go_to_gate':
+    " Don't forget to recheck your bags, then go to gate __gate__.",
```
RobinCsl: Here too ;-)

packages/localization/src/vocabularies/ManageMyBooking.js
tbergq: Yes, it is intended, because it has to be added after another translation

packages/localization/src/vocabularies/ManageMyBooking.js
tbergq: Yes, it is intended, because it has to be added after another translation

packages/localization/src/vocabularies/ManageMyBooking.js
tbergq: Yes, it is intended, because it has to be added after another translation

packages/localization/src/vocabularies/ManageMyBooking.js
RobinCsl: Hmm. I believe it should be independent from all the possible contexts where the translation is used. Maybe add a `<Translation passThrough=" " />` between the rendering of these translations.

app/MMB/src/components/exploreCity/ExploreText.js
```diff
@@ -5,25 +5,17 @@ import { View } from 'react-native';
...
-        <Text style={styles.title}>{props.title}</Text>
-        <Text style={styles.subText}>{props.text}</Text>
-      </View>
-      <View style={[styles.padding, styles.buttonContainer]}>
-        <ExploreButton cityName={props.city} />
-      </View>
-    </React.Fragment>
+    <View style={styles.padding}>
+      <Text style={styles.title}>{props.title}</Text>
+      <Text style={styles.subText}>{props.text}</Text>
```
RobinCsl: I don't know if that suggestion makes sense, but you could have added it here, instead of manually?

tbergq: Sorry, I could have added what?

PR #970, opened 2018-08-09
Adding alert to multicity timeline trip overview

RobinCsl:


app/MMB/src/scenes/tripOverview/MulticityTimeline.js
```diff
@@ -3,26 +3,125 @@
...
+      const departure = idx(trip, _ => _.departure);
+      const arrivalLocationId =
+        idx(arrivalPreviousTrip, _ => _.airport.locationId) || '';
+      const departureLocationId =
+        idx(departure, _ => _.airport.locationId) || '';
+      const arrivalCityName =
+        idx(arrivalPreviousTrip, _ => _.airport.city.name) || '';
+      const departureCityName = idx(departure, _ => _.airport.city.name) || '';
+
+      if (arrivalLocationId !== departureLocationId) {
```
tbergq: Isn't this more or less the same as you did in `ReturnTimeline`? Would it be possible to gather this logic in one component?

Just an idea, not sure if it is doable, but if the `Timeline` takes an array of Trips instead, could not the `Timeline` perform this check?

RobinCsl: I guess the logic to have the warnings between segments of one trip could be inside `Timeline`. However, when considering the departure of, say, the outbound and the arrival of the inbound, I could not perform the check inside of `Timeline`, as this is specific to a Return journey. 
Similarly for Multicity trips, I will still need to check that each previous trip's arrival matches or not the current trip's departure.

RobinCsl: Inside `Timeline` could live some logic about checking that legs/segments' arrival and departure coincide or not. However, for the relationship between trips, these somehow need to live in the parent component.
I could however abstract the logic in some helper file, dealing with trips.

tbergq: > these somehow need to live in the parent component

Why? Say timeline accepts `Leg[]` could it not then check that arrival of leg[index] === leg[index + 1].departure?

RobinCsl:
@tbergq Thank you for your suggestions; it should be better now.

PR #974, opened 2018-08-09
Explore airports and destinations - tests

lukewalczak:


.jest.json
```diff
@@ -1,7 +1,5 @@
 {
-  "collectCoverage": false,
   "preset": "react-native",
-  "verbose": false,
```
tbergq: what happend here?

lukewalczak: by default, it is set as `false`

https://github.com/facebook/jest/blob/master/packages/jest-cli/src/lib/update_global_config.js#L120-#L122

PR #976, opened 2018-08-10
Hotels: providing new screen to AppRegistry for new design

tbergq:


app/hotels/src/HotelsContext.js
```diff
@@ -4,27 +4,67 @@ import * as React from 'react';
...
+  const reg = /^\d{4}-\d{2}-\d{2}$/;
+  return reg.test(input);
+};
+
+export const getAsUtcDate = (input: ?string): Date | null => {
+  if (input == null || !validateDate(input)) {
+    return null;
+  }
+  const [year, month, date] = input.split('-').map(item => parseInt(item));
+  return new Date(Date.UTC(year, month - 1, date));
```
RobinCsl: Sometimes, JavaScript is weird; why is the month the only one to be 0-indexed?

app/hotels/src/map/allHotels/NewAllHotelsMap.js
```diff
@@ -0,0 +1,110 @@
...
+
+  render = () => {
+    const {
+      checkin,
+      checkout,
+      cityId,
+      currency,
+      roomsConfiguration,
+    } = this.props;
+    if (checkin === null || checkout === null) {
```
RobinCsl: TODO warning too?

tbergq: Yes, should be

tbergq: Actually, here it is not needed. The first screen the user will see is AllHotelsScreen which will show error. 

On tablet the user will se the error on the left half. I think that should be sufficient. Or do you think otherwise?

RobinCsl: Then if the map is not 'functioning', the button to access it should be disabled if there is an error on the first screen. Or if the button is not disabled, I would still show an error for people not to think it is a bug that the map is not loading.

tbergq: I guess we could disable the map button when I fix the header.

PR #977, opened 2018-08-10
iOS: Integrate logic for apple wallet button

lukewalczak:


app/MMB/src/scenes/appleWallet/WalletContent.js
```diff
@@ -2,11 +2,29 @@
...
 import DateAndTime from './DateAndTime';
+import WalletContext from '../../context/WalletContext';
+
+const createAddHandler = pkpassUrl => async () => {
+  try {
+    await PassBook.addPass(pkpassUrl);
+  } catch (err) {
+    return Alert.translatedAlert(undefined, {
+      passThrough:
+        'There was an error during communication with our server. Please try again later',
```
tbergq: You need to create a translation for this in `ManageMyBooking.js`

lukewalczak: 🤦‍♂️

tbergq:
Also, I think we have to: 

A: Add a scroll view, or 
B: Expand the height of the view

image

lukewalczak:
I am more inclined to use `ScrollView`

tbergq:
I did a quick check on an iPhone 5 (the smaller one) it seems what they do is expand the view, and reduce the image.

tbergq:
We should be able to do this by replacing the hardcoded heights with flex I guess

lukewalczak:
Can you please attach the screenshot from iPhone 5?

tbergq:
it is so pretty 🙈 
file

PR #980, opened 2018-08-13
Hotels: Add missing screens to new navigation stack

tbergq:


app/hotels/src/navigation/NewNavigationStack.js
```diff
@@ -1,15 +1,24 @@
-// @flow
+// @flow strict
```
RobinCsl: Shall we use `flow strict` everywhere now?

app/hotels/src/navigation/NewNavigationStack.js
```diff
@@ -1,15 +1,24 @@
...
 } from '@kiwicom/mobile-navigation';
+import { Platform } from 'react-native';
 import { withMappedNavigationAndConfigProps as withMappedProps } from 'react-navigation-props-mapper';
+import { Translation } from '@kiwicom/mobile-localization';
+import { Dimensions, Device } from '@kiwicom/mobile-shared';
 
 import SingleHotel from './singleHotel/SingleHotelNavigationScreen';
 import SearchResults from './allHotels/SearchResultsScreen';
 import AllHotelsMap from './allHotelsMap/AllHotelsMapNavigationScreen';
+import AdditionalPropsInjecter from './singleHotel/AdditionalPropsInjector';
```
RobinCsl: ...Inject**o**r

app/hotels/src/navigation/NewNavigationStack.js
tbergq: I have not enforced it strictly. 

The difference is you can not use `Function`, `any` `Object` etc.

I use it where it makes sense. If I need to use `any` or `Object` for some reason, I use just `Flow`

PR #981, opened 2018-08-13
Insurance context: order summary

RobinCsl:


app/MMB/src/scenes/tripServices/insurance/insuranceOverviewScene/orderSummary/OrderSummary.js
```diff
@@ -2,45 +2,48 @@
...
     <View style={styleSheet.wrapper}>
-      <View style={styleSheet.row}>
-        <View style={styleSheet.item}>
-          <Text>
-            <Translation passThrough="TODO" />
-          </Text>
-        </View>
-        <Price amount={-1} currency="WTF" style={styleSheet.price} />
-      </View>
+      <InsuranceRow insuranceType="TRAVEL_PLUS" />
```
tbergq: It is just the translation that is different right?

I think it is better to pass in the translation than the insurance type in that case.

tbergq: Oh, right, there are something more

RobinCsl: The issue is that I compute the quantity there and I need it in the translation.

PR #982, opened 2018-08-13
Hotels: Add custom header to new search result list

tbergq:


app/hotels/src/navigation/allHotels/SearchResultsHeader.js
```diff
@@ -0,0 +1,95 @@
...
+                          weekday: 'long',
+                          day: '2-digit',
+                          month: 'long',
+                        })}
+                      />
+                    }
+                  />
+                )}
+              </View>
+              {checkin !== null && (
```
tbergq: Not showing the button, so navigation to that hotels map is not possible, then showing only error on search result screen is sufficient.

packages/localization/src/DateFormatter.js
```diff
@@ -76,6 +93,9 @@ function DateFormatter(rawDate: Date = DateUtils.getUTCNow()) {
         pad(rawDate.getUTCDate())
       );
     },
+
+    // Pass inn your own configuration
```
RobinCsl: in--n--

PR #993, opened 2018-08-16
Hotels: show error when dates are wrong

tbergq:


packages/localization/src/vocabularies/Hotels.js
```diff
@@ -17,6 +17,8 @@ export default {
   'hotels_search.all_hotels.please_search':
     'Please search for your desired destination.',
   'hotels_search.all_hotels_search_list.no_hotels_found': 'No hotels found.',
+  'hotels_search.all_hotels_search.date_error':
+    'Ops, there seems to be something wrong with the chosen dates',
```
RobinCsl: O**o**ps?

tbergq: Yeah, you are right, I had to look that up, I did not know. Hopefully, this error should never be shown in production. 🤞

RobinCsl: 🤞

tbergq:
@RobinCsl 🚀

PR #999, opened 2018-08-20
Hotels: disable values from changing in visible popup

### LINKED ISSUE #996: Disappearing selection in filters
[Trello](https://trello.com/c/Wc3fltQC/308-bug-disappearing-selection-in-filters)

## Steps to reproduce

- Select one filter
- While loading, open another filter
- Change some value, but do not save, wait until loading finished
- Filter gets reset to default value.

tbergq:
closes #996

### DIFF
--- a/app/hotels/src/filter/hotelFacilities/HotelFacilitiesPopup.js
+++ b/app/hotels/src/filter/hotelFacilities/HotelFacilitiesPopup.js
@@ -98,7 +98,8 @@ export default class HotelFacilitiesPopup extends React.Component<
     facilities: this.props.facilities,
   };
 
-  static getDerivedStateFromProps = ({ facilities }: Props) => ({ facilities });
+  static getDerivedStateFromProps = ({ facilities, isVisible }: Props) =>
+    isVisible ? null : facilities;
 
   handleCheckboxOnPress = (option: string) => () =>
     this.setState(state => {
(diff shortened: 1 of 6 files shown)


app/hotels/src/filter/hotelFacilities/HotelFacilitiesPopup.js
```diff
@@ -98,7 +98,14 @@ export default class HotelFacilitiesPopup extends React.Component<
     facilities: this.props.facilities,
   };
 
-  static getDerivedStateFromProps = ({ facilities }: Props) => ({ facilities });
+  static getDerivedStateFromProps = ({ facilities, isVisible }: Props) => {
+    if (!isVisible) {
```
RobinCsl: I find it a bit counter-intuitive that `!isVisible` returns non null.

tbergq: Should I rather flip it to `if(isVisble) return null`?

RobinCsl: Maybe rename the variable to make it clearer why `isVisible` corresponds to `null`. Something like 
```
({facilities, isVisible: isFilterSelectionVisible}:Props) => {
const isPopUpVisible = !isFilterSelectionVisible;
if (isPopUpVisible) {
return ...}
return null;
```

tbergq: I don't agree. The component is named `...Popup` and isVisible should be quite clear what is going on. 

I would also expect that something from outside should not change my popup while I am looking at it. 

I will make the returns a little neater though.

tbergq:
@RobinCsl 🚀

PR #1004, opened 2018-08-21
MMB: Add hotels menu item in trip services

tbergq:


RobinCsl:
When on the test account I check this for Prague-Copenhagen, it does do anything. It would be nice to disable the button or not show it at all if there are no relevant locations returned by the API?

tbergq:
I does not do anything yet because it is a oneway booking. It has only one destination, so it should open the modal directly, which is a TODO

app/MMB/src/scenes/tripServices/HotelMenuItem.js
```diff
@@ -0,0 +1,98 @@
...
+class HotelMenuItem extends React.Component<Props, State> {
+  state = {
+    isPopupVisible: false,
+  };
+
+  menuItemOnPress = () => {
+    const relevantLocations =
+      idx(this.props.data, _ => _.hotel.relevantLocations) || [];
+    if (relevantLocations.length === 1) {
+      // TODO: Navigate to hotels modal
```
RobinCsl: I missed that part, got it now!

PR #1005, opened 2018-08-21
MMB: Open hotels package from trip services

tbergq:


app/MMB/src/navigation/NavigationStack.js
```diff
@@ -42,6 +43,22 @@ const TravelDocumentStack = StackNavigator(
   },
 );
 
+// This is not good enough since it just opens as a new page, not modal
+// It is complicated, needs to be improved later
```
RobinCsl: We would need to create an issue to make the process of opening windows as modals a lot easier.

app/MMB/src/scenes/tripServices/hotelMenuItem/HotelMenuItem.js
```diff
@@ -0,0 +1,150 @@
...
+      dataSaverEnabled: false, // TODO
+      checkin: '2018-09-10', // TODO
+      checkout: '2018-09-27', // TODO
+      onNavigationStateChange: () => {}, // TODO
+      onBackClicked: null, // TODO:
+      dimensions: this.props.dimensions,
+      version: 'rn-development', // TODO
+      cityName,
+      cityId,
+      roomsConfiguration: [{ adultsCount: 1, children: [] }],
```
RobinCsl: Since you could get this information from querying passengers, this would be TODO as well, right?

tbergq: yes

PR #1006, opened 2018-08-21
Connect graphql to transportation trip services

lukewalczak:


app/MMB/src/scenes/tripServices/LocationPopupButton.js
```diff
@@ -30,18 +33,26 @@ class LocationPopupButton extends React.Component<Props> {
...
-          )}
+          <View style={styleSheet.locationWrapper}>
+            {this.props.displayIata ? (
+              <Translation
+                passThrough={`in ${cityName || ''} (${locationId || ''})`}
+              />
+            ) : (
+              <Translation passThrough={cityName || ''} />
+            )}
+            {this.props.displayDate && (
```
tbergq: this prop seems a bit redundant. Date is already optional. How about just
```
{this.props.date != null && }
```

app/MMB/src/scenes/tripServices/TransportationMenuItem.js
```diff
@@ -0,0 +1,122 @@
...
+  +onOpenWebview: string => void,
+|};
+
+type State = {|
+  popupVisible: boolean,
+|};
+
+class TransportationMenuItem extends React.Component<Props, State> {
+  state = {
+    popupVisible: false,
```
tbergq: could you rename to isPopupVisible

app/MMB/src/scenes/tripServices/TransportationMenuItem.js
```diff
@@ -0,0 +1,122 @@
...
+            return (
+              <LocationPopupButton
+                key={whitelabelURL}
+                data={location}
+                date={date}
+                whitelabelURL={whitelabelURL}
+                onPress={this.openTransportationLink}
+                displayIata={false}
+                displayDate={
+                  index === 0 || index === relevantLocations.length - 1
```
tbergq: do this check in date instead, and remove this props

if index > 0 || index < relevantLocations.length - 1 pass in null, else pass in date

app/MMB/src/scenes/tripServices/LocationPopupButton.js
```diff
@@ -30,18 +32,26 @@ class LocationPopupButton extends React.Component<Props> {
...
-          )}
+          <View style={styleSheet.locationWrapper}>
+            {this.props.displayIata ? (
+              <Translation
+                passThrough={`in ${cityName || ''} (${locationId || ''})`}
+              />
+            ) : (
+              <Translation passThrough={cityName || ''} />
+            )}
+            {this.props.date !== null && (
```
tbergq: this will now render if date is undefined, which one would not expect. 

change it to `this.props.date !== null`

app/MMB/src/scenes/tripServices/LocationPopupButton.js
```diff
@@ -15,6 +16,7 @@ type Props = {|
   +data: LocationPopupButtonType,
   +onPress: string => void,
   +displayIata?: boolean,
+  +date?: string,
```
tbergq: also, this should maybe be `string | null`

app/MMB/src/scenes/tripServices/LocationPopupButton.js
```diff
@@ -30,18 +32,26 @@ class LocationPopupButton extends React.Component<Props> {
...
-          )}
+          <View style={styleSheet.locationWrapper}>
+            {this.props.displayIata ? (
+              <Translation
+                passThrough={`in ${cityName || ''} (${locationId || ''})`}
+              />
+            ) : (
+              <Translation passThrough={cityName || ''} />
+            )}
+            {this.props.date && (
```
tbergq: if date is equal to '' this will crash on android

lukewalczak: will refactor to `!=`

tbergq: thanks

PR #1007, opened 2018-08-21
Insurance Menu Item displayed conditionally; can trigger alert if les…

RobinCsl:
…s than 48 hours before


app/MMB/src/scenes/tripServices/InsuranceMenuItem.js
```diff
@@ -1,20 +1,73 @@
...
+    }
+  };
+
+  isMoreThan48HoursBefore = () => {
+    const departureLocalTime = this.getDepartureLocalTime();
+    return (
+      departureLocalTime &&
+      DateUtils.diffInHours(
+        new Date(departureLocalTime),
+        new Date(Date.now()),
```
tbergq: I think you should departure.utc and DateUtils.getUTCNow() instead

app/MMB/src/scenes/tripServices/__tests__/InsuranceMenuItem.test.js
```diff
@@ -0,0 +1,301 @@
...
+import { Alert } from '@kiwicom/mobile-localization';
+
+import { InsuranceMenuItem } from '../InsuranceMenuItem';
+
+describe('InsuranceMenuItem', () => {
+  const dateNow = Date.now;
+  let spy;
+  beforeEach(() => {
+    spy = jest.spyOn(Alert, 'translatedAlert');
+    global.Date.now = jest.fn(() => 1482363367071);
```
tbergq: you could use `MockDate` for this

app/MMB/src/scenes/tripServices/__tests__/InsuranceMenuItem.test.js
```diff
@@ -0,0 +1,301 @@
...
+        },
+      ],
+      trip: {
+        departure: {
+          localTime: '2016-12-24T23:36:07.071Z',
+        },
+      },
+    };
+
+    // $FlowExpectedError: $refType missing
```
tbergq: could you tag it with `// $FlowRelayIssue: https://github.com/facebook/relay/issues/2394` instead?

app/MMB/src/scenes/tripServices/InsuranceMenuItem.js
```diff
@@ -24,8 +77,43 @@ export default class InsuranceMenuItem extends React.Component<Props> {
...
+          departure {
+            localTime
+          }
+        }
+      }
+      ... on BookingMulticity {
+        trips {
+          departure {
+            localTime
+          }
```
tbergq: this pattern is repeating
```
departure {
  localTime
}
```

it could be extracted to a new fragment and be passed date from this component accordingly.

app/MMB/src/scenes/tripServices/InsuranceMenuItem.js
RobinCsl: `departure.utc` is not available in the GraphQL response; this is what we have in GraphQL server:
```
    time: {
      type: GraphQLDateTime,
      resolve: ({ when }: DepartureArrival): ?Date =>
        when == null ? null : when.utc, // intentional ==, can be null or undefined
    },

    localTime: {
      type: GraphQLDateTime,
      resolve: ({ when }: DepartureArrival): ?Date =>
        when == null ? null : when.local, // intentional ==, can be null or undefined
    },
```
So you'd rather I use `time` as opposed to `localTime`?

tbergq: The user could be in a different time zone 48 hours-ish before departure. That could lead to wrong comparisasion.

RobinCsl: Of course, I will change it!

RobinCsl:
@tbergq 🚀

PR #1009, opened 2018-08-22
MMB: Fix hotels modal

tbergq:


tbergq:
No crashing header, and you can close the modal

PR #1012, opened 2018-08-23
MMB: Add checkin, checkout and roomsconfiguration to hotels in tripservices

tbergq:


app/MMB/src/scenes/tripServices/hotelMenuItem/HotelMenuItem.js
```diff
@@ -43,21 +55,46 @@ class HotelMenuItem extends React.Component<PropsWithContext, State> {
...
+            DateUtils(new Date(checkin)).addDays(1),
+          ).formatForMachine()
+        : dates.checkout;
+    return {
+      checkin,
+      checkout,
+    };
+  };
+
+  openHotelsModal = ({ cityId, cityName, ...rest }: HotelData) => {
```
RobinCsl: Imho, it would be clearer that you de-structure checkin and checkout.

tbergq: you are probably right, to name more variables 😞 

I will fix it 😄

tbergq:
@RobinCsl 🚀

PR #1031, opened 2018-08-28
Hotels: Add single hotels context

tbergq:


app/hotels/src/navigation/singleHotel/SingleHotelNavigationScreen.js
```diff
@@ -61,18 +62,17 @@ class SingleHotelNavigationScreen extends React.Component<Props> {
 
   render() {
     return (
-      <SingleHotel
-        onGoToPayment={this.goToPayment}
-        onGoToMap={this.goToMap}
-        currency={this.props.currency}
-        language={this.props.language}
```
RobinCsl: Is `language` not necessary at all anymore?

tbergq: When we started with hotels, we had this prop passed in from native. Now we can fetch it from `DeviceInfo`

RobinCsl: I see!

PR #1036, opened 2018-08-28
Android card module Native<->JS protocol

ferrannp:
- Prepared to be implemented by native apps that consume it
- Provides a mock implementation using shared preferences

### DIFF
--- a/packages/rnmodules/card/CardManager.js
+++ b/packages/rnmodules/card/CardManager.js
@@ -4,8 +4,8 @@ import { NativeModules } from 'react-native';
 
 type CardType = {
   cardholder: string,
-  expiryMonth: number,
-  expireYear: number,
+  expiryMonth: string,
+  expiryYear: string,
   number: string,
 };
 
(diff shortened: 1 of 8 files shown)


packages/rnmodules/card/CardManager.js
```diff
@@ -0,0 +1,18 @@
...
+  expiryYear: string,
+  number: string,
+};
+
+export const getCard = (): Promise<CardType> | null => {
+  return NativeModules.RNCardManager.getCard();
+};
+
+export const saveCard = (card: CardType) => {
+  NativeModules.RNCardManager.saveCard(card);
```
tbergq: Could this return some kind of indication on whether the save was successful or not?

packages/rnmodules/card/CardManager.js
```diff
@@ -0,0 +1,18 @@
+// @flow
+
+import { NativeModules } from 'react-native';
+
+type CardType = {
+  cardholder: string,
+  expiryMonth: string,
+  expiryYear: string,
+  number: string,
```
tbergq: also this should be strict type

PR #1041, opened 2018-08-29
MMB: Delete locally stored boarding passes

tbergq:


app/MMB/src/scenes/tickets/TicketDeleteButton.js
```diff
@@ -75,11 +75,34 @@ export class TicketDeleteButton extends React.Component<
...
     const existsEticket = await RNFetchBlob.fs.exists(this.getEticketPath());
+    const existsBoardingPasses = await this.hasLocalBoardingPasses();
+    this.setState({ hasLocalFiles: existsEticket || existsBoardingPasses });
+  };
 
-    this.setState({ hasLocalFiles: existsEticket });
+  hasLocalBoardingPasses = async () => {
+    const boardingPasses = await RNFetchBlob.fs.ls(this.getBoardingPassPath());
+    return boardingPasses.some(file =>
+      file.includes(`-bookingId:${this.props.bookingId}`),
```
RobinCsl: I guess it is unlikely that there would be pdfs which are not boarding passes and which would have a name matching this criterion in this folder. But there could be collisions. I believe it's fine, just wanted to mention it.

tbergq: since we store it in `/boardingPasses` folder structure, it would be unwise to put anything but boarding passes there.

But of course, it is not fool proof 🙈

PR #1044, opened 2018-08-29
Adding current address to address picker screen

RobinCsl:


app/MMB/src/scenes/tripServices/transportation/TransportationMap.js
```diff
@@ -138,21 +138,22 @@ export class TransportationMap extends React.Component<Props, State> {
...
-        latitude,
-        longitude,
-      });
-    }
-
-    this.props.navigation.navigate('AddressPickerScreen', {
-      currentAddress,
-    });
+  openLocationPicker = () => {
+    navigator.geolocation.getCurrentPosition(
```
tbergq: we have a geolaction context, maybe you could use that one instead?

RobinCsl: Ah, of course! And actually, I just saw in that context file
```
// TODO: Show alert to tell user to go to settings an allow us to access location
// Need alert with translations first
```
which I can do in this PR too 😄

RobinCsl:
@tbergq Refactored quite a bit, and concerns are clearer now I believe. Let me know what you think 😃

packages/shared/src/GeolocationContext.js
```diff
@@ -61,3 +103,22 @@ class Provider extends React.Component<Props, State> {
 }
 
 export default { Consumer, Provider };
+
+export const withGeolocationContext = (Component: React.ElementType) => {
+  const withGeolocationContext = (props: {}) => {
+    return (
+      <Provider shouldNotUpdateGeolocationComponentDidMount={true}>
```
tbergq: It is better if this provider is just defined once application wide. Please rather add it to `MMB/index.js`

Also not sure about this prop. Maybe we should just always fail silently instead, or pass a callback to `updateGeolocation` so the consumer can decide what to do with the error.

RobinCsl:
@tbergq I tried to implement your suggestions, let me know your thoughts!

app/MMB/src/scenes/tripServices/transportation/TransportationMap.js
```diff
@@ -138,46 +135,15 @@ export class TransportationMap extends React.Component<Props, State> {
     }
   }
 
-  openLocationPicker = async () => {
-    const { currentLocation } = this.state;
-    let currentAddress;
+  shouldComponentUpdate = (nextProps: Props, nextState: State) => {
```
tbergq: this seems abundant. When I comment it, it has just as many rerenders as when it is there

app/MMB/src/context/GoogleMapsContext.js
```diff
@@ -3,30 +3,82 @@
...
+  getAddress = async (coordinate: Coordinate) => {
+    const { latitude, longitude } = coordinate;
+    if (latitude != null && longitude != null) {
+      const address = await fetch(
+        `https://maps.googleapis.com/maps/api/geocode/json?latlng=${latitude},${longitude}&key=${
+          this.props.googleMapsAPIKey
+        }`,
+      )
+        .then(res => res.json())
+        .then(json => json.results[0].address_components);
```
tbergq:
```
const res = await fetch(
        `https://maps.googleapis.com/maps/api/geocode/json?latlng=${latitude},${longitude}&key=${
          this.props.googleMapsAPIKey
        }`,
      )
const json = await res.json();
return json.results[0].address_components;
```

packages/shared/src/GeolocationContext.js
```diff
@@ -1,39 +1,89 @@
...
 
+type Coordinate = {|
+  +latitude: number,
+  +longitude: number,
+|};
+
 type Props = {|
-  children: React.Node,
+  +children: React.Node,
+  +shouldNotUpdateGeolocationComponentDidMount?: boolean,
```
tbergq: this props is not being used, can you remove it?

packages/shared/src/GeolocationContext.js
```diff
@@ -1,39 +1,89 @@
...
+      ...defaultState,
+      actions: {
+        updateGeolocation: this.updateGeolocation,
+      },
+    };
+  }
 
   componentDidMount = () => {
-    this.updateGeolocation();
+    this.updateGeolocation(() => {}, () => {});
```
tbergq: if both parameters are optional, why do you need to pass a noop function?

app/MMB/src/context/GoogleMapsContext.js
```diff
@@ -15,18 +15,16 @@ type Props = {|
   +googleMapsAPIKey: string,
 |};
 
-type State = {|
-  +googleMapsAPIKey: string,
-|};
+type State = {| googleMapsAPIKey: string |};
```
tbergq: this could also be read-only, since it will not be changed through setState.

packages/shared/src/GeolocationContext.js
```diff
@@ -1,39 +1,89 @@
...
+          onPress: this.openSettings,
+          style: 'default',
+        },
+      ],
+    );
+  };
+
+  updateGeolocation = (
+    dealWithLocation?: (coordinate: Coordinate) => void = () => {},
+    onError?: () => void = this.alertOpenSettings,
```
tbergq: just leave this as `onError?: () => void` and let the consumer decide what to do, so we don't have to pass in two empty functions every time

PR #1047, opened 2018-08-30
Adding custom input field in navigation header on address location in…

RobinCsl:
…put screen


tbergq:
@RobinCsl you have an unhappy test

PR #1050, opened 2018-08-30
Eslint: Add no-did-update-set-state rule

tbergq:


app/MMB/src/scenes/tripOverview/TimelineTitle.js
```diff
@@ -29,38 +30,25 @@ type PropsWithContext = {|
...
-          _warning =>
-            _warning.timelineTitle.localTime == localTime &&
-            _warning.timelineTitle.iataCode == iataCode,
-        );
-        this.setState({ warning });
-      }
-    }
-  }
+class TimelineTitle extends React.Component<PropsWithContext> {
+  getWarning = memoize(
```
RobinCsl: This is to have it executed only once, right?

tbergq: `memoize-one simply remembers the last arguments, and if the function is next called with the same arguments then it returns the previous result.`

app/MMB/src/scenes/tripOverview/TimelineTitle.js
```diff
@@ -69,7 +57,10 @@ class TimelineTitle extends React.Component<PropsWithContext, State> {
     const localTime = idx(data, _ => _.localTime);
     const cityName = idx(data, _ => _.airport.city.name);
     const iataCode = idx(data, _ => _.airport.locationId);
-    const warningStyle = this.state.warning ? styleSheet.warningStyle : null;
+    const warnings = idx(this.props, _ => _.warnings) || [];
+    const warning = this.getWarning(warnings, localTime, iataCode);
+
+    const warningStyle = warning ? styleSheet.warningStyle : null;
```
RobinCsl: Do you want to add `!= null` like below?

app/MMB/src/scenes/tripOverview/TimelineTitle.js
RobinCsl: I see, to optimize calls. Cool.

tbergq:
@RobinCsl 🚀

PR #1054, opened 2018-08-31
Add e2e testing setup

tbergq:


e2e/init.js
```diff
@@ -0,0 +1,22 @@
+// @flow
+
+import detox from 'detox';
+import adapter from 'detox/runners/jest/adapter';
+
+import { detox as config } from '../package.json';
+
+jest.setTimeout(120000);
```
RobinCsl: Does that mean that any e2e test must be 2 minutes long max?

tbergq: Probably yes, lets increase it when needed, I have not added any tests yet, other than a proof of concept. Also, it is not running on the CI so.

RobinCsl: Sounds good to me!

PR #1058, opened 2018-08-31
Refund insurance mutation

RobinCsl:
This is still WIP, but in case you could have a look at the relay store/cache not updating correctly.


RobinCsl:
@tbergq Improved it 🚀

app/MMB/src/scenes/tripServices/insurance/refundScene/RefundScene.js
```diff
@@ -0,0 +1,171 @@
...
+                <Translation id="mmb.trip_services.insurance.refund.confirm" />
+              }
+              onPress={this.onPress}
+            />
+          </View>
+        </View>
+        <Modal
+          visible={this.state.isSubmitting}
+          transparent={true}
+          onRequestClose={this.noop}
```
tbergq: maybe we should add this modal to shared, so we can reuse it in other components in the future?

PR #1059, opened 2018-09-03
MMB: Add trip overview titles for oneway and return bookings

tbergq:


app/MMB/src/scenes/tripOverview/tripTitle/TripTitle.js
```diff
@@ -4,23 +4,30 @@ import * as React from 'react';
...
+const TripTitle = ({ type, data, index }: PropsWithContext) => {
   return (
     <View style={styles.row}>
       {type === 'BookingMulticity' && <MulticityTitle data={data} />}
-      {/* TODO: Return */}
-      {/* TODO: OneWay */}
+      {type === 'BookingReturn' && <ReturnTitle isOutbound={index === 0} />}
+      {type === 'BookingOneWay' && (
+        <TripTitleText>
+          <Translation id="mmb.trip_overview.trip_title.outbound" />
```
RobinCsl: `trip_title.oneway` or something similar?

tbergq: it is actually also outbound, so I see no need for another key here

RobinCsl: Ah, okay! 👍

PR #1062, opened 2018-09-03
MMB: Trip overview add icons, model and terminal

tbergq:


app/MMB/src/scenes/tripOverview/TimelineDeparture.js
```diff
@@ -31,24 +38,44 @@ function TimelineDeparture(props: Props) {
...
+                  id="mmb.flight_overview.timeline.terminal"
+                  values={{
+                    terminal,
+                    city,
+                    iata,
+                  }}
+                />
+              }
+            />
+          )}
```
RobinCsl: There should also be one like that for the TimelineArrival_routeStop where the icon is the landing plane.

packages/localization/src/vocabularies/ManageMyBooking.js
```diff
@@ -70,6 +70,8 @@ export default {
     "Warning: you will depart from a different place. You'll arrive at __1_arrival__ and leave from __2_departure__.",
   'mmb.flight_overview.timeline.warning.different_airport_return_first':
     'Warning: your departure point differs from your return one. You will depart from __1_departure__ and return to __2_arrival__.',
+  'mmb.flight_overview.timeline.terminal':
+    '__1_city__ - __2_iata__ - Terminal - __3_terminal__',
```
RobinCsl: I think it should be 
```
'__1_city__ __2_iata__ - Terminal - __3_terminal__'
```

app/MMB/src/scenes/tripOverview/TimelineDeparture.js
tbergq: It is no landing icon in the current app

tbergq: I will do this in another PR. There is still more missing parts here anyways.

PR #1083, opened 2018-09-10
Refactor iOS playground

lukewalczak:
New iOS playground provides:

1. Present `OldHotels`
2. Push `OldHotels`
3. Present `NewHotels`


ios/RNNativePlayground/RNOldHotelsOptions.h
```diff
@@ -0,0 +1,8 @@
+#import <UIKit/UIKit.h>
+#import <RNKiwiMobile/RNKiwiMobile.h>
+
+@interface RNOldHotelsOptions : UIViewController <RNKiwiOptions>
```
ferrannp: I would keep everywhere same naming (like in JS). So, `RNHotels...` and `RNNewHotels...`, without the `Old` anywhere.

ios/RNNativePlayground/RNOldHotelsOptions.h
```diff
@@ -0,0 +1,8 @@
...
+
+- (instancetype)initWithDimensions:(NSDictionary *)dimensions andNavigationMode:(NSString *)lastNavigationMode;
```
ferrannp: Is this still good? like... can we just have something like `initWithParams` and pass both? Just saying.

ios/RNNativePlayground/ViewController.m
```diff
@@ -22,67 +26,63 @@ - (instancetype)initWithCoder:(NSCoder *)coder {
      * RNKiwiViewController does exactly the same thing for you behind the scenes if you use
      * the former approach.
      */
-    [[RNKiwiSharedBridge sharedInstance] initBridgeWithOptions:self];
+    [[RNKiwiSharedBridge sharedInstance] initBridgeWithOptions:[[RNOldHotelsOptions alloc] init]];
```
ferrannp: This is confusing me. Why it uses `RNOldHotelsOptions` ? Isn't this shared between old and new?

ios/RNNativePlayground/ViewController.m
```diff
@@ -108,9 +108,8 @@ - (NSString *)localizedStringWithKey:(NSString *)key {
 - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldReceiveTouch:(UITouch *)touch {
   
   if (self.navigationController.interactivePopGestureRecognizer == gestureRecognizer) {
-    return [_vc isInteractivePopGestureAllowed];
+    return [_activeVc isInteractivePopGestureAllowed];
```
ferrannp: Was this change necessary?

lukewalczak: Yes. Before there was only `vc`. Now there are `activeVc` and `hotelsVc`, but we would like to disable gestures in the currently opened view controller.

ios/RNNativePlayground/RNOldHotelsOptions.h
lukewalczak: Refactored

ios/RNNativePlayground/RNOldHotelsOptions.h
lukewalczak: Refactored

grabbou:
Looks good to me. Overall, I think it's worth explaining the motivation behind this PR a little bit more. Generally speaking, it's based on findings when working on the documentation (part of another PR) to make sure the interface ends up being much easier than it is right now.

There are few changes, but it's much smaller and simpler now.

PR #1090, opened 2018-09-11
 iOS documentation

lukewalczak:


docs/README.md
```diff
@@ -5,10 +5,13 @@ Documentation describing how to integrate the official native app and React Nati
...
 
 ### [Android](android)
+
 On Android we use a gradle library called `rnkiwimobile`.
 
-### iOS
-TODO
+### [iOS](ios)
+
+On iOS we use a `.framework` generated by script.
```
grabbou: Probably better to say: "On iOS, we use a framework called `RNKiwiMobile`"

docs/ios/README.md
```diff
@@ -0,0 +1,31 @@
+# iOS distribution
+
+The source code can be found in [RNNativePlayground](https://github.com/kiwicom/mobile/tree/master/ios/RNNativePlayground)
+
+## Test playground app
+
+### KiwiHotels module
+
+1. To run `KiwiHotels` module with the packager, change [`RNKiwiConstants.m`](https://github.com/kiwicom/mobile/blob/master/ios/RNKiwiMobile/RNKiwiConstants.m) like follows:
```
grabbou: This step is not needed here. Probably we could create a separate section, called `Running from packager` or `Dev mode`.

docs/ios/README.md
```diff
@@ -0,0 +1,31 @@
...
+}
+```
+
+2. Then run the packager from the root of this repo:
+
+```bash
+yarn start
+```
+
+3. Finally just open the project in XCode (use `reactNativeApp.xcworkspace`) and choose `RNNativePlayground` scheme.
```
grabbou: Nit: `Finally, open the project` (it's a good practice to avoid using "simply" and "just" as these things are never that easy!)

Nit: I would suggest changing "... and choose" to ". Make sure you have "RNNativePlayground" scheme selected before running. 

Would be great to link `scheme selected` to an Apple documentation where it shows how to do it with XCode. I assume there might be people using it that have little experience (is that wrong assumption?)

docs/ios/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+![](../../.docs/assets/open_playground.gif)
+
+_NOTE_: If you want to test playground with `jsbundle` go directly to step 3.
```
grabbou: Once we remove points 1/2 as per my first comment in this section, we can remove this :) 

Also, instead of saying `jsbundle`, better to say: "If you want to test production bundle..."

docs/ios/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Usage in a real app
+
+Paste the `.framework` within real app.
```
grabbou: You can't really `paste` it. You should break this documentation down into multiple steps, including (but not limited to) the following points:
- Link static library with the target where you want to use the app
- Import RNKiwiViewController and instantiate it
- Push it

Good place to mention pre-initialising the bridge by explicitly calling it. That's also where we will mention constants

docs/ios/deploy.md
```diff
@@ -0,0 +1,23 @@
+# Deploy
+
+## Deploy automatically
+
+Every merge to `master` branch will provide a new release and upload the `.framework` into github releases.
```
grabbou: provides, instead of `will provide`

docs/ios/deploy.md
```diff
@@ -0,0 +1,23 @@
...
+
+To check current releases [click here](https://github.com/kiwicom/mobile/releases)
+
+If we want to trigger it manually, we can do so from the root of this repo:
```
grabbou: Instead of `we`, let's use `you`.

docs/ios/native-dependencies.md
```diff
@@ -0,0 +1,39 @@
+# Native dependencies
+
+## Add a new external native dependency
+
+All iOS code is kept in `ios` folder and using the same `Podfile` that's why process of adding a new external native dependency can be achieved by running a command:
```
grabbou: This is a bit unclear to me. I would simply title this paragraph: "Adding a React Native dependency", because `link` will not work with pure CocoaPods modules not downloaded from `npm`. This may get confusing.

docs/ios/native-dependencies.md
```diff
@@ -0,0 +1,39 @@
...
+
+```bash
+react-native link NATIVE_DEPENDENCY_NAME
+```
+
+## Add a new internal native dependecy
```
grabbou: This one is about adding external native dependencies and internal packages too.

docs/ios/react-native-module.md
```diff
@@ -0,0 +1,100 @@
...
+
+To create a new React Native module just:
+
+1. Create a new folder inside `app` folder, eg: `app/newModule`, where `newModule` stands for the new module name.
+
+2. Inside `newModule/index.js` export all standalone views.
+
+## Using a React Native module in a playground
+
+In the [`app/native`](https://github.com/kiwicom/mobile/blob/master/app/native.js) import all views from a module of your choice and register them, which then you can use by the defined name, eg:
```
grabbou: Nit: them, which then you can use by the defined name, eg:

Should be: register them by name that you will later refer to from the native side.

docs/ios/react-native-module.md
```diff
@@ -0,0 +1,100 @@
...
+```js
+// Hotels
+AppRegistry.registerComponent('KiwiHotels', () => HotelsStandalonePackage);
+AppRegistry.registerComponent(
+  'NewKiwiHotels',
+  () => NewHotelsStandAlonePackage
+);
+```
+
+Registered module has to be then allocated in along with proper `moduleName` and `initialProperties`, like below:
```
grabbou: Nit: allocated in? Probably a better choice is to use `instantiated`

docs/ios/react-native-module.md
```diff
@@ -0,0 +1,100 @@
...
+
+```objc
+  RNKiwiViewController *vc = [[RNKiwiViewController alloc] initWithModule:@"KiwiHotels"
```
grabbou: For the sake of demo, we can probably make `initialProperties` much smaller. Also this one has some formatting issues.

docs/ios/README.md
grabbou: Oh, just noticed you did that in a different page. Then, either link to it from here, or move it here and link from there. Your choice! 🗡

grabbou:
Much better now! Thank you :) I suspect we will get more polished stage later.

Unfortunately, I don't have enough permissions to land this PR. We gotta wait :)

PR #1095, opened 2018-09-14
[MOBILE-841] Adding Mapbox on single hotel view

RobinCsl:
Could someone check that it works correctly on Android?


android/app/build.gradle
```diff
@@ -186,6 +186,7 @@ dependencies {
     implementation "com.trinerdis.skypicker:logging:0.2.1-SNAPSHOT"
     implementation("com.google.android.gms:play-services-base:$playServicesVersion")
     implementation("com.google.android.gms:play-services-maps:$playServicesVersion")
+    compile project(':mapbox-react-native-mapbox-gl')
```
ferrannp: Should be `implementation`.

RNAndroidPlayground/rnkiwimobile/build.gradle
```diff
@@ -48,6 +48,7 @@ dependencies {
     api "com.trinerdis.skypicker:react-native-maps:$reactNativeMaps-SNAPSHOT"
     api "com.trinerdis.skypicker:react-native-vector-icons:$reactNativeVectorIcons-SNAPSHOT"
     api "com.trinerdis.skypicker:react-native-tooltips:$reactNativeTooltips-SNAPSHOT"
+    api "com.trinerdis.skypicker:mapbox-react-native-mapbox-gl:$reactNativeTooltips-SNAPSHOT"
```
ferrannp: Should be `$reactNativeMapBox` and be defined in `dependencies.gradle`.

ferrannp:
For now @lukewalczak is trying to run this on iOS playground and me on Android playground. Both are having problems. We need to fix this before merging it.

ferrannp:
After conversation with @Blackhex we came to conclusion that we cannot just use this library as it is. The RN wrapper uses:

```
compile('com.mapbox.mapboxsdk:mapbox-android-sdk:5.4.1@aar') {
    transitive=true
}
```

But Kiwi app uses `6.5.0`. The code from this library is not compatible with `5.4.1` and seeing repo state, our best bet might be to fork and fix.

tbergq:
> seeing repo state, our best bet might be to fork and fix.

Which would also mean we have to maintain it. Does not sound like a good option

tbergq:
@RobinCsl can we close this PR for now? I guess we have to wait for new version of mapbox?

RobinCsl:
Absolutely; here is the PR we need to watch
https://github.com/mapbox/react-native-mapbox-gl/pull/1377

PR #1097, opened 2018-09-17
Hotels: hide tab bar inside hotel package

tbergquist-godaddy:
Add SafeAreaView around content

### DIFF
--- a/app/core/src/navigation/stacks/HotelsStack.js
+++ b/app/core/src/navigation/stacks/HotelsStack.js
@@ -10,7 +10,7 @@ import Homepage from '../../screens/hotelsStack/Homepage';
 import SingleHotelsPackageWrapper from '../../screens/hotelsStack/SingleHotelPackageWrapper';
 import NewHotelsPackageWrapper from '../../screens/hotelsStack/NewHotelsPackageWrapper';
 
-export default StackNavigator(
+const Stack = StackNavigator(
   {
     Home: withMappedProps(Homepage),
     SingleHotelPackage: withMappedProps(SingleHotelsPackageWrapper),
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -13,6 +13,7 @@ YellowBox.ignoreWarnings([
   'Class GenericShare was not exported. Did you forget to use RCT_EXPORT_MODULE()',
   'Class WhatsAppShare was not exported. Did you forget to use RCT_EXPORT_MODULE()',
   'Class GooglePlusShare was not exported. Did you forget to use RCT_EXPORT_MODULE()',
+  '{"name":"Invariant Violation","framesToPop":1}',
```
RobinCsl: Are we really meant to ignore this warning? I think it is linked to react-navigation; it is not happy about some stuff. We would really need to refactor it I think.

tbergq: I would love to remove it, anyways, this only helps while useing external debugger so I can just as well remove it

app/core/src/navigation/stacks/HotelsStack.js
```diff
@@ -10,7 +10,7 @@ import Homepage from '../../screens/hotelsStack/Homepage';
 import SingleHotelsPackageWrapper from '../../screens/hotelsStack/SingleHotelPackageWrapper';
 import NewHotelsPackageWrapper from '../../screens/hotelsStack/NewHotelsPackageWrapper';
 
-export default StackNavigator(
+const Stack = StackNavigator(
```
RobinCsl: We need to find a way to make this 'on-demand', because there are other views in MMB, like Insurances, where it would be needed to. But it does not seem as straightforward as it is (more) deeply nested.

tbergq: This is an entirely different problem, since in production, the tab bar is provided by native code. We would need a bridge to hide it

tbergq:
@RobinCsl fixed the index.js file

app/core/src/navigation/stacks/HotelsStack.js
RobinCsl: Oh right, of course. I will create an issue for that.

PR #1099, opened 2018-09-18
[WIP] - Update React Native to 0.57.0

ferrannp:


tbergq:
Logging in to MMB, click on any booking gives me this red screen

image

ferrannp:
@tbergq your issues should be fixed. This PR is ready to be reviewed and tested.

app/hotels/src/map/HotelDetailPreview.js
```diff
@@ -20,13 +20,13 @@ type Props = {|
 |};
 
 const getReview = review => {
-  if (review && review.count) {
+  if (review && review.count != null) {
```
tbergq: I would guess review.count !== 0 is more relevant in this case?

babel.config.js
```diff
@@ -0,0 +1,24 @@
+// @flow
+
+// $FlowFixMe
```
tbergq: Please use // $FlowExpectedError: and explain why we are supressing flow

packages/shared/src/view/__tests__/AdaptableLayout.wide.test.js
```diff
@@ -12,28 +12,21 @@ jest.mock('../../Device', () => ({
...
   ).toMatchSnapshot();
 });
 
-it('renders empty element if there is no wide layout', () => {
+it('renders narrow component if there is no wide one', () => {
   expect(
     Renderer.create(
       <Dimensions.Provider dimensions={{ width: 200, height: 500 }}>
-        <AdaptableLayout renderOnNarrow={<ThrowsError />} />
+        <AdaptableLayout renderOnNarrow={<RendersCorrectly />} />
```
mrtnzlml: How is this related to RN 57? This is obviously wrong - see the generated snapshot: `RendersCorrectly` is not rendered at all and therefore this test doesn't work anymore. It was correct with `ThrowsError` component, or not? (the same for the first test) cc @tbergq

tbergq: Thanks, I missed that detail

PR #1101, opened 2018-09-18
Hotels: List to map transition

tbergq:


app/hotels/src/allHotels/NewAllHotels.js
```diff
@@ -1,24 +1,28 @@
...
 
 export default function NewAllHotels() {
   return (
-    <ScrollView contentContainerStyle={styles.scrollViewContainer}>
-      <FilterStripe />
-      <NewAllHotelsSearch />
-    </ScrollView>
+    <View style={styles.container}>
+      <View style={styles.absoluteFill}>
+        <View style={{ zIndex: 1000 }}>
```
RobinCsl: I think it'd be a good idea to have a list of 'official' zIndex's so that it does not become impossible to track which layer is on top of which. Maybe 5 different zIndex's should be enough. What do you think?

RobinCsl: (Besides, this one style is inline :P)

app/hotels/src/allHotels/RenderResults.js
```diff
@@ -0,0 +1,124 @@
+// @flow strict
```
RobinCsl: Could it be renamed `RenderSearchResults` to make it a bit more evocative?

app/hotels/src/allHotels/NewAllHotels.js
tbergq: I know, I was fixing that 🙈 

It turns out we already have this in orbit tokens, so I will use from there

app/hotels/src/allHotels/RenderResults.js
```diff
@@ -0,0 +1,124 @@
...
+type PropsWithContext = {|
+  ...Props,
+  +show: ResultType,
+|};
+
+const topValue = 1000;
+const lowValue = -100;
+
+class RenderResults extends React.Component<PropsWithContext> {
+  mapAnimation = new Animated.Value(topValue);
```
RobinCsl: Are you instantiating these class properties twice? Once here and once below in the constructor?

app/hotels/src/allHotels/RenderResults.js
```diff
@@ -0,0 +1,124 @@
...
+          useNativeDriver: true,
+        }),
+      ]).start();
+    }
+  };
+  render = () => (
+    <View style={styles.container}>
+      <Animated.View
+        style={{
+          ...StyleSheet.absoluteFillObject,
```
RobinCsl: Is it possible to make this style not inline? I guess it does not make a difference to define in render before the return statement.

app/hotels/src/map/allHotels/NewAllHotelsMap.js
```diff
@@ -91,6 +84,7 @@ class NewAllHotelsMap extends React.Component<Props> {
                     filterParams.hotelFacilities,
                   ),
                 },
+                first: 50,
```
RobinCsl: Will this never change?

tbergq: no, not for maps

app/hotels/src/navigation/allHotels/SearchResultsContext.js
```diff
@@ -0,0 +1,65 @@
...
+  };
+
+  render = () => (
+    <ContextProvider value={this.state}>{this.props.children}</ContextProvider>
+  );
+}
+
+export default { Consumer, Provider };
+
+export function withSearchResultsContext(select: (state: State) => Object) {
```
RobinCsl: We should create some utility function that converts our contexts into these HOC `withContext`, because there are many like this in the project!

app/hotels/src/allHotels/RenderResults.js
tbergq: I could make this part a cached style, but we need inline styles for Animated.View to work anyways

app/hotels/src/navigation/allHotels/SearchResultsContext.js
tbergq: Feel free to make PR with it

tbergq:
@RobinCsl 🚀

PR #1104, opened 2018-09-19
Hotels: New design for detail page

tbergq:


packages/navigation/index.js
```diff
@@ -93,7 +94,13 @@ export const StackNavigatorOptions = {
 export const createTransparentHeaderStyle = (dim: DimensionType) => {
   if (Platform.OS === 'ios' || Device.isWideLayout(dim)) {
     // normal header on iOS and wide Android
-    return {};
+    return {
+      headerStyle: {
+        backgroundColor: 'transparent',
+        marginTop: 20,
+      },
+      headerTransparent: true,
```
RobinCsl: Is it somehow possible to change the color of the native header (phone signal, battery level, etc.) to be in a different color than black when the image at the top is really dark? And then switch it back to black once you scroll down and enter white areas (or re-instate a non-transparent header?)

app/hotels/src/singleHotel/roomList/formatBeddingInfo.js
```diff
@@ -5,17 +5,12 @@ import idx from 'idx';
...
-  const type = idx(room, _ => _.type);
   const beddingOptions = idx(room, _ => _.bedding) || [];
-  const bedding = beddingOptions
+  return beddingOptions
     .map(beddingOption => {
       const type = idx(beddingOption, _ => _.type);
       const amount = idx(beddingOption, _ => _.amount);
       return [amount, type].filter(value => value != null).join(' ');
     })
     .join(' or ');
```
RobinCsl: Not part of the PR, but is this translated later on?

app/hotels/src/singleHotel/roomList/RoomBadges.js
```diff
@@ -60,7 +57,32 @@ export const RoomBadges = (props: Props) => {
     );
   }
 
-  return badges;
```
RobinCsl: I think you forgot to clean this file. It looks like `badges` is defined for no reason now and the code is almost duplicated.

app/hotels/src/singleHotel/RoomSummary.js
```diff
@@ -0,0 +1,71 @@
+// @flow strict
+
+import * as React from 'react';
+import { Translation } from '@kiwicom/mobile-localization';
+import { Text, StyleSheet, Price } from '@kiwicom/mobile-shared';
+import { defaultTokens } from '@kiwicom/mobile-orbit';
+import { View } from 'react-native';
+
+type Money = {|
```
RobinCsl: Maybe `PriceType` instead?

app/hotels/src/singleHotel/HotelDetailScreen.js
```diff
@@ -112,31 +120,43 @@ export class HotelDetailScreen extends React.Component<Props, State> {
...
+                data={availableHotel.availableRooms}
+                select={this.selectRoom}
+                deselect={this.deselectRoom}
+                selected={selected}
+              />
+              <BrandLabel />
+            </LayoutSingleColumn>
+          </ScrollView>
+          <View style={styles.buttonContainer}>
+            <RoomSummary
```
RobinCsl: Maybe the props should be renamed `guestNumber`/`guestQuantity` and `roomNumber`/`roomQuantity` to make it more descriptive of what it is.

packages/navigation/index.js
tbergq: Yes, it is possible, I didn't notice that detail, but yes, it is set to white on the design, we should definitely change it back to black when we scroll down

app/hotels/src/singleHotel/roomList/formatBeddingInfo.js
tbergq: It is not and it should be, I will add an todo comment here

app/hotels/src/singleHotel/roomList/RoomBadges.js
tbergq: good catch 😄

app/hotels/src/singleHotel/HotelDetailScreen.js
tbergq: I will go for guestCount/roomCount

tbergq:
@RobinCsl 🚀

app/hotels/src/singleHotel/HotelDetailScreen.js
RobinCsl: That's much better than my suggestions :D

app/hotels/src/singleHotel/HotelDetailScreen.js
```diff
@@ -112,31 +145,48 @@ export class HotelDetailScreen extends React.Component<Props, State> {
...
-            />
-            <Header hotel={availableHotel.hotel} />
-            <HotelInformation hotel={availableHotel.hotel} />
-            <RoomList
-              data={availableHotel.availableRooms}
-              select={this.selectRoom}
-              deselect={this.deselectRoom}
-              selected={selected}
+      <SafeAreaView style={styles.safeArea} forceInset={{ top: 'never' }}>
+        <StatusBar backgroundColor="blue" barStyle="dark-content" />
```
RobinCsl: Should this backgroundColor use the defaultTokens?

app/hotels/src/singleHotel/HotelDetailScreen.js
```diff
@@ -51,6 +68,20 @@ export class HotelDetailScreen extends React.Component<Props, State> {
...
 
+  componentWillUnmount = () => {
+    if (Platform.OS === 'ios') {
+      StatusBar.setBarStyle('dark-content');
+    }
+  };
+
+  onScroll = (e: NativeEvent) => {
+    if (Platform.OS === 'ios' && e.nativeEvent.contentOffset.y > 180) {
+      StatusBar.setBarStyle('dark-content');
```
RobinCsl: I would also not make it transparent after we passed the picture, but I guess it's up to the designer. Imho, you cannot really read the status bar correctly once it's over some text.

tbergq: No, default is the default value. But sorry, there is another method `setHidden` which will hide it. Personally id dislike hiding it

RobinCsl: Yeah, I would not like it to be hidden either. So I'll approve it as it is.

tbergq:
```
Apple's official word is: 

Think twice before hiding the status bar if your application is not a game or full-screen media-viewing application. Although these apps might permanently hide the status bar, you should understand the ramifications of this design decision. Permanently hiding the status bar means that users must quit your app to find out, for example, whether they need to recharge their device.
```

PR #1105, opened 2018-09-19
Change entry-file on buildAndroidJavascript

ferrannp:


ferrannp:
Ok I will add it manually for now in the script. Then I opened https://github.com/kiwicom/mobile/issues/1106.

PR #1107, opened 2018-09-19
Hotels modals new design

denbalogh:
Hello, this is my first pull request here at kiwi.
The slider, used in ScorePopup and PricePopup is still WIP.


app/hotels/src/filter/Delimiter.js
```diff
@@ -0,0 +1,45 @@
+// @flow
```
tbergq: We already have these components, they can be imported as
`import { SeparatorFullWidth, SeparatorTrimmed } from '@kiwicom/mobile-navigation'`

app/hotels/src/filter/hotelFacilities/HotelFacilitiesPopup.js
```diff
@@ -124,11 +126,13 @@ export default class HotelFacilitiesPopup extends React.Component<
           key={i}
           isChecked={facilities.indexOf(key) >= 0}
           onPress={this.handleCheckboxOnPress(key)}
-          style={i < facilitiesListKeys.length - 1 && styles.delimiter}
         >
           <View style={styles.checkbox}>
-            <Icon name={facility.icon} size={20} style={styles.facilityIcon} />
+            <Icon name={facility.icon} size={24} style={styles.facilityIcon} />
```
tbergq: Please try to use icons from https://orbit.kiwi/visual-style/icons/

they can be used like

```
import { TextIcon } from '@kiwicom/mobile-shared';
<TextIcon code="&#xe001;" orbit={true} />
```

packages/shared/src/buttons/ButtonSecondary.js
```diff
@@ -0,0 +1,57 @@
+// @flow strict
```
tbergq: This differs from button only by backgroundColor right?
Can we rather reuse the button component and pass it type="secondary" or something like that?

packages/shared/src/forms/Checkbox.js
```diff
@@ -21,18 +21,16 @@ export default function Checkbox(props: Props) {
...
-              size={Platform.select({
-                android: 25, // solves checkbox row jumping on Android devices
-                ios: 26,
-              })}
-              color={defaultTokens.paletteProductNormal}
-            />
-          </View>
-        )}
+        <View style={[styles.check, props.isChecked ? styles.checked : {}]}>
+          <Icon
```
tbergq: use orbit icon

packages/shared/src/forms/Slider.js
```diff
@@ -1,9 +1,10 @@
 // @flow
 
 import * as React from 'react';
-import { View } from 'react-native';
-import type { OnLayout } from '@kiwicom/mobile-shared';
+import { View, Text } from 'react-native';
```
tbergq: Use text from `@kiwicom/mobile-shared` instead

app/hotels/src/filter/hotelFacilities/HotelFacilitiesPopup.js
```diff
@@ -120,17 +121,29 @@ export default class HotelFacilitiesPopup extends React.Component<
...
+                orbit={true}
+                style={styles.facilityIcon}
+              />
+              {facility.text}
+            </View>
+          </Checkbox>
+          {i < facilitiesListKeys.length - 1 && (
+            <View style={styles.separatorEnd}>
+              <SeparatorTrimmed
+                gapSizeStart={24 + 7}
```
tbergq: Is it a reason why it is 24 + 7 rather than 31?

denbalogh: I wrote it like this for readability, it's iconSize + margin, but I guess it wasn't that obvious 😆The margin is actually 10, but it wasn't properly aligned in the emulator, when I used orbit icons, so I changed it to 7 and now it doesn't make sense 😆I'll fix it.

tbergq: In that case it would have been better to make to variables, iconSize & margin and done `gapSizeStart={iconSize + margin}`

denbalogh:
Hi, I think I am finished with the task. Can you review me please @tbergq @RobinCsl

tbergq:
@denbalogh Seems you need to fix some tests. You should run `yarn test-ci` before creating PR 😊

app/hotels/src/filter/hotelFacilities/HotelFacilitiesPopup.js
```diff
@@ -120,17 +121,29 @@ export default class HotelFacilitiesPopup extends React.Component<
...
-          style={i < facilitiesListKeys.length - 1 && styles.delimiter}
-        >
-          <View style={styles.checkbox}>
-            <Icon name={facility.icon} size={20} style={styles.facilityIcon} />
-            {facility.text}
-          </View>
-        </Checkbox>,
+        <React.Fragment key={i}>
+          <Checkbox
+            isChecked={facilities.indexOf(key) >= 0}
```
tbergq: please use `isChecked={facilities.includes(key)}` instead

tbergq:
You need to add a `SafeAreaView` from `react-navigation` here (the one from `react-native` is broken. 

Also, the price you show in the slider is the price pr night, so if the designers wish to show price pr stay, you need to multiply it with the number of days, or create a new graphql query to get the price pr stay

image

tbergq:
Also this happens sometimes, and the slider sometimes `jumps` is there a way to make it smoother?

image

tbergq:
SliderLabels seems to have infinite re-render

skjermbilde 2018-09-21 kl 08 27 29

packages/shared/src/SliderLabels.js
```diff
@@ -0,0 +1,148 @@
...
+          paddingLeft: startOffset,
+          paddingRight: endOffset,
+        });
+      }
+    }
+
+    requestAnimationFrame(this.setPadding);
+  };
+
+  saveFullWidth = (e: any) => {
```
RobinCsl: Please, don't use `any` as a type. There should be `type OnLayout` available in `@kiwicom/mobile-shared`.

packages/shared/src/buttons/Button.js
```diff
@@ -44,14 +50,13 @@ const styleSheet = StyleSheet.create({
     justifyContent: 'center',
     alignItems: 'center',
     paddingHorizontal: 11,
+    height: 44,
+    borderRadius: 6,
+  },
+  primary: {
     backgroundColor: defaultTokens.paletteProductNormal,
-    android: {
```
RobinCsl: Are we sure that there are no differences between Android and iOS anymore?

app/hotels/src/filter/hotelFacilities/HotelFacilitiesPopup.js
```diff
@@ -120,17 +121,29 @@ export default class HotelFacilitiesPopup extends React.Component<
...
+            <View style={styles.checkbox}>
+              <TextIcon
+                code={facility.icon}
+                orbit={true}
+                style={styles.facilityIcon}
+              />
+              {facility.text}
+            </View>
+          </Checkbox>
+          {i < facilitiesListKeys.length - 1 && (
```
RobinCsl: Could you give that a name that is a bit more expressive to understand why the rendering is done or not?

app/hotels/src/filter/price/PricePopup.js
```diff
@@ -49,36 +56,35 @@ export default class PricePopup extends React.Component<Props, State> {
...
+          <Text style={styles.subtitle}>
+            <Translation id="hotels_search.filter.price_popup.subtitle" />
+          </Text>
         </Text>
+        <SliderLabels
+          max={max}
+          min={min}
+          startValue={start}
+          endValue={end}
+          startLabel={<Translation passThrough={`${start} ${currency}`} />}
```
RobinCsl: This is not a good idea: I think you should make it a translation id with two variables:
```
'new.translation.id': '__1_amount__ __2_currency__'
```
and then use it as
```
<Translation id='new.translation.id' values={{amount: "23", currency: "EUR"}} />
```

The reason is that in other languages, the currency sign can be in front, like in English: $20.

This solution is still not perfect if we had to deal with prices over 1,000, as you would need to add a comma in English, but not in French where it would be a space (1 000) for instance. Same with decimal values.

packages/shared/src/buttons/Button.js
denbalogh: Only text-transform seems to be different, but that's another component.

denbalogh:
Hello,
can you review again, please?

tbergq:
@denbalogh could you rebase master into your branch first. Your PR is on a lower version of react-native.

app/hotels/src/filter/price/PriceFilter.js
```diff
@@ -125,16 +176,31 @@ type Props = {|
   +isActive: boolean,
 |};
 
+const calculateDaysOfStay = (
```
tbergq:
```
import { DateUtils } from '@kiwicom/mobile-localization';


const daysOfStay = DateUtils.diffInDays(checkout, checkin);
```

app/hotels/src/filter/price/PricePopup.js
```diff
@@ -41,52 +50,92 @@ export default class PricePopup extends React.Component<Props, State> {
...
+            <Translation id="hotels_search.filter.price_popup.title" />
+            <Translation passThrough=" " />
+            <Text style={styles.subtitle}>
+              <Translation id="hotels_search.filter.price_popup.subtitle" />
+            </Text>
+          </Text>
+          <SliderLabels
+            max={max}
+            min={min}
+            start={{
```
tbergq: I would prefer startLabel={<Translation .../>} and startValue={...}

packages/navigation/src/Separators.js
```diff
@@ -40,12 +43,22 @@ type SeparatorTrimmedType = {|
  * |                           |
  */
 export function SeparatorTrimmed(props: SeparatorTrimmedType) {
-  const { gapSizeStart, gapSizeEnd, color } = props;
+  const { gapSizeStart, gapSizeEnd, color, height } = props;
```
tbergq: how about:
```
export function SeparatorTrimmed(props: SeparatorTrimmedType) {
  const { gapSizeStart, gapSizeEnd, color, height } = props;
  return (
    <View style={styleSheet.row}>
      <View style={[styleSheet.gap, { width: gapSizeStart, height }]} />
      <SeparatorFullWidth color={color} height={height} />
      <View style={[styleSheet.gap, { width: gapSizeEnd, height: height }]} />
    </View>
  );
}

SeparatorTrimmed.defaultProps = {
  gapSizeEnd: 0,
  height: 1,
};
```

packages/shared/src/SliderLabels.js
```diff
@@ -0,0 +1,177 @@
...
+          this.state.labelStartWidth -
+          this.state.labelEndWidth -
+          gap,
+      );
+    } else {
+      return Math.floor(this.state.width - this.state.labelStartWidth - gap);
+    }
+  };
+
+  calculateMarkerOffset = (type: 'start' | 'end'): number => {
```
tbergq: Why is this one function with a flag instead of

`calaculateMarkerStartOffset` and `calaculateMarkerEndOffset`?

packages/shared/src/SliderLabels.js
```diff
@@ -0,0 +1,177 @@
...
+      const endMarkerOffset = this.calculateMarkerOffset('end');
+      const endLabelHalf = this.state.labelEndWidth / 2;
+      const endLabelOffset =
+        endMarkerOffset < endLabelHalf ? 0 : endMarkerOffset - endLabelHalf;
+
+      if (startLabelOffset + endLabelOffset < maxPadding) {
+        if (
+          this.state.paddingLeft !== startLabelOffset ||
+          this.state.paddingRight !== endLabelOffset
+        ) {
```
tbergq: Here we are on 3rd nested if level, it is hard to read. Please try and refactor it a bit.

packages/shared/src/forms/__tests__/__snapshots__/Checkbox.test.js.snap
```diff
@@ -36,15 +36,21 @@ exports[`checked 1`] = `
...
           "justifyContent": "center",
         }
       }
     >
-      <Icon
-        color="#00a991"
-        name="check"
-        size={26}
+      <TextIcon
+        code="S"
```
tbergq: Maybe we should also test that it renders correctly on android since there now are different icons?

denbalogh:
@tbergq I fixed the comments, except the one with testing checkbox icon for android. I checked it on emulator, but I don't know how to write a test for it. Can you give me some hints?

app/hotels/src/filter/price/PriceFilter.js
```diff
@@ -125,20 +180,32 @@ type Props = {|
   +isActive: boolean,
 |};
 
+const calculateDaysOfStay = (checkin, checkout) => {
+  if (checkin && checkout) {
+    return DateUtils.diffInDays(checkout, checkin);
+  } else {
+    return 1;
```
tbergq: I am not sure this is the best fallback value, better to return null

app/hotels/src/filter/price/PriceFilter.js
```diff
@@ -125,20 +180,32 @@ type Props = {|
...
-            <PriceFilter
-              {...props}
-              currentSearchStats={currentSearchStats}
-              currency={currency}
-            />
-          )}
-        </HotelsSearchContext.Consumer>
-      )}
+      {({ currency, checkin, checkout }) => {
+        const daysOfStay = calculateDaysOfStay(checkin, checkout);
```
tbergq: then you could check here of daysOfStay === null and in that case just return null. If any of the dates are missing, there is obvoiusly some error from native side.

packages/shared/src/SliderLabels.js
```diff
@@ -0,0 +1,190 @@
...
+type State = {|
+  width: number,
+  labelStartWidth: number,
+  labelStartAtMax: boolean,
+  labelEndWidth: number,
+  paddingLeft: number,
+  paddingRight: number,
+|};
+
+export default class SliderLabels extends React.Component<Props, State> {
```
tbergq: Also there is a lot of logic in this class, could you please also add some tests?

denbalogh: I added a test

packages/shared/src/__tests__/SliderLabels.test.js
```diff
@@ -0,0 +1,31 @@
...
+    <SliderLabels
+      startLabel={<Translation passThrough="start label" />}
+      startValue={46}
+      endLabel={<Translation passThrough="end label" />}
+      endValue={850}
+      max={1000}
+      min={1}
+    />,
+  );
+});
```
tbergq: Do these tests cover this logic?

```
if (isBelowMaxPadding) {
      if (hasOffsetChanged) {
        this.setState({
          paddingLeft: startLabelOffset,
        });
      }
       if (this.state.labelStartAtMax) {
        this.setState({ labelStartAtMax: false });
      }
    } else if (this.state.labelStartAtMax === false) {
      this.setState({ labelStartAtMax: true });
    }
```

app/hotels/src/filter/price/PriceFilter.js
```diff
@@ -125,20 +180,35 @@ type Props = {|
   +isActive: boolean,
 |};
 
+const calculateDaysOfStay = (checkin, checkout) => {
+  if (checkin && checkout) {
+    return DateUtils.diffInDays(checkout, checkin);
+  } else {
+    return null;
+  }
```
tbergq: this else block is unnecessary

```
if (checkin && checkout) {
    return DateUtils.diffInDays(checkout, checkin);
  } 
  return null; 
```
achieves just the same

PR #1108, opened 2018-09-19
Hotels: New design for roompicker button

tbergq:


app/hotels/src/singleHotel/roomPicker/IncrementDecrement.js
```diff
@@ -0,0 +1,105 @@
...
+  currency,
+  increment,
+  selectableCount,
+}: Props) {
+  return (
+    <View style={styles.container}>
+      <View style={styles.row}>
+        <View style={styles.priceContainer}>
+          <Text style={styles.countText}>
+            {Platform.select({
```
RobinCsl: Is it possible to make it clearer that only the prop textTransform is added on Android? Maybe creating the prop before the return statement and then just injecting or spreading a prop object?

app/hotels/src/singleHotel/roomPicker/SelectButton.js
```diff
@@ -0,0 +1,64 @@
...
+  +currency: string | null,
+  +increment: () => void,
+|};
+
+export default function SelectButton({ increment, price, currency }: Props) {
+  return (
+    <Touchable
+      style={[
+        styles.container,
+        styles.borderRadiusLeft,
```
RobinCsl: I cannot see this in the styles object. (Idem for `styles.borderRadiusRight`)

packages/shared/src/buttons/IncrementDecrementButtons.js
```diff
@@ -45,78 +35,60 @@ const Button = ({
 };
 
 export default function IncrementDecrementButtons(props: {|
-  onIncrement: () => void,
-  onDecrement: () => void,
-  number: number,
-  min?: number,
-  max?: number,
+  +onIncrement: () => void,
+  +onDecrement: () => void,
```
RobinCsl: Idem here.

packages/shared/src/buttons/IncrementDecrementButtons.js
```diff
@@ -45,78 +35,60 @@ const Button = ({
...
+        touchable={!disableDecrement}
+        onPress={props.onDecrement}
+      />
+      {props.showNumber && (
+        <Text style={props.numberStyle}>
+          <Translation passThrough={props.number} />
+        </Text>
+      )}
+      <Button
+        icon={<TextIcon code="&#xe122;" orbit={true} style={styles.icon} />}
```
RobinCsl: I am just starting to question adding this prop `orbit={true}` everywhere. Won't it be annoying once the migration to Orbit icons is complete to remove everything? Just thinking out loud.

app/hotels/src/singleHotel/roomPicker/IncrementDecrement.js
tbergq: Could you show with an example, I am not sure what you mean 😄

app/hotels/src/singleHotel/roomPicker/SelectButton.js
tbergq: Yes, this is the result of copy/paste, I deleted them now

packages/shared/src/buttons/IncrementDecrementButtons.js
tbergq: What?

packages/shared/src/buttons/IncrementDecrementButtons.js
tbergq: Just search and replace orbit={true} -> ''

app/hotels/src/singleHotel/roomPicker/IncrementDecrement.js
RobinCsl: Of course.
So before `return`, have something like: 
```
const myAndroidProps = Platform.OS === 'ios' ? {} : {textTransform: 'uppercase'};
```

and then just write 
```
<Translation id="the.id" {...myAndroidProps} />
```

packages/shared/src/buttons/IncrementDecrementButtons.js
tbergq: but yeah, maybe it is time to go the extra mile and just remove the old icons

packages/shared/src/buttons/IncrementDecrementButtons.js
RobinCsl: Oh it looks like my other comment was not taken into account: I meant for you to maybe extract the Props type to make it more readable.

packages/shared/src/buttons/IncrementDecrementButtons.js
RobinCsl: There's only 92 files to go through... 🍰

tbergq:
@RobinCsl 🚀

PR #1115, opened 2018-09-21
Shared: Fix AdaptableLayout.wide test

tbergq:


packages/shared/src/view/__tests__/AdaptableLayout.wide.test.js
```diff
@@ -11,6 +11,9 @@ jest.mock('../../Device', () => ({
 }));
 
 const RendersCorrectly = () => 'I am going to render without any problems...';
+const ThrowsError = () => {
+  throw new Error('do not render');
+};
 
 it('renders wide components', () => {
```
mrtnzlml: This test should contain `renderOnNarrow={<ThrowsError />}` as well. (see: https://github.com/kiwicom/mobile/pull/1099/files#diff-657350a5fb31355862c78e0f2c88b19c) The difference is significant - it works now as well but you don't know whether it's because `renderOnNarrow` is undefined or not called. But when you add there this component with the exception then you can be really sure that this prop is not used in this situation at all (which is exactly what you want). Maybe it's not so obvious? I am still a little bit confused by the change... :)

tbergq: Yeah, it should be fixed now 😄

PR #1117, opened 2018-09-21
Hotels: placing close and swipe list correctly for new design

tbergq:


app/hotels/src/map/allHotels/HotelSwipeList.js
```diff
@@ -89,49 +59,64 @@ class HotelSwipeList extends React.Component<Props, State> {
...
-      <AdaptableLayout
-        renderOnWide={
-          <View style={[styles.fullWidth, styles.wide]}>{child}</View>
-        }
-        renderOnNarrow={<View style={styles.fullWidth}>{child}</View>}
-      />
+      <View style={styles.container}>
+        <BottomSheet openHeight={openHeight} closedHeight={closedHeight}>
+          <BottomSheetHandle />
+          {data.length ? (
```
RobinCsl: If it is zero, then you'd have no results too, right?

tbergq: that is right

tbergq: Since this is a ternary it will render the error if data.length is null, undefined or 0. 

if this was {data.length && <.../>} it would cause crash on android if data.length was 0

PR #1122, opened 2018-09-24
Feature/new all hotels map

tbergq:
Should be ready to be merged. There might be a few details missing.


app/hotels/src/map/hotelDetailPreview/HotelDetailPreview.js
```diff
@@ -0,0 +1,130 @@
...
+import { Translation } from '@kiwicom/mobile-localization';
+import idx from 'idx';
+import { defaultTokens } from '@kiwicom/mobile-orbit';
+
+import type { HotelDetailPreview_availability } from './__generated__/HotelDetailPreview_availability.graphql';
+import { HotelDetailConsumer } from './HotelDetailPreviewContext';
+import HotelReviewScore from '../../components/HotelReviewScore';
+
+type ContainerProps = {|
+  +availability: $FlowFixMeProps,
```
RobinCsl: Can we do something about this?

tbergq:
@RobinCsl 🚀

app/hotels/src/map/hotelDetailPreview/HotelDetailPreview.js
```diff
@@ -0,0 +1,125 @@
...
+            </Text>
+            <View style={styles.row}>
+              <Stars
+                rating={idx(availability, _ => _.hotel.rating.stars)}
+                style={styles.stars}
+              />
+              <HotelReviewScore hotel={idx(availability, _ => _.hotel)} />
+            </View>
+            {price &&
+              price.currency != null &&
```
RobinCsl: I think we would need to refactor this component to accept a price object with properties currency amount and it would handle this kind of logic. I'll create an issue for it.

PR #1124, opened 2018-09-24
Android rnkiwimobile: Share bridge and init it in advance + Fragments

### LINKED ISSUE #853: Investigate how to init the bridge (and share it) in advance for the hybrid app

### LINKED ISSUE #1093: Embedding RN screen in Android fragment
If there will be a requirement to put RN screen/feature inside a tab navigation page on Android in the future it would be needed to render that screen inside a Fragment. Therefore it is needed to investigate whether this is technically possible with the current RN architecture and if yes provide some demo of this concept.

### LINKED ISSUE #1119: RNKiwiMobile activity should extend PermissionAwareActivity and handle it

ferrannp:
Fixes #853, fixes #1093 and fixes #1119.

This PR gives us:

- RN bridge will init when opening the Android app. Thus, when navigating to a RN view, it should be immediate (no white screen).
- When closing a RN view and coming back, the same should apply.
- We should be able to load RN views in Android Fragments.

## Progress

- [x] Init the bridge in advance (and only once)
- [x] Share bridge
- [x] Handle all remaining TODOs

For fragments:

- [x] Load RN app in fragment
- [x] Handle fragment lifecycle + back button
- [x] Handle all remaining TODOs

## How it works

We init the bridge and JS on our Application:

```kt
class PlaygroundApplication : Application() {

  lateinit var reactNativeHost: ReactNativeHost

  override fun onCreate() {
    super.onCreate()
    reactNativeHost = RNKiwiHost(this, getJSEntryPoint(), getPackages())
    // Init the bridge and everything in advance
    reactNativeHost.reactInstanceManager.createReactContextInBackground()
  }

  private fun getJSEntryPoint(): String {
    return "app/native"
  }

  // Those are the packages are module needs
  private fun getPackages(): MutableList<ReactPackage> {
    val hotelModulesInjection = HotelsModulesInjection()

    return mutableListOf(RNTooltipsPackage(),
        RNDeviceInfoPackage(),
        MapsPackage(),
        RNCurrencyManagerPackage(hotelModulesInjection.currencyCallback),
        RNTranslationManagerPackage(hotelModulesInjection.translationCallback),
        RNLoggingPackage(hotelModulesInjection.hasActiveBooking),
        VectorIconsPackage())
  }
}
```

The base activity or fragment will just use it:

```kt
reactInstanceManager = getReactNativeInstanceManager()

reactRootView = ReactRootView(this)
reactRootView.startReactApplication(reactInstanceManager, getModuleName(), getInitialProperties())
```

When we extend the base activity or fragment, we will `override` the instance:

```kt
  override fun getReactNativeInstanceManager(): ReactInstanceManager = (application as PlaygroundApplication).reactNativeHost.reactInstanceManager
```


RNAndroidPlayground/app/src/main/java/com/kiwi/rnandroidplayground/MainActivity.kt
```diff
@@ -3,16 +3,27 @@ package com.kiwi.rnandroidplayground
...
+
+    button_hotels_fragment.setOnClickListener {
+      startActivity(Intent(this, HotelActivityFragment.getViewModelClass()))
+    }
+  }
+
+  override fun onDestroy() {
+    super.onDestroy()
+    // TODO check where to use this
+    (application as PlaygroundApplication).reactNativeHost.reactInstanceManager.onHostDestroy(this)
```
Blackhex: In production application it would be probably the best to initialize the RN instance in main's activity onCreate()/onDestroy() since Application's onDestroy is tend to be used only for debugging and is not called in production app. Possibly this can be called also in https://developer.android.com/reference/android/app/Application.html#onTrimMemory(int) when memory is low or critical and if user has currenlty no RN screen opened but this would require more logic to handle this state properly and to reinitialize the RN on demand.

RNAndroidPlayground/react-native/build.gradle
```diff
@@ -11,7 +11,7 @@ buildscript {
         }
     }
     dependencies {
-        classpath "com.android.tools.build:gradle:$gradlePluginVersion"
+        classpath 'com.android.tools.build:gradle:3.1.4'
```
Blackhex: I would advise to use variable for Gradle version from dependencies.gradle here and also in RNAndroidPlayground/build.gradle.

RNAndroidPlayground/app/src/main/java/com/kiwi/rnandroidplayground/HotelFragment.kt
```diff
@@ -0,0 +1,36 @@
...
+class HotelFragment : RNKiwiFragment() {
+
+  override fun getModuleName(): String {
+    return "NewKiwiHotels"
+  }
+
+  override fun getReactNativeHost(): ReactNativeHost = (activity.application as PlaygroundApplication).reactNativeHost
+
+  // TODO extract code repetition
+  override fun getInitialProperties(): Bundle? {
```
Blackhex: I would rather use POJO (POKO actually :-)) for defining the data scructure than Bundle to be type safe.

RNAndroidPlayground/app/src/main/java/com/kiwi/rnandroidplayground/HotelActivityFragment.kt
```diff
@@ -0,0 +1,37 @@
...
+
+import android.app.Activity
+import android.os.Bundle
+
+import com.facebook.react.modules.core.DefaultHardwareBackBtnHandler
+
+class HotelActivityFragment : Activity(), DefaultHardwareBackBtnHandler {
+
+  companion object {
+    fun getViewModelClass(): Class<HotelActivityFragment> =
```
Blackhex: I do not get why is this method calleed getViewModelClass when it is actually returning view class?

RNAndroidPlayground/app/src/main/java/com/kiwi/rnandroidplayground/MainActivity.kt
ferrannp: Mmm yes so I guess you are kinda "free" on where you want to init RN. I don't think it would hang Splash screen by the way given the fact that it's done in the background?

RNAndroidPlayground/rnkiwimobile/src/main/java/com/kiwi/rnkiwimobile/RNKiwiFragment.kt
```diff
@@ -0,0 +1,136 @@
...
+//            val reactInstanceMgr = nReactNativeHost.reactInstanceManager
+//
+//            // onDestroy may be called on a ReactFragment after another ReactFragment has been
+//            // created and resumed with the same React Instance Manager. Make sure we only clean up
+//            // host's React Instance Manager if no other React Fragment is actively using it.
+//            if (reactInstanceMgr.lifecycleState != LifecycleState.RESUMED) {
+//                reactInstanceMgr.onHostDestroy(activity)
+//                nReactNativeHost.clear()
+//            }
+      nReactNativeHost.reactInstanceManager.onHostDestroy(activity!!)
```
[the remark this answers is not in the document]
ferrannp: Yup, there is a `TODO` in this method so still working on it.

RNAndroidPlayground/react-native/build.gradle
ferrannp: Yeah... Somehow my Android Studio updates those versions, it changes it like this instead of changing the value of the variable.

RNAndroidPlayground/app/src/main/java/com/kiwi/rnandroidplayground/MainActivity.kt
Blackhex: Depends on if it is 100% async or some part of that call is synchronous. I haven't tried that, I am just currious.

PR #1126, opened 2018-09-25
Hotels: New design for drop marker

tbergq:


packages/shared/src/__tests__/__snapshots__/DropMarker.test.js.snap
```diff
@@ -1,31 +1,167 @@
...
+    style={
+      Object {
+        "backgroundColor": "#0176D2",
+        "borderRadius": 2,
+        "flex": 1,
+      }
+    }
+  >
+    <TextIcon
+      code=""
```
RobinCsl: Is this going to be any different when we change the icon? It's not crucial but maybe this would not detect a new icon.

packages/shared/src/buttons/__tests__/IncrementDecrementButtons.playground.test.js
```diff
@@ -0,0 +1,63 @@
...
+    false,
+    'IncrementDecrementButtons',
+  );
+});
+
+it('render disabled', () => {
+  const onIncrement = jest.fn();
+  const onDecrement = jest.fn();
+  PlaygroundRenderer.render(
+    <View style={{ backgroundColor: 'white', padding: 10 }}>
```
RobinCsl: Maybe we could create a component container for these tests?

packages/shared/src/__tests__/__snapshots__/DropMarker.test.js.snap
tbergq: I fear that there is very little we can do about this.

packages/shared/src/buttons/__tests__/IncrementDecrementButtons.playground.test.js
tbergq: There is probably much that could be improved on the DX of our Playground. Feel free to look at it when we have spare time 🙂

RobinCsl: Impossible at the moment 😄

PR #1145, opened 2018-09-26
Relay: upgrade graphql

tbergq:


tbergq:
> Is that the only breaking change? Wow.

It's not even breaking change, it is just deprecated 😄

PR #1153, opened 2018-09-27
MMB Components: replacing TextIcons with Orbit icons

RobinCsl:
Waiting on updated font file for Orbit icons


RobinCsl:
@tbergq @ferrannp 🚀

PR #1159, opened 2018-09-28
Relay: make cache work for connections

tbergq:


packages/relay/src/Environment.js
```diff
@@ -86,11 +89,25 @@ export default function createEnvironment(accessToken: string = '') {
...
+                    selection => selection.kind !== 'LinkedHandle',
+                  ),
+                },
+              };
+
               // we have all data in the store to fulfill this query so let's
               // load it from the memory first and call the API after that
               // this will make the UI feel really fast
-              observer.next(store.lookup(operationSelector.root));
+              observer.next(store.lookup(root));
```
mrtnzlml: Could you please also catch the error here? Now it's hidden as you said.

packages/relay/src/Environment.js
```diff
@@ -24,6 +25,7 @@ const GRAPHQL_URL = 'https://graphql.kiwi.com/';
 const store = new Store(new RecordSource());
 
 function fetchFromTheNetwork(networkHeaders, operation, variables, observer) {
+  Emitter.emit('load-started');
```
mrtnzlml: Maybe something like `relay-load-started` would be better? Just in case you will use the emitter even for other things.

packages/relay/src/Environment.js
tbergq: It is actually catched, now I finally understand were the weird error message on the all hotels list came from 🙈

tbergq:
@mrtnzlml 🚀

PR #1167, opened 2018-10-02
Hotels: use new more performant query

tbergq:


app/hotels/src/components/HotelReviewScore.js
```diff
@@ -109,14 +106,3 @@ const style = StyleSheet.create({
     },
   },
 });
-
-export default createFragmentContainer(
```
RobinCsl: Why don't you use fragments anymore?

app/hotels/src/map/allHotels/HotelSwipeItem.js
```diff
@@ -32,11 +32,21 @@ export class HotelSwipeItem extends React.Component<Props> {
...
+    const name = idx(data, _ => _.name);
+    const price = idx(data, _ => _.price);
+    const thumbnailUrl = idx(data, _ => _.mainPhoto.thumbnailUrl);
+    const stars = idx(data, _ => _.rating.stars);
+    const score = idx(data, _ => _.review.score);
     return (
       <TouchableWithoutFeedback onPress={this.handlePress}>
         <View style={{ width }}>
-          <HotelDetailPreview availability={data} />
+          <HotelDetailPreview
```
RobinCsl: What's the benefit of making explicit all the props instead of passing one object?

PR #1177, opened 2018-10-04
Hotels: Improve image loader on search page

tbergq:


packages/shared/src/image/NetworkImage.js
```diff
@@ -41,11 +40,14 @@ export const NetworkImage = function NetworkImage(imageProps: Props) {
...
   };
+
   return <Image {...newProps} />;
 };
 
+const DefaultIndicator = () => null;
+
 NetworkImage.defaultProps = {
-  indicator: IconLoading,
+  indicator: DefaultIndicator,
```
RobinCsl: So no visual indication that something is loading, right?

tbergq: Yes, the designers exact works:

`I would remove the loading indicator on the hotel images (they are quite distracting, I would use just plain background with *Cloud light* color)`

PR #1179, opened 2018-10-04
Replacing remaining Icons by TextIcons

RobinCsl:
I removed entirely `NumberControl.js` as it is not used anywhere, bad idea?


tbergq:
Sounds like a good idea. We don't need dead code.

PR #1183, opened 2018-10-04
Removing orbit=true to TextIcons; removing spfont

RobinCsl:


tbergq:
so @ferrannp should we now deploy an new Maven artifact, since we remove spfont?

Issue #1129, opened 2018-09-26
Add more docs on Android


ferrannp:
After #1124 is merged.

- ~Init bridge in advance~. Done 4abee943b7ae9bb5c27338a920eece3c6087eb63
- ~Share bridge~. Done 4abee943b7ae9bb5c27338a920eece3c6087eb63
- Add more on how to test a new native dependency (commenting code out)
- Add docs on how to test a dependency form react-native-native-modules

PR #1192, opened 2018-10-05
Add usage-in-real-app.md for Android + initial hotel props fixes

ferrannp:
First two points of https://github.com/kiwicom/mobile/issues/1129 + fixes on initial props of Hotels (some were left hardcoded).

### DIFF
--- a/android/playground/src/main/java/com/kiwi/rnandroidplayground/HotelsData.kt
+++ b/android/playground/src/main/java/com/kiwi/rnandroidplayground/HotelsData.kt
@@ -1,17 +1,30 @@
 package com.kiwi.rnandroidplayground
 
+import android.util.Log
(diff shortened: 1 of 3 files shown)


android/playground/src/main/java/com/kiwi/rnandroidplayground/HotelsData.kt
```diff
@@ -8,8 +8,8 @@ object HotelsData {
   fun getInitialProperties(): RNHotelsInitialProperties {
     return RNHotelsInitialProperties(
         "en",
-        "2018-12-01",
-        "2018-12-25",
+        "EUR",
+        "",
         "2018-12-20",
         "2018-12-25",
```
lukewalczak: Maybe change hardcoded dates for `Today` and `Tomorrow` dates. After 25.12 there will be `no results`.

PR #1194, opened 2018-10-05
Hotels: Set delayPressIn for photo in detail screen

tbergq:
use lowRes for hotel room image instead of thumbnail


app/hotels/src/singleHotel/header/Header.js
```diff
@@ -54,7 +54,7 @@ export class Header extends React.Component<Props> {
     const photosCount = idx(hotel, _ => _.photos.edges.length) || 0;
 
     return (
-      <Touchable onPress={this.openGallery}>
+      <Touchable onPress={this.openGallery} delayPressIn={40}>
```
RobinCsl: Why do you add this delay?

tbergq: When we press a button it changes opacity to indicate that we are pressing it. When you put your finger on it and start scrolling, it should not show this opacity, so this delay should be sufficient to prevent that behavior.

tbergq: NB, I added a couple of more of them

PR #1198, opened 2018-10-09
Fix: menu group titles are bold and uppercase

RobinCsl:


app/MMB/src/scenes/tripServices/TripServiceRefreshContainer.js
```diff
@@ -78,12 +78,7 @@ class TripServiceRefreshContainer extends React.Component<Props, State> {
       {/* TODO: ordered services - how does it work here? */}
 
       <TitledMenuGroup
-        title={
-          <Translation
-            id="mmb.trip_services.general_services"
-            textTransform="uppercase"
```
tbergq: Why are you removing this?

tbergq: Aha, I see

RobinCsl: I am just removing the `textTransform="uppercase"` as it is now applied to all similar components

PR #1201, opened 2018-10-09
Hotels: Fix status bar on gallery

tbergq:
Gallery and payments as screens instead of new modal

### DIFF
--- a/app/hotels/src/navigation/singleHotel/SingleHotelStack.js
+++ b/app/hotels/src/navigation/singleHotel/SingleHotelStack.js
@@ -1,15 +1,12 @@
 // @flow
 
-import * as React from 'react';
 import { Dimensions as RNDimensions } from 'react-native';
 import { withMappedNavigationAndConfigProps as withMappedProps } from 'react-navigation-props-mapper';
 import {
-  HeaderTitle,
   StackNavigator,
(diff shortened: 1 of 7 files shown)


app/hotels/src/navigation/singleHotel/SingleHotelStack.js
```diff
@@ -17,10 +17,40 @@ import GalleryGrid from '../../gallery/GalleryGrid';
...
+      },
+    },
+    GalleryGrid: {
+      screen: withMappedProps(GalleryGrid, AdditionalPropsInjector),
+    },
+    Payment: {
+      screen: withMappedProps(PaymentScreen),
+      navigationOptions: {
+        mode: 'modal',
+        headerTitle: (
```
RobinCsl: Can we move this to the PaymentScreen component?

tbergq: Probably yes

tbergq: But we will get this error 😄 

image

tbergq: @RobinCsl 🚀

PR #1207, opened 2018-10-10
Fix: Past trips now show from most recent to oldest

RobinCsl:


tbergq:
@RobinCsl you have some merge conflicts here

RobinCsl:
@tbergq 🚀

PR #1210, opened 2018-10-10
MMB: Download and unzip pkpass on android

tbergq:


app/MMB/src/scenes/tickets/boardingPasses/mobileTicket/WalletPassenger.js
```diff
@@ -39,7 +61,37 @@ class WalletPassenger extends React.Component<PropsWithContext> {
...
+    return `pkpasses/${this.props.bookingId}/${flightNumber}/${databaseId}`;
+  };
+
+  downloadPkpass = async () => {
+    try {
+      const url = idx(this.props.data, _ => _.url) || '';
+      const databaseId = idx(this.props.data, _ => _.passenger.databaseId);
+
+      if (this.props.flightNumber != null && databaseId != null) {
+        const folderPath = this.getFolderPath();
```
RobinCsl: So if the path is invalid because `databaseId` or `flightNumber` in `getFolderPath` is `''`, then you will just warn that it failed, right?

tbergq: Currently, nothing will happen

tbergq: There could be an else with an alert telling some data is missing 🤷‍♂️

tbergq: I added a couple of TODO's here

PR #1215, opened 2018-10-11
Add: (partially) Connecting Payment Form

RobinCsl:
Displaying Card image conditionally on card number

@tbergq I believe I should add tests too, so I'll tag it WIP


app/MMB/src/scenes/tripServices/insurance/paymentScene/CardImage.js
```diff
@@ -0,0 +1,61 @@
...
+type Props = {|
+  +cardType: CardType,
+|};
+
+const CardImage = ({ cardType }: Props) => {
+  let source;
+  switch (cardType) {
+    case 'MASTERCARD': {
+      source = require('./images/MasterCardLogo.imageset/mastercard.png');
+      break;
```
tbergq: Does this really work?

https://github.com/facebook/react-native/issues/6391#issuecomment-194568185

app/MMB/src/scenes/tripServices/insurance/paymentScene/InsurancePaymentContext.js
```diff
@@ -0,0 +1,207 @@
...
+});
+
+type Props = {|
+  +children: React.Node,
+|};
+
+type State = {|
+  cardNumber: string,
+  cardType: string,
+  expiryDate: { month: string, year: string },
```
tbergq: should also be strict type

app/MMB/src/scenes/tripServices/insurance/paymentScene/InsurancePaymentContext.js
```diff
@@ -0,0 +1,207 @@
...
+  securityCode: string,
+  cardHoldersName: string,
```
tbergq: `cardholderName` maybe?

app/MMB/src/scenes/tripServices/insurance/paymentScene/InsurancePaymentContext.js
```diff
@@ -0,0 +1,207 @@
...
+  | 'MASTERCARD'
+  | 'VISA'
+  | 'MAESTRO'
+  | 'AMERICAN_EXPRESS'
+  | 'DISCOVER'
+  | 'DINERS_CLUB'
+  | 'MIR'
+  | 'UNKNOWN';
+
+function detectCardType(cardNumber: string, strict: boolean = false): CardType {
```
tbergq: could you add some test for this?

app/MMB/src/scenes/tripServices/insurance/paymentScene/CardImage.js
RobinCsl: Well, the image was changing when I tried different card numbers

RobinCsl:
I added some tests to verify that it detects the right card. 
The current implementation is just for displaying the right image; it does not check if the card is valid.

PR #1216, opened 2018-10-12
CodePush for Android

ferrannp:
- [x] Remove linking step from `configureApplication` --> When you link a library it will be already pushed and with the structure we have, it is dangerous to call `link` without specifying the package
- [x] Create Android SNAPSHOT to Trinerdis
- [x] Set up stage on Android
- [x] Hide deploying keys
- [x] `--gradle-file` to match `rnkiwimobile` version range by default
- [x] Use `overrideAppVersion`. See https://github.com/Microsoft/react-native-code-push/issues/1432
- [x] Remove buildAndroidJavascript.sh and tweak Gitlab CI
- [x] Create scripts in package.json
- [x] Be able to pass the key when instantiating the library


scripts/configureApplication.js
```diff
@@ -52,9 +51,6 @@ API_KEY_GOOGLE_MAPS=${vault.API_KEY_GOOGLE_MAPS}
 log('Setting up ENV variables...');
 fs.writeFileSync(path.join(__dirname, '..', '.env'), envTemplate);
 
-log('Linking all native dependencies...');
-child_process.execSync('yarn react-native link');
```
ferrannp: Figure out what to do instead of this. Probably only linking assets.

scripts/codePush.js
```diff
@@ -0,0 +1,35 @@
...
+
+const targetBinaryVersion =
+  packageJson['rnkiwimobile']['code-push-target-binary-version'];
+
+const entryFile = path.join(__dirname, '..', 'app/native.js');
+
+function processCommand() {
+  const flag = process.argv[2];
+  if (!flag || !flag.match(/^(--production|--staging)$/)) {
+    throw new Error('You need to pass either --staging or --production flag.');
```
tbergq: maybe staging could be default, since we most of the time will push to staging?

app/native.js
```diff
@@ -26,3 +27,6 @@ AppRegistry.registerComponent(
 );
 
 // ManageMyBookings
+
+// CodePush
```
tbergq: I think we can delete these comments

PR #1221, opened 2018-10-12
Fix: Single Hotel Details - see picture detail instead of gallery

RobinCsl:


tbergq:
Who asked for this change? I really don't like it. 

First of all, if I have the thumbnails, I can find and enlarge the fotos that interests me. Probably I would want to see 2 or 3 out of 120 photos. 

Second, it does not make sense to directly open a modal with images, then when I close it, I am at the grid. I would expect to end up where I was when I clicked the photo.

RobinCsl:
So after our discussion from this morning, what do you think we should be doing about this one?

tbergq:
🤷‍♂️ One one replied to it yet. I would say two options

1. Leave it as is (my preferred way)
2. Open the modal (🙈) then when you close it, you are on hotel detail screen, and just delete gallery screen.

RobinCsl:
Okay, let's wait for a reply from someone.

RobinCsl:
After discussion with designer: 
> ​​Well okay, let’s keep gallery view
So closing this issue.

PR #1234, opened 2018-10-17
Fix: drop shadow to increment-decrement room number button

RobinCsl:


app/hotels/src/singleHotel/roomPicker/IncrementDecrement.js
```diff
@@ -71,10 +70,11 @@ const styles = StyleSheet.create({
     flex: 1,
     borderRadius: 6,
     borderColor: defaultTokens.paletteInkLighter,
-    borderWidth: 1,
+    borderWidth: 0.5,
+    backgroundColor: defaultTokens.paletteWhite,
     height: 44,
     paddingHorizontal: 12,
-    shadowColor: Color.dark.$10,
```
tbergq: Btw, I think this is the only usage of that color. Can you delete it from the color file?

tbergq: Nvm, it is used `BottomSheet.js`

Issue #1229, opened 2018-10-17
Tablet view of Hotel Search Results: add map like on mobile


RobinCsl:
Tablet view of hotel search results has changed in the design; it is possible to switch between list view and map view, as is already possible on mobile.

PR #1237, opened 2018-10-19
Add: Hotels map available on tablet

RobinCsl:
Small refactor to avoid 5-level prop-drilling

/!\ We need to agree on a Hotels Context /!\


RobinCsl:
Fixes #1229

PR #1255, opened 2018-10-26
Add gitlab-ci

tbergq:


.gitlab-ci.yml
```diff
@@ -0,0 +1,28 @@
...
+  - yarn || true # Don't blow up when 'yarn' doesn't exist
+
+
+test_ci:
+  stage: test_ci
+  script:
+    - yarn test-ci
+
+release:
+  stage: test-ci
```
mrtnzlml:
```suggestion
  stage: release
```

.gitlab-ci.yml
```diff
@@ -0,0 +1,28 @@
...
+  - release
+
+cache:
+  key: ${CI_COMMIT_REF_SLUG}
+  paths:
+    - node_modules/
+  policy: pull
+
+before_script:
+  - yarn || true # Don't blow up when 'yarn' doesn't exist
```
mrtnzlml: I would personally get rid of this. You need it only in one job anyway.

.gitlab-ci.yml
```diff
@@ -0,0 +1,28 @@
...
+
+stages:
+  - test_ci
+  - release
+
+cache:
+  key: ${CI_COMMIT_REF_SLUG}
+  paths:
+    - node_modules/
+  policy: pull
```
mrtnzlml: I am not sure about `pull` only. Does it mean that you will not push `node_modules` after the install? Then I don't see the point of this cache. 🙂

PR #1266, opened 2018-10-30
Hotels: Replace context arrow functions with class functions

tbergq:


app/hotels/src/HotelsContext.js
```diff
@@ -118,19 +118,21 @@ export default { Consumer, Provider };
...
-    };
-
-    // $FlowExpectedError: We need to pass on the navigationOptions if any, flow does not know about it, but a react component might have it
-    if (Component.navigationOptions) {
-      WithHotelsContext.navigationOptions = Component.navigationOptions;
+      render() {
+        return <Consumer>{this.mapStateToProps}</Consumer>;
+      }
     }
+    // $FlowExpectedError: We need to pass on the navigationOptions if any, flow does not know about it, but a react component might have it
```
RobinCsl: This should probably be removed.

tbergq: Did you try to remove it? I think it is still needed 😊

RobinCsl: screen shot 2018-10-31 at 09 00 22
No error it seems.

tbergq: Oh, sorry, I missed that there was one below as well, this can obviously be removed, good catch.

PR #1271, opened 2018-10-31
 Hotels: Get original room id from graphql query

tbergq:
And revert the reverted commit 🤕


app/hotels/src/singleHotel/__tests__/__snapshots__/PaymentScreen.test.js.snap
```diff
@@ -1,5 +1,5 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`creates correct URL 1`] = `"http://test.url.lol&checkin=1970-01-02&interval=2&nr_rooms_7709411_91461863_1_1_0=1&nr_rooms_7709404_91461863_0_1_0=2&label=kiwi-ios-react-rn-test&lang=en&selected_currency=EUR"`;
```
RobinCsl: So you don't have the room number anymore?

tbergq: It comes from graphql instead. Since we replaced deprecated originalId with relay masked id, we need the graphql query to unmask it. Just like for the hotel id

PR #1279, opened 2018-11-02
Add: e2e test for gallery modal; e2e setup improved

RobinCsl:


tbergq:
Maybe we at some point should start thinking about having these tests in separate files? Like Hotels.gallery, Hotels.paymentScreen?

Not saying now, but.., what do you think?

RobinCsl:
Absolutely. I guess I can do this now. And have all these functions in a separate helpers file.

RobinCsl:
@tbergq 🚀

PR #1286, opened 2018-11-05
Improvement: Prevent selecting more rooms than there are guests

RobinCsl:
I believe this prop-drilling should be improved.

### DIFF
--- a/app/hotels/src/singleHotel/roomPicker/__tests__/RoomPicker.test.js
+++ b/app/hotels/src/singleHotel/roomPicker/__tests__/RoomPicker.test.js
@@ -15,6 +15,7 @@ it('renders as expected', () => {
       selectableCount={1}
       increment={VoidAction}
       decrement={VoidAction}
+      disabled={false}
     />,
   );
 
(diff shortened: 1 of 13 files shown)


app/hotels/src/singleHotel/roomPicker/__tests__/__snapshots__/RoomPicker.test.js.snap
```diff
@@ -30,6 +32,7 @@ exports[`renders as expected 2`] = `
 exports[`renders as expected 3`] = `
 <IncrementDecrement
   decrement={[Function]}
+  disabled={false}
```
tbergq: maybe we could add a test where it is disabled?

app/hotels/src/singleHotel/roomPicker/__tests__/RoomPicker.test.js
```diff
@@ -35,6 +37,18 @@ it('renders as expected', () => {
       selectableCount={2}
       increment={VoidAction}
       decrement={VoidAction}
+      disabled={false}
+    />,
+  );
+
+  PlaygroundRenderer.render(
```
RobinCsl: @tbergq

PR #1287, opened 2018-11-05
Hotels: Add stay 22 detail query support

tbergq:
Waiting for gql PR to be merged first

### DIFF
--- a/app/hotels/src/navigation/allHotels/SearchResultsScreen.js
+++ b/app/hotels/src/navigation/allHotels/SearchResultsScreen.js
@@ -23,7 +23,9 @@ import {
 } from './SearchResultsContext';
 import SingleHotelContainer from '../../singleHotel/SingleHotelContainer';
 import type { RoomsConfiguration } from '../../singleHotel/AvailableHotelSearchInput';
-import SingleHotelContext from '../singleHotel/SingleHotelContext';
+import SingleHotelContext, {
+  type ApiProvider,
+} from '../singleHotel/SingleHotelContext';
 
 type Props = {|
   +navigation: NavigationType,
@@ -35,6 +37,7 @@ type Props = {|
   +lastNavigationMode?: 'present' | 'push',
   +setResultType: (show: ResultType) => void,
   +show: ResultType,
+  +cityId?: string,
 |};
 
 const noop = () => {};
@@ -134,26 +137,33 @@ class SearchResultsScreen extends React.Component<Props> {
     }
   };
 
-  render = () => (
-    <SingleHotelContext.Provider
-      hotelId={''}
-      checkin={new Date(this.props.checkin)}
-      checkout={new Date(this.props.checkout)}
-      roomsConfiguration={this.props.roomsConfiguration}
-    >
-      <LayoutDoubleColumn
-        menuComponent={
-          <View style={styles.container}>
-            <NewAllHotels />
-            <View style={styles.button}>
-              <CloseButton onPress={this.onClosePress} />
+  render() {
+    // Stay 22 only supports searching with gps coordinates, if we get passed a cityId from native
+    // we are using booking.com, if we don't get cityId, we should have coordinates, and we are using stay22 provider
+    const apiProvider: ApiProvider =
+      this.props.cityId == null ? 'stay22' : 'booking';
+    return (
+      <SingleHotelContext.Provider
+        hotelId={''}
+        checkin={new Date(this.props.checkin)}
+        checkout={new Date(this.props.checkout)}
+        roomsConfiguration={this.props.roomsConfiguration}
+        apiProvider={apiProvider}
+      >
+        <LayoutDoubleColumn
+          menuComponent={
+            <View style={styles.container}>
+              <NewAllHotels />
+              <View style={styles.button}>
+                <CloseButton onPress={this.onClosePress} />
+              </View>
             </View>
-          </View>
-        }
-        containerComponent={<SingleHotelContainer goBack={noop} />}
-      />
-    </SingleHotelContext.Provider>
-  );
+          }
+          containerComponent={<SingleHotelContainer goBack={noop} />}
+        />
+      </SingleHotelContext.Provider>
+    );
+  }
 }
 
 const styles = StyleSheet.create({
--- a/app/hotels/src/navigation/singleHotel/SingleHotelContext.js
+++ b/app/hotels/src/navigation/singleHotel/SingleHotelContext.js
@@ -4,12 +4,15 @@ import * as React from 'react';
 
 import type { RoomConfigurationType } from '../../HotelsContext';
 
+export type ApiProvider = 'booking' | 'stay22';
+
 const defaultState = {
   hotelId: '',
   checkin: new Date(),
   checkout: new Date(),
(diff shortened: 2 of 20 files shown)


app/hotels/src/navigation/singleHotel/SingleHotelContext.js
```diff
@@ -41,6 +46,7 @@ class Provider extends React.Component<Props, State> {
       checkin: props.checkin,
       checkout: props.checkout,
       roomsConfiguration: props.roomsConfiguration,
+      apiProvider: props.apiProvider ?? 'booking',
```
RobinCsl: Oooh, nice :)

app/hotels/src/singleHotel/SingleHotelContainer.js
```diff
@@ -81,32 +82,49 @@ type Props = {|
...
-      checkout={checkout}
-      roomsConfiguration={roomsConfiguration}
-      {...this.props}
-    />
-  );
+    apiProvider,
+  }: SingleHotelState) => {
+    if (apiProvider === 'booking') {
+      return (
+        <SingleHotelContainer
```
RobinCsl: So we are passing the same props here, right? Just the wrapper component is changing. Would it be possible to have the logic do something like:
```
let Component = Stay22SingleHotel;
if (apiProvider === 'booking') {
  Component = SingleHotelContainer;
}
return (
  <Component {...props} />
);
```

app/hotels/src/singleHotel/Stay22SingleHotel.js
```diff
@@ -0,0 +1,72 @@
...
+  +hotelId: string,
+  +checkin: Date,
+  +checkout: Date,
+  +roomsConfiguration: RoomConfigurationType,
+|};
+
+export default class Stay22SingleHotel extends React.Component<Props> {
+  getNumberOfGuests = () => {
+    return this.props.roomsConfiguration.reduce((acc, curr) => {
+      return acc + curr.adultsCount + curr.children.length;
```
RobinCsl: It looks like we have this function popping up in a few places. I know it is simple, but maybe extracting it would be better if ever the object structure changes.

tbergq: Yes, absolutely, I did not remember already doing this function. I will extract it to the context.

tbergq:
@RobinCsl please review again

app/hotels/src/singleHotel/roomList/RoomRow.js
```diff
@@ -75,17 +74,14 @@ export class RoomRow extends React.Component<Props> {
...
-      _ => _.room.roomPhotos[0].lowResUrl,
-    );
-    const amount = idx(availableRoom, _ => _.minimalPrice.amount) || null;
-    const currency = idx(availableRoom, _ => _.minimalPrice.currency) || null;
-    const selectableCount =
-      idx(availableRoom, _ => _.incrementalPrice.length) || 0;
-    const id = idx(availableRoom, _ => _.id) || '';
-    const selectedCount = idx(this.props.selected, _ => _[id]) || 0;
-    const room = idx(availableRoom, _ => _.room);
+    const thumbnailUrl = availableRoom?.room?.roomPhotos?.[0]?.highResUrl;
```
RobinCsl: thumbnail is given in high resolution: is that intended?

tbergq: yes, testers complain about thumbNailUrl to be to bad quality, I upgraded to lowResUrl which was fine, however stay22 only provides thumbNail and highRes url

app/hotels/src/navigation/allHotels/SearchResultsScreen.js
```diff
@@ -134,26 +137,31 @@ class SearchResultsScreen extends React.Component<Props> {
...
-    >
-      <LayoutDoubleColumn
-        menuComponent={
-          <View style={styles.container}>
-            <NewAllHotels />
-            <View style={styles.button}>
-              <CloseButton onPress={this.onClosePress} />
+  render() {
+    const apiProvider: ApiProvider =
+      this.props.cityId == null ? 'stay22' : 'booking';
```
RobinCsl: I don't really understand this logic.

app/hotels/src/singleHotel/SingleHotelContainer.js
```diff
@@ -13,6 +13,7 @@ import HotelsContext, {
 import SingleHotelContext, {
   type State as SingleHotelState,
 } from '../navigation/singleHotel/SingleHotelContext';
+import Stay22SingleHotel from './Stay22SingleHotel';
```
RobinCsl: We could even lazy load it using `Suspense` and `lazy`; what do you think? Does it make sense in RN?

app/hotels/src/navigation/allHotels/SearchResultsScreen.js
tbergq: I added a comment explaining

app/hotels/src/singleHotel/roomList/RoomRow.js
RobinCsl: Could we document this somewhere?

app/hotels/src/singleHotel/SingleHotelContainer.js
tbergq: > Does it make sense in RN?

No 😄

app/hotels/src/singleHotel/roomList/RoomRow.js
tbergq: I will check with the provider if they can add lowResUrl also, so hopefully we can change it back. 

another possible approach is to use lowResUrl, and fallback to highResUrl if lowResUrl is null.

app/hotels/src/singleHotel/roomList/RoomRow.js
```diff
@@ -52,22 +51,26 @@ export class RoomRow extends React.Component<Props> {
...
     const { availableRoom } = this.props;
-    const photosArray = idx(availableRoom, _ => _.room.roomPhotos) || [];
+    const photosArray = availableRoom?.room?.roomPhotos ?? [];
 
-    const photos = photosArray.map(photo => ({
-      key: idx(photo, _ => _.id) || '',
-      highRes: idx(photo, _ => _.highResUrl) || '',
-      lowRes: idx(photo, _ => _.lowResUrl) || '',
-    }));
+    const photos = photosArray.map(photo => {
```
RobinCsl:
```suggestion
    const photos = photosArray.map(photo => {
      const key =  photo?.id ?? '';
      const highRes = photo?.highResUrl ?? '';
      const lowRes = photo?.lowResUrl ?? highRes; // Fallback to highResUrl since stay22 do not provide lowResUrl
      
      return {
        key,
        highRes,
        lowRes,
      };
    });
```

RobinCsl: Sorry, that does not seem to replace the lines below... But I think it's better that all the logic is in the `return` statement or before the `return` statement, and not mixed like it is now. Thoughts?

tbergq: makes sense

PR #1290, opened 2018-11-06
Adding provider to RNLogging module

RobinCsl:
I've set it to WIP because I absolutely need feedback from @ferrannp and @lukewalczak as I don't really know how to test that what I have done works.


app/hotels/src/singleHotel/PaymentScreen.js
```diff
@@ -57,12 +57,18 @@ export class PaymentScreen extends React.Component<PaymentParameters> {
     };
   };
   componentDidMount = () => {
-    Logger.ancillaryDisplayed(Logger.Type.ANCILLARY_STEP_PAYMENT);
+    Logger.ancillaryDisplayed(
+      Logger.Type.ANCILLARY_STEP_PAYMENT,
+      Logger.Provider.ANCILLARY_PROVIDER_BOOKINGCOM,
```
tbergq: This could be wrong. I think you need to wait for my PR, then you can fetch the apiProvider from the context

RobinCsl: For now, it seems like the PaymentScreen is only for Booking.com (the graphql query is 
```
graphql`
          query PaymentScreenQuery(
            $hotelId: ID
            $roomConfig: [RoomConfigInput]
          ) {
            hotelPaymentUrls(hotelId: $hotelId, roomConfig: $roomConfig) {
              bookingComPaymentUrl
            }
          }
        `
```

tbergq: Yeah, you are right, I will improve this when I add support for stay22.

ios/RNModules/RNLogging/RNLogging/RNLoggingModule.h
```diff
@@ -2,8 +2,8 @@
 
 @protocol RNLogger <NSObject>
 
-- (void)ancillaryDisplayed:(nullable NSString *)type;
-- (void)ancillaryPurchased:(nullable NSString *)type;
+- (void)ancillaryDisplayed:(nullable NSString *)type andProvider:(nullable NSString *)provider;;
```
pistelak: Please, do not use `and` in name of the last parameter. Although, "and" sounds good in this case, it makes addition of next parameter harder. 

More info about naming: https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/NamingMethods.html 

> Don’t use “and” to link keywords that are attributes of the receiver. Although “and” may sound good in this example, it causes problems as you create methods with more and more keywords.

RobinCsl: Thanks for that. I've changed it to just `provider`.

RobinCsl:
@tbergq 🚀

tbergq:
@RobinCsl, it is still with WIP label. Should I merge it anyways?

RobinCsl:
Bumped the version of RNKiwiMobile and of RNNativeModule as per @ferrannp's recommandation

PR #1294, opened 2018-11-07
Hotels: Fix gallery page for stay 22

tbergq:


app/hotels/src/singleHotel/header/Header.js
```diff
@@ -34,13 +33,20 @@ export type Props = {
...
-    const photosEdges = idx(hotel, _ => _.photos.edges) || [];
-    const images = photosEdges.map(edge => ({
-      key: idx(edge, _ => _.node.id) || '',
-      lowRes: idx(edge, _ => _.node.lowResUrl) || '',
-      highRes: idx(edge, _ => _.node.highResUrl) || '',
-    }));
-    const hotelName = idx(hotel, _ => _.name);
+    const photosEdges = hotel?.photos?.edges ?? [];
+
+    const images = photosEdges.map(edge => {
```
RobinCsl: Is it the second time we write this logic? Should we extract it?

tbergq: It is not the exact same, this is a connection, the other one is not. 

I could extract it, but then I would have to map this connection, and then pass it to the extracted function to have it mapped yet again 🤷‍♂️

RobinCsl: Okay, then no need

Issue #1274, opened 2018-10-31
Create script to link assets


ferrannp:
Since this PR https://github.com/kiwicom/mobile/pull/1216 we will not call `react-native link` so we need a sript to link only assets. We can check the link repo to mimic it.

PR #1298, opened 2018-11-07
Patching react-native-code-push for linking assets

ferrannp:
Fixes https://github.com/kiwicom/mobile/issues/1274.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -155,6 +155,7 @@ android {
 }
 
 dependencies {
+    compile project(':react-native-code-push')
     implementation project(':react-native-zip-archive')
     implementation project(':react-native-share')
     implementation project(':react-native-pdf')
--- a/ios/reactNativeApp.xcodeproj/project.pbxproj
+++ b/ios/reactNativeApp.xcodeproj/project.pbxproj
@@ -5,7 +5,6 @@
 	};
 	objectVersion = 46;
 	objects = {
-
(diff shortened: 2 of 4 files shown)


android/app/build.gradle
```diff
@@ -155,6 +155,7 @@ android {
 }
 
 dependencies {
+    compile project(':react-native-code-push')
```
ferrannp: In this way it will detect it's already "linked" even if `app` does not use it.

tbergq: shouldn't it be `implemenation` and not `compile`?

ios/reactNativeApp.xcodeproj/project.pbxproj
```diff
@@ -5,7 +5,6 @@
 	};
 	objectVersion = 46;
 	objects = {
-
```
tbergq: Just out of curiosity, why do these changes happen from time to time?

ferrannp: Yeah I don't really know, it looks it differs depending who makes them... I guess we can investigate it in the future but yeah...

android/app/build.gradle
ferrannp: Yeah but link puts `compile` still. I can change it manually however as this is the `app` it does not really matter.

tbergq: Ok, then I'll allow it 🙂

PR #1301, opened 2018-11-08
Hotels: Replace idx with optional chaining in allHotels folder

tbergq:


app/hotels/src/allHotels/HotelTitle.js
```diff
@@ -16,15 +15,15 @@ type Props = {|
 |};
 
 function HotelTitle({ data }: Props) {
-  const hotelName = idx(data, _ => _.name);
-  const hotelStars = idx(data, _ => _.rating.stars);
+  const hotelName = data.name ?? '';
+  const hotelStars = data.rating?.stars;
 
-  const price = data.price || null;
+  const price = data.price ?? null;
```
RobinCsl: I am wondering if `??` makes sense here. My understanding is that `price` is `null` if `data.price` is `null` or `undefined` with `??`, when any falsy value would make `price` `null` with `||`. Correct?

tbergq: Yes, with data.price || null, const price would be null if price was 0, so that is definitely not correct.

Also, since the price component expects type ?Price, it does not matter if it is null or undefined. So yes, I will remove the nullish coalescence here

tbergq: 🚀

PR #1312, opened 2018-11-12
Shared: Add withContext component

tbergq:


packages/shared/src/withContext/withContext.js
```diff
@@ -0,0 +1,53 @@
...
+       * props are changing
+       */
+      shouldComponentUpdate(nextProps: T) {
+        return !isEqual(select(nextProps), select(this.props));
+      }
+
+      render() {
+        const children = React.cloneElement(this.props.children, {
+          ...(this.props: Object),
+          testID: 'withContext',
```
RobinCsl: Will this overwrite the `testID` we might be using for e2e testing for instance?

tbergq: You are right! I should do this differently

tbergq:
> That means there are other contexts we could refactor with this utility.

yes, that was the idea.

PR #1315, opened 2018-11-12
Hotels: Use withHotelsContext

tbergq:


app/hotels/src/HotelsFilterContext.js
```diff
@@ -98,9 +99,16 @@ class Provider extends React.Component<Props, State> {
...
+  render() {
+    return (
+      <ContextProvider value={this.state}>
+        {this.props.children}
+      </ContextProvider>
+    );
+  }
 }
 
+export const withHotelsFilterContext = (select: State => Object) =>
```
RobinCsl: Would it be possible to say that `select` returns an object with keys a subset of the keys of `State`?

RobinCsl: Well, I guess you might want to rename sometimes...

tbergq: it does not have to be a subset of state e.g.

```
const select = ({ currency, checkin, checkout, currentSearchStats }) => ({
  currency,
  checkin,
  checkout,
  currentSearchStats,
  daysOfStay: calculateDaysOfStay(checkin, checkout),
});
```

app/hotels/src/filter/price/PriceFilter.js
```diff
@@ -188,28 +188,12 @@ const calculateDaysOfStay = (checkin, checkout) => {
...
-        daysOfStay={daysOfStay}
-      />
-    );
-  };
+const select = ({ currency, checkin, checkout, currentSearchStats }) => ({
+  currency,
+  checkin,
+  checkout,
+  currentSearchStats,
+  daysOfStay: calculateDaysOfStay(checkin, checkout),
```
RobinCsl: Now, that is elegant! I wonder however if `select`'s scope should just be about selecting items from the State of the context, and not provide additional props which can be inferred from it. Maybe this kind of computations should stay within the component. WDYT?

app/hotels/src/map/singleHotel/SingleHotelMapScreen.js
```diff
@@ -61,19 +63,18 @@ class SingleHotelMapScreen extends React.Component<Props> {
...
-      {this.renderQuery(currency, roomsConfiguration)}
-    </SingleHotelContext.Consumer>
-  );
-
   render() {
     return (
-      <HotelsContext.Consumer>
-        {this.renderHotelsContext}
-      </HotelsContext.Consumer>
+      <SingleHotelContext.Consumer>
```
RobinCsl: Could this context not be extracted too?

app/hotels/src/singleHotel/HotelDetailScreen.js
```diff
@@ -220,30 +219,23 @@ type Props = {|
...
-      width={width}
-      getGuestCount={getGuestCount}
-    />
+  renderInner = ({ width }) => (
+    <HotelDetailScreen {...this.props} width={width} />
   );
 
   render() {
-    return <Dimensions.Consumer>{this.renderDimensions}</Dimensions.Consumer>;
+    return <Dimensions.Consumer>{this.renderInner}</Dimensions.Consumer>;
```
RobinCsl: What about `withDimensionContext`?

app/hotels/src/filter/price/PriceFilter.js
tbergq: I don't think it makes any difference in this case, but I don't think we should limit it. 

There could be a use case if in various class functions we would use daysOfStay props, it is better to have it generated and passed in like this I guess.

app/hotels/src/map/singleHotel/SingleHotelMapScreen.js
tbergq: Step by step, I am also considering merging those two contexts into onw

app/hotels/src/singleHotel/SingleHotelContainer.js
```diff
@@ -103,21 +93,25 @@ export default class SingleHotelContainerWithContext extends React.Component<
       checkout,
       roomsConfiguration,
       hotelId,
-      currency,
+      currency: this.props.currency,
```
RobinCsl: This is maybe not necessary if this lives in `this.props` since you are using `...this.props` below.

app/hotels/src/singleHotel/HotelDetailScreen.js
tbergq: Next PR

app/hotels/src/singleHotel/SingleHotelContainer.js
tbergq: done

app/hotels/src/filter/price/PriceFilter.js
RobinCsl: True, and even that logic could be integrated into the context.

PR #1299, opened 2018-11-08
Added new logs

pistelak:
- https://jira.kiwi.com/browse/MOBILE-929: 

 * Hotels: Results displayed (searchQuery or params)
 * Hotels: Filter tag set
 * Hotels: Detail: Opened
 * Hotels: Detail: Description expanded
 * Hotels: Detail: Map opened
 * Hotels: Detail: Room selected (hotelId, roomType)
 * Hotels: Gallery opened (type: room/hotel)
 * Hotels: Book now pressed (hotelId, rooms, guests, price, formattedPrice)
 * Hotels: Detail: Abandoned

PR #1317, opened 2018-11-12
Feature: Improving logging module for Hotels

RobinCsl:
This is pending on #1299 
It will also be necessary to bump the RN native modules and RNKiwiMobile versions, and probably do something with CodePush too.

### DIFF
--- a/android/dependencies.gradle
+++ b/android/dependencies.gradle
@@ -29,6 +29,6 @@ ext {
     testRunnerVersion = "1.0.2"
     espressoVersion = "3.0.2"
 
-    loggingVersion = "0.2.4"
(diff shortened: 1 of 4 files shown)


RobinCsl:
@ferrannp @Blackhex Could you please have a look at this PR? Thank you!

Blackhex:
I haven't check if it complies with the actual logs definition but from the technical point of view it seems OK.

tbergq:
Is this ready to be merged
@RobinCsl @ferrannp @Blackhex

RobinCsl:
I think I wanted one last review by @ferrannp but forgot to ask for it 😊, and then we should be able to merge it.

PR #1322, opened 2018-11-13
Fix: Text on tags to lower case on Android

RobinCsl:


packages/localization/src/vocabularies/Hotels.js
```diff
@@ -25,7 +25,7 @@ export default {
   'hotels_search.all_hotels_map.allow_localization_or_search':
     'Please go to settings and allow localization, or go back and search for your desired destination',
   'hotels_search.date_picker.select_date': 'Select date',
-  'hotels_search.filter.free_cancellation': 'free cancellation',
+  'hotels_search.filter.free_cancellation': 'Free cancellation',
```
tbergq: actually is should be lower cased. Our `Rating` filter is the only one that is wrong.

RobinCsl: Oh really? Because in the design, they all seem to be capitalized.

PR #1332, opened 2018-11-14
Replace lifecycle arrow function with class function

tbergq:


app/hotels/README.md
```diff
@@ -1,7 +1,7 @@
 ```js
 import { HotelsStandalonePackage } from '@kiwicom/react-native-app-hotels';
 
-const render = () => <HotelsStandalonePackage
+const render() <HotelsStandalonePackage
```
RobinCsl: This does not work here, even though it's just a README file

tbergq: Hehe, I was pretty sure I undid that change 🙈

docs/common/create-new-react-native-feature.md
```diff
@@ -32,7 +32,7 @@ We are also working on `MMB` as a new standalone package, and hopefully there wi
 ```js
 type Props = {||};
 class CustomStandalonePackage extends React.Component<Props> {
-  render = () => <NavigationStack />;
+  render() <NavigationStack />;
```
RobinCsl: Likewise here

PR #1333, opened 2018-11-14
Hotels: Prepare payment screen for stay22 integration

tbergq:
Note that we now get to the webview with stay22, but the room is not selected because they provide us with wrong id's, I have asked them to correct this.


app/hotels/src/singleHotel/paymentScreen/PaymentWebView.js
```diff
@@ -0,0 +1,65 @@
...
+} from '../../HotelsContext';
+
+type Props = {|
+  +url: ?string,
+  +apiProvider: ApiProvider,
+|};
+
+export class PaymentWebView extends React.Component<Props> {
+  onNavigationStateChange = (event: WebViewStateChangeEvent) => {
+    if (!event.loading && event.url.includes('booking.com/confirmation')) {
```
RobinCsl: So it's still the same URL. This would probably need a comment to explain why it's the case? Otherwise it looks like it's a mistake.

app/hotels/src/singleHotel/paymentScreen/__tests__/BookingPaymentScreen.test.js
```diff
@@ -0,0 +1,43 @@
...
+
+it('creates correct URL', () => {
+  expect(
+    new BookingPaymentScreen({
+      hotelId: '1',
+      checkin: new Date(ONE_DAY),
+      checkout: new Date(3 * ONE_DAY), // 2 days later
+      roomConfig: [
+        {
+          roomId: '7709411_91461863_1_1_0',
```
RobinCsl: Is this a real ID? Or looks like a real ID? Maybe we should change that.

tbergq: Not really sure, but it has been in the code base since January

PR #1336, opened 2018-11-15
Hotels: Translate bedding info

tbergq:


app/hotels/src/singleHotel/roomList/BeddingInfo.js
```diff
@@ -19,16 +19,29 @@ type Props = {
...
+          <Translation
+            key={`key-${index}`}
+            id="single_hotel.bedding_info.or"
+          />,
+        );
+      }
+      beddingTranslation.push(
+        <Translation
+          key={`${type || ''}-${amount || ''}`}
+          passThrough={[amount, type].filter(value => value != null).join(' ')}
```
RobinCsl: I don't know if that always works for any language. But I guess that's good enough for now?

tbergq: Amount and type is being returned translated from booking.com based on the language parameter, so this should be fine.

RobinCsl: Sure, but I meant the order of the words, maybe in some languages you would have `bananas 2` and not `2 bananas` like we are used to.

PR #1337, opened 2018-11-15
Improvement: Translation Deployer deletes unused keys on PhraseApp an…

RobinCsl:
…d issues warnings

@tbergq I am not sure it works entirely as only 57 keys seem to be copied over to the test project, when about 100 should be copied as far as I know. I don't understand why.
Also, the find-in-files library seem to be opening too many files when we loop over the array of keys to find. I tried reducing the number of loops by having more hardcoded folders, etc. But I am not sure how we could improve this.


tbergq:
I deleted all keys from phraseApp test project, and ran the script in your branch. Then there is 120 keys, which is the number it should be. So that part is ok, however I do get a crash when trying to find files:

` UnhandledPromiseRejectionWarning: Error: EMFILE: too many open files, scandir 'app'`

packages/localization/src/Deployer.js
```diff
@@ -121,6 +136,64 @@ const createTranslation = async (keyId, translationString) => {
...
+    ...Object.keys(shared),
+  ];
+};
+
+const findUnusedKeys = async () => {
+  const unusedKeys = await Promise.all(
+    Object.keys(translations).map(async tr_key => {
+      const filesFoundWithQuotes = await findFiles(`'${tr_key}'`);
+      const filesFoundWithDoubleQuotes = await findFiles(`"${tr_key}"`);
+
```
tbergq: I think `findFiles(tr_key);` should be sufficient

it does not matter if it is in wrapped in single or double quotes

RobinCsl: Well, if you had:
`hello.world` - unused key
`hello.world.mate` - used key
then `hello.world` would not be detected as unused key.

packages/localization/src/Deployer.js
```diff
@@ -121,6 +136,64 @@ const createTranslation = async (keyId, translationString) => {
...
+  const accessibility = await findInFiles.find(key, 'packages/accessibility');
+  const assets = await findInFiles.find(key, 'packages/assets');
+  const config = await findInFiles.find(key, 'packages/config');
+  const navigation = await findInFiles.find(key, 'packages/navigation');
+  const orbit = await findInFiles.find(key, 'packages/orbit');
+  const pdf = await findInFiles.find(key, 'packages/pdf');
+  const playground = await findInFiles.find(key, 'packages/playground');
+  const relay = await findInFiles.find(key, 'packages/relay');
+  const rnmodules = await findInFiles.find(key, 'packages/rnmodules');
+  const shared = await findInFiles.find(key, 'packages/shared');
```
tbergq: Also maybe add folders to an array, and loop that array, I guess it scales better.

packages/localization/src/Deployer.js
tbergq: Ahh, of course

tbergq:
I am starting to think that it might be be better to do this manually 😞

packages/localization/src/Deployer.js
RobinCsl: That's what I did first, but then the performance was much worse somehow. It took twice the amount of time.

packages/localization/src/Deployer.js
```diff
@@ -121,6 +136,64 @@ const createTranslation = async (keyId, translationString) => {
...
+    ...Object.keys(playground),
+    ...Object.keys(relay),
+    ...Object.keys(rnmodules),
+    ...Object.keys(shared),
+  ];
+};
+
+const findUnusedKeys = async () => {
+  const unusedKeys = await Promise.all(
+    Object.keys(translations).map(async tr_key => {
```
tbergq: also, please on snake_case

packages/localization/src/Deployer.js
tbergq: Ok, then please add a comment as of why you chose this approach 😊

PR #1343, opened 2018-11-20
Hotels: Show offline screen for all hotels search results screen

tbergq:
Some issues/discoveries
• We might need to render different offline screens (since close does not make sense for non modal screens)
• the fixed position close button needs to be rendered inside the queryrenderer, if not it will render on top of offline screen buttons, or be missing on other error screens. Meaning we will not be able to close the modal on errors, which is bad UX.

Therefore a renderOfflineScreen passing the retry functions was the best solution I found. Also passing the footer component is a result of the above. 

Let me know if you can think of any improvements.

### DIFF
--- a/app/hotels/src/appRegistry/SingleHotelStandalonePackage.js
+++ b/app/hotels/src/appRegistry/SingleHotelStandalonePackage.js
@@ -22,6 +22,7 @@ type Props = {
   +onBackClicked: () => void,
   +dimensions: DimensionType,
   +version: string,
+  +lastNavigationMode?: string,
 };
 
 class SingleHotelStandAlonePackage extends React.Component<Props> {
@@ -45,6 +46,8 @@ class SingleHotelStandAlonePackage extends React.Component<Props> {
         roomsConfiguration={this.props.roomsConfiguration}
         hotelId={this.props.hotelId}
         apiProvider="booking" // Not sure if this is in use by native team, but for now, just support booking.com
+        lastNavigationMode={this.props.lastNavigationMode}
+        onBackClicked={this.props.onBackClicked}
       >
         <SingleHotelStack
           screenProps={screenProps}
--- a/packages/relay/index.js
+++ b/packages/relay/index.js
@@ -57,7 +57,9 @@ type CommitMutationConfig = {|
 export type QueryRendererProps = {|
   +query: string,
   +render: (props: Object) => React.Node,
+  +renderOfflineScreen?: ?(retry: () => void) => React.Node,
   +variables?: Object,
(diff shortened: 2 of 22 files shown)


app/hotels/src/appRegistry/SingleHotelStandalonePackage.js
```diff
@@ -45,6 +46,8 @@ class SingleHotelStandAlonePackage extends React.Component<Props> {
         roomsConfiguration={this.props.roomsConfiguration}
         hotelId={this.props.hotelId}
         apiProvider="booking" // Not sure if this is in use by native team, but for now, just support booking.com
```
RobinCsl: I know it's not part of this PR, but the comment seems to indicate this is not true anymore?

tbergq: No, this comment is definitely true. I have never seen this feature been used in the app, I asked if they are using it, but no reply. I Don't really think it is in use.

packages/relay/index.js
```diff
@@ -57,7 +57,9 @@ type CommitMutationConfig = {|
 export type QueryRendererProps = {|
   +query: string,
   +render: (props: Object) => React.Node,
+  +renderOfflineScreen?: ?(retry: () => void) => React.Node,
```
RobinCsl: Why are these optional props also nullable/undefined?

tbergq: This is the issue with exporting/sharing prop types.

Since they QueryRendererProps is used by PrivateApiRenderer, PublicApiRenderer and QueryRenderer. The prop is optional when passing it to PrivateApiRenderer and PublicApiRenderer, but they both pass it along to QueryRenderer, so it will always be passed to QueryRenderer, but it could be undefined.

packages/shared/src/offlineScreen/OfflineScreen.js
```diff
@@ -19,40 +20,51 @@ type Props = {|
...
-        <Translation id="shared.offline_screen.offline_title" />
-      </Text>
-      <Text style={styles.text}>
-        <Translation id="shared_offline_screen.offline_text" />
-      </Text>
-      <View style={styles.imageContainer}>
-        <Image source={OfflineImage} />
+    <SafeAreaView style={styles.flexItem}>
+      <View style={styles.container}>
+        <View style={styles.flexItem} />
```
RobinCsl: Is this one necessary?

tbergq: Yes, it is to create the offset above the text, so it looks ok on both tablet and mobile

PR #1356, opened 2018-11-26
Flow: Remove unnecessary flow suppressions

tbergq:


app/hotels/src/allHotels/AllHotelsSearchList.js
```diff
@@ -10,13 +10,12 @@ import type { AllHotelsSearchList as AllHotelsSearchListProps } from './__genera
 import { withHotelsContext, type HotelsContextState } from '../HotelsContext';
 
 type Props = {|
-  +data: AllHotelsSearchListProps,
+  +data: AllHotelsSearchListProps | Array<?empty>,
```
RobinCsl: Why is that not part of Relay's generated types?

RobinCsl: Yes, I guess my comment was not helpful. 😃

tbergq: Ahh, I missed the `why`

Well, I don't know why 🙈

PR #1361, opened 2018-11-28
Enhance: DropMarker is an image

### LINKED ISSUE #1225: [IMPROVEMENT] DropMarker component
The existing component does not look exactly like the design.
Besides there are some artefacts like `size` props from before; there should still be a way to make the `DropMarker` bigger.

RobinCsl:
Closes #1225


packages/shared/src/DropMarker/index.js
```diff
@@ -0,0 +1,37 @@
+// @flow strict
```
tbergq: please don't use the index.js name

I cant find any files using search for filenames if we have 100 files named index.js

tbergq:
Looks good, just please rename and the file, and it is good to go

RobinCsl:
@tbergq Should be fine now.

PR #1365, opened 2018-11-29
Hotels: Replace scroll view with flatlist in search result list

tbergq:


app/hotels/src/allHotels/RenderSearchResults.js
```diff
@@ -91,16 +91,19 @@ export class RenderSearchResults extends React.Component<Props> {
             transform: [{ translateY: this.listAnimation }],
           }}
         >
-          {/*  Note: it's not possible to use FlatList here because it's wrapped with ScrollView and it causes performance issues.*/}
-          <ScrollView contentContainerStyle={styles.content}>
```
tbergq: This comment is no longer true, since we now render the FilterStripe above the scroll view

PR #1372, opened 2018-12-03
Downgrade react-native-maps

### LINKED ISSUE #1371: Hotels freezes on Android
We are not sure since when but Hotels starts freezing only on Android. JS and native thread go down to 0fps.

@tbergq says that removing these native modules solves the problem:

```kt
new RNLoggingPackage(),
new RNTranslationManagerPackage(),
new RNCurrencyManagerPackage(),
```

ferrannp:
Fixes #1371.

The new version after so many months was dropping tones of Android fps and freezing the app.

### DIFF
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -9,7 +9,7 @@
     "@ptomasroos/react-native-multi-slider": "^1.0.0",
     "querystring": "^0.2.0",
     "react-native-image-progress": "^1.1.1",
-    "react-native-maps": "0.22.0",
+    "react-native-maps": "0.21.0",
     "react-native-modal": "^7.0.0",
     "react-native-read-more-text": "^1.1.0"
   },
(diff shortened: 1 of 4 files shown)


packages/shared/package.json
```diff
@@ -9,7 +9,7 @@
     "@ptomasroos/react-native-multi-slider": "^1.0.0",
     "querystring": "^0.2.0",
     "react-native-image-progress": "^1.1.1",
-    "react-native-maps": "0.22.0",
+    "react-native-maps": "0.21.0",
     "react-native-modal": "^7.0.0",
     "react-native-read-more-text": "^1.1.0"
   },
```
tbergq: please also update ´code-push-target-binary-version´

tbergq: nevermind, I did it

PR #1377, opened 2018-12-04
Add Stay22 options to Android playground

### LINKED ISSUE #1362: Expose stay22 version in playground
We needs to expose this version also in the playground now. 

Basically, just pass cityId = null and coordinates = {latitude: 12.123, longitude: 321.321} when starting `NewKiwiHotels`

- [x] ios
- [x] android

ferrannp:
Closes #1362.

@tbergq let me know if this is what we expect.

### DIFF
--- a/android/rnkiwimobile/src/main/java/com/kiwi/rnkiwimobile/hotels/RNHotelsInitialProperties.kt
+++ b/android/rnkiwimobile/src/main/java/com/kiwi/rnkiwimobile/hotels/RNHotelsInitialProperties.kt
@@ -6,9 +6,9 @@ data class RNHotelsInitialProperties(
     val checkin: String,
     val checkout: String,
     val cityName: String,
-    val cityId: String,
+    var cityId: String?,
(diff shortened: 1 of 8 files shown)


android/rnkiwimobile/src/main/java/com/kiwi/rnkiwimobile/hotels/RNHotelsInitialProperties.kt
```diff
@@ -6,9 +6,9 @@ data class RNHotelsInitialProperties(
     val checkin: String,
     val checkout: String,
     val cityName: String,
-    val cityId: String,
+    var cityId: String?,
     val roomsConfiguration: ArrayList<RNHotelsRoomsConfiguration>,
-    val hotelsCoordinates: RNHotelsCoordinates
+    val coordinates: RNHotelsCoordinates
```
tbergq: We would need to change version on `/.build/package.json` right?

Issue #1376, opened 2018-12-04
Patch react-native-maps in node_modules


ferrannp:
As we downgraded, in order for Android project to sync, we need to remove:

```xml
<uses-sdk android:minSdkVersion="16" />
```

from `package.json`.

PR #1379, opened 2018-12-05
Patching react-native-maps for Android Studio to be able to sync

ferrannp:
Fixes https://github.com/kiwicom/mobile/issues/1376.

### DIFF
--- a/scripts/configureApplication.js
+++ b/scripts/configureApplication.js
@@ -105,3 +105,27 @@ try {
 }
 
 log('Configuration complete!');
+
+// Needed before 0.22.0 for Android Studio to sync
+log('Patching react-native-maps');
+const androidManifestReactNativeMaps = fs.readFileSync(
+  path.join(
+    __dirname,
+    '..',
+    'node_modules/react-native-maps/lib/android/src/main/AndroidManifest.xml',
+  ),
+  'utf-8',
+);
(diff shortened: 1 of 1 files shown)


scripts/configureApplication.js
```diff
@@ -105,3 +105,27 @@ try {
...
+const androidManifestReactNativeMaps = fs.readFileSync(
+  path.join(
+    __dirname,
+    '..',
+    'node_modules/react-native-maps/lib/android/src/main/AndroidManifest.xml',
+  ),
+  'utf-8',
+);
+const newAndroidManifestReactNativeMaps = androidManifestReactNativeMaps.replace(
+  '<uses-sdk android:minSdkVersion="16" />',
```
RobinCsl: How can we be sure this will work for all Android versions? 
Can you tell me more about the issue itself?

ferrannp: It will. It works anyway (building), the problem is that if you open it with Android Studio it will not sync (making the work with Android Studio harder). This is solved in newer versions `react-native-map` but we had to downgrade: https://github.com/kiwicom/mobile/issues/1371.

Thing is that long ago Android removed those minSdkVersion from `AndroidManifest.xml` but that old `react-native-maps` version still has it.

RobinCsl: Thank you for this detailed answer!

PR #1383, opened 2018-12-07
Hotels: Show dropmarker image on single hotel map

### LINKED ISSUE #1381: Dropmarker is not visible on single-hotel-map

tbergq:
closes #1381 

I have resized the dropmarker.png image to be able to pass it directly to react-native-maps Marker since there is no way to style it through their api. 

I am not sure the size is 100% correct, but now it is visible on both platforms.


RobinCsl:
Just to make sure, is the positioning still the same as before? I imagine that resizing the image will have changed the relative position of the image compared to before?

tbergq:
We pass the coordinates and the source of the image file into react-native-maps, I expect them to take care of this, but yeah, it seems to be spot on

image

RobinCsl:
Marvellous!

PR #1385, opened 2018-12-14
Add universal-components and use icons in filter buttons

tbergq:


app/hotels/src/filter/FilterButton.js
```diff
@@ -2,26 +2,26 @@
...
-  +icon?: React.Element<typeof TextIcon>,
+  +icon?: React.Element<typeof Icon>,
 |};
 
 const FilterButton = ({ title, isActive, onPress, icon }: Props) => {
   const filterIcon =
     icon == null
       ? null
       : React.cloneElement(icon, {
-          style: [
```
RobinCsl: Is this why you wanted style prop to be added?
I am not sure how I feel about this anymore. It's kind of awkward, but it's also probably because trying to make it work on the web relies on my temporary hack. I think we might need to stabilize that component before using it extensively, and especially here.
What do you think?

tbergq: First I thought it would be nice that it works like `react-native-vector-icons` which takes name, color, size and style. 

The main reason I want to allow styles is for margin and padding. Now I have to wrap the icon in a view to allow margins

tbergq: > I think we might need to stabilize that component before using it extensively

What do we expect to change? I never asked for the size or color prop to be removed. Just to add a new style prop.

tbergq: Well, margins and padding was what came to mind. There could be other things also. 

If the thing is that you don't want to allow changing fontSize and color you could do something like

```
type IconProps = {|
//... some props
  +style?: $Rest<StylePropType, {| +color: string, +fontSize: number|}>
|}
```

which would disallow changing these style properties

RobinCsl: However, upon closer inspection, it seems like flow types are not being picked up (probably because of the `index.js` and `index.native.js` inside `@kiwicom/universal-components`). Writing
```
import { Icon } from '@kiwicom/universal-components;

...
<Icon name="john-snow" />
...
```
does not trigger any flow error.

tbergq:
Should I close this PR @RobinCsl ?

PR #1387, opened 2018-12-17
Fix: translation strings compatible with native code implementation

RobinCsl:


mrtnzlml:
The main reason why it's not very readable is the way how it's written. Look at this modified test:

```js
it('handles natively translated strings -- with several parameters', () => {
   const translationId = 'hotels.gallery.pagination';
   const translatedString = '%1$@ of %2$@';
   expect(getTranslation(translatedString, translationId)).toBe(
     '__1_photoNumber__ of __2_totalPhotos__',
   );
});
```

So the question must be asked: where is the `photoNumber` and `totalPhotos` comming from? As you can see it's pretty magic so you have to dig deep in order to uderstand it. Better way how to implement it would be to accept the vocabulary as an argument which means larger refactoring. The tests would be very obvious though after this change.

RobinCsl:
I tried to implement your suggestion and I believe it is clearer now. Could you review again?

PR #1389, opened 2019-01-02
Set code push install mode to immediate

tbergq:


tbergq:
Yes, no native changes here. Let's see how it works, ferran mentioned it could be a blink while it installs. So let's see. If it does not look good, we can always remove it again

PR #1391, opened 2019-01-02
Added missing nullability annotations 

pistelak:
This PR adds missing nullability annotations to RNLoggingModule. Please, note that nullability annotations has to be defined also as part of `RCT_EXPORT_METHOD`.


tbergq:
Let's wait to merge this until android is done with their release 🤞

PR #1398, opened 2019-01-03
Fixing Enum type from hotelsGalleryOpened

ferrannp:


android/react-native-native-modules/src/main/java/com/skypicker/reactnative/nativemodules/logging/RNLoggingModule.kt
```diff
@@ -163,9 +163,9 @@ class RNLoggingModule(reactContext: ReactApplicationContext, activeBooking: Bool
     }
 
     @ReactMethod
-    fun hotelsGalleryOpened(type: HotelsGalleryOpened.Type) {
+    fun hotelsGalleryOpened(type: String) {
         eventSenders.sendEvent(
-                HotelsGalleryOpened(type)
+            HotelsGalleryOpened(HotelsGalleryOpened.Type.valueOf(type))
```
RobinCsl: Ah, right. It would not compile without wrapping `type` like this.

ferrannp: Yeah because you can only pass certain types through the bridge (like `String`).

PR #1400, opened 2019-01-04
Show warning when selected rooms don't have room for everyone

tbergq:
We want to let the user know when he is searching for hotels for 5 persons,
but selects beds for less than 5 persons

Also, we wait for next android release before we merge this feature. It can be deployed rapidly with code-push anyways 😄

### DIFF
--- a/app/hotels/src/singleHotel/bookNow/BookNow.js
+++ b/app/hotels/src/singleHotel/bookNow/BookNow.js
@@ -7,7 +7,7 @@ import {
   withNavigation,
   type NavigationType,
 } from '@kiwicom/mobile-navigation';
-import { DeviceInfo, Translation } from '@kiwicom/mobile-localization';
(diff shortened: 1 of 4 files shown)


app/hotels/src/singleHotel/bookNow/BookNow.js
```diff
@@ -40,9 +42,37 @@ export class BookNow extends React.Component<Props> {
...
+          id: 'single_hotel.book_now.alert_title',
+        },
+        {
+          id: 'single_hotel.book_now.alert_body',
+          values: { numberOfGuests },
+        },
+        [
+          {
+            text: { id: 'shared.button.cancel' },
+          },
```
RobinCsl:
```suggestion
            style: 'destructive',
          },
```

RobinCsl: It would make UX better in my opinion.
simulator screen shot - iphone x - 2019-01-07 at 16 06 21

PR #1401, opened 2019-01-04
Fix tests failing in Lima timezone

tbergq:


app/MMB/src/scenes/tripServices/hotelMenuItem/HotelMenuItem.js
```diff
@@ -45,6 +45,11 @@ export type HotelData = {|
   +checkout: string | null,
 |};
 
+const convertInputStringToDate = (date: string) => {
```
RobinCsl: I thought it could live in DateUtils. What do you think?

RobinCsl: And with some test?

tbergq: Yeah, sure, I will move it later 👍

tbergq: We actually have this possibility through a private function in DateUtils, I will make it public.

tbergq:
@RobinCsl 🚀

PR #1405, opened 2019-01-07
Add @kiwicom/eslint-config

tbergq:


tbergq:
Tests and e2e tests pass. Should be good to go after we fix last of the failing lints due to .ios .android

mrtnzlml:
I definitely see an improvement here. My favorite parts are the fixed Relay queries (the unnecessary fields)... :)

tbergq:
> My favorite parts are the fixed Relay queries

Yeah, maybe some parts in MMB broke 🤷‍♂️ Somewhere i had to deactivate it, and hope to fix it later. Since we have some cases where we use the data from the query in other files (which we should not)

app/MMB/package.json
```diff
@@ -11,6 +11,8 @@
     "@kiwicom/mobile-pdf": "^0",
     "@kiwicom/mobile-relay": "^0",
     "@kiwicom/mobile-shared": "^0",
+    "@kiwicom/rnmodules": "^0",
+    "@kiwicom/react-native-app-hotels":"^0",
```
RobinCsl: Why was that dependency added to MMB?

mrtnzlml: I bet because you are using hotels somewhere inside MMB?

mrtnzlml: > Hmmm, not sure...

https://github.com/kiwicom/mobile/blob/34684baf3557f0fef451c0f555505bc948855351/app/MMB/src/navigation/NavigationStack.js#L11 :)

PR #1408, opened 2019-01-11
Update react-native-code-push to 5.5.1, react-native to 0.57.8 and react to 16.6.3

### LINKED ISSUE #1407: Update CodePush
We are interested in:

> Improvements:
Added the rollback retry mechanism that allows the application to attempt to reinstall an update that was previously rolled back.

ferrannp:
Fixes #1407.

Looks like they fixed the Snapshot issue they introduced in some version (as you can see in the updated ones).

Please test that every screen I might not be aware of works as expected.

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,6 +93,6 @@
     "test-runner": "jest"
   },
   "rnkiwimobile": {
-    "code-push-target-binary-version": "6.0.0"
+    "code-push-target-binary-version": "7.0.0"
   }
 }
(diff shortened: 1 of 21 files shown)


package.json
```diff
@@ -94,6 +93,6 @@
     "test-runner": "jest"
   },
   "rnkiwimobile": {
-    "code-push-target-binary-version": "6.0.0"
+    "code-push-target-binary-version": "7.0.0"
```
RobinCsl: We need to modify another file for this to work, right?

PR #1415, opened 2019-01-14
Hotels: Use rest of new log functions

### LINKED ISSUE #1390: Add rest of new logs
See comment in #1386

tbergq:
closes #1390

### DIFF
--- a/packages/shared/src/Logger.js
+++ b/packages/shared/src/Logger.js
@@ -2,6 +2,8 @@
 
 import { NativeModules } from 'react-native';
 
+import currencyFormatter from './CurrencyFormatter';
+
 const LoggingModule = NativeModules.RNLoggingModule;
 
 const {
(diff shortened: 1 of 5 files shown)


packages/shared/src/Logger.js
```diff
@@ -72,18 +74,24 @@ const HotelGalleryType = {
...
-  price: number,
-  formattedPrice: string,
-) {
+async function hotelsBookNowPressed({
+  hotelID,
+  rooms,
+  guests,
+  price,
+}: HotelsBookNowArgs) {
+  const formattedPrice = await currencyFormatter(price, 'EUR');
```
RobinCsl: Is it always going to be `EUR`?

tbergq: Yes, I guess that makes most sense, or we could change it to CZK later if they prefer that.

RobinCsl: But we are getting the currency there as well: https://github.com/kiwicom/mobile/pull/1415/files#diff-6b6ced58fb2208dc5f52e08d6530f058R38, so I am confused why this value needs to be hardcoded?

tbergq: we always get the price in EUR from booking.com. The currencyFormatter takes price in euro and formats it to the desired currency. 

```
const price = 150 // This is price in euro
const currency = 'EUR'
const formattedCurrency = currencyFormatter(price, currency);
console.log(formattedCurrency)
// 150€
```

In this case, this is just what we want to log. This formatting is for us. I don't see why we would log the price in the users currency.

tbergq: hmm, now I got confused. let me double check that

tbergq: Sorry, I was wrong. 

I will change this.

PR #1417, opened 2019-01-14
Hotels: Fix relay/unused-fields in HotelDetailScreen

tbergq:
Refactor countBookingPrice function into HotelDetailScreen since it is
not used elsewhere anymore


app/hotels/src/singleHotel/HotelDetailScreen.js
```diff
@@ -32,7 +30,11 @@ import {
   type HotelsContextState,
   type ApiProvider,
 } from '../HotelsContext';
-import countBookingPrice from './bookNow/countBookingPrice';
+
+type Money = {|
```
RobinCsl: Maybe `Price`, to be more consistent with the naming elsewhere?

app/hotels/src/singleHotel/HotelDetailScreenContext.js
```diff
@@ -32,8 +32,8 @@ type Props = {|
 |};
 
 type Money = {|
```
RobinCsl: Ah, but it's `Money` here. Well. It would be nice to make it more consistent 😄

PR #1419, opened 2019-01-14
Add polyfill for Object.setPrototypeOf

### LINKED ISSUE #1416: Object.setPrototypeOf is undefined on android
This causes app to crash if we get timeout error on android. 

mrtnzlml/quick-payments#44

tbergq:
closes #1416

### DIFF
--- a/app/polyfills/setPrototypeOf.js
+++ b/app/polyfills/setPrototypeOf.js
@@ -1,0 +1,10 @@
+// @flow
+
+// see https://github.com/kiwicom/mobile/issues/1416
+// $FlowExpectedError: We want to add this polyfill for android
+Object.setPrototypeOf =
+  Object.setPrototypeOf ||
+  function(obj, proto) {
+    obj.__proto__ = proto; // eslint-disable-line no-proto
+    return obj;
+  };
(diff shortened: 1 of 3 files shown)


app/polyfills/setPrototypeOf.js
```diff
@@ -0,0 +1,9 @@
+// @flow
+
+// $FlowExpectedError: We want to add this polyfill for android
```
RobinCsl: Could you add a link to the issues on react-native repo as well as the one you created here?

tbergq: Added a link to our issue. I don't think there is a specific issue for it in RN, it is a well known fact that the android JS environment is missing in some areas. 

I also expect this to be a temporary fix, as it will be fixed within the fetch package in a future release.

tbergq:
@RobinCsl 🚀

mrtnzlml:
@tbergq Does this actually work correctly?

tbergq:
Yes it does, it gives me error instance of TimeoutError === true

tbergq:
@mrtnzlml it will not work on [IE](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/setPrototypeOf#Polyfill), but that is ok for this project

PR #1435, opened 2019-01-16
Hotels: Rename hotelFacilities to hotelAmenities

tbergq:
The button is called Amenities, this is to make naming more consistent

### DIFF
--- a/app/hotels/src/allHotels/NewAllHotelsSearch.js
+++ b/app/hotels/src/allHotels/NewAllHotelsSearch.js
@@ -59,7 +59,7 @@ class NewAllHotelsSearch extends React.Component<Props> {
       roomsConfiguration,
       currency,
       orderBy,
-      filterParams,
+      filterParams: { hotelAmenities, ...filterParamsRest },
     } = this.props;
 
     return (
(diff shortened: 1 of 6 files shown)


app/hotels/src/allHotels/NewAllHotelsSearch.js
```diff
@@ -59,7 +59,7 @@ class NewAllHotelsSearch extends React.Component<Props> {
       roomsConfiguration,
       currency,
       orderBy,
-      filterParams,
+      filterParams: { hotelAmenities, ...filterParmsRest },
```
RobinCsl: `filterParmsRest`, it looks like there's a type in the variable name.

tbergq: If you mean typo, you are correct, I will fix it

tbergq:
@RobinCsl 🚀

app/hotels/src/allHotels/NewAllHotelsSearch.js
RobinCsl: Yes, I meant typo 😃

PR #1437, opened 2019-01-16
Hotels: Use icon component in HotelAmenities

tbergq:
Refactor component structure for hotel amenities


RobinCsl:
Maybe this was also the case in the previous PR's, but the icon size is smaller with your changes. Is it an issue?

tbergq:
> Is it an issue

I think not, we are now more `orbit`. I also checked the figma for this, and the icon height is 16, so I say it we had to big icons before 🙂

PR #1438, opened 2019-01-16
Replace TextIcon with Icon in hotels app

tbergq:


app/hotels/src/singleHotel/roomList/RoomBadges.js
```diff
@@ -67,8 +79,10 @@ const styles = StyleSheet.create({
     fontSize: 12,
   },
   iconStyle: {
-    color: defaultTokens.paletteBlueNormal,
     fontSize: 12,
+    width: 12,
```
RobinCsl: Should we expose the helper function from orbit components to create the styles automatically?

tbergq: No, I think not (not yet atleast). We have some custom sizes on stuff in mobile, and I think they are `design-bugs`

But I could maybe refactor this to the Icon component in mobile. 🤔 
But that is for other PR

PR #1441, opened 2019-01-17
Shared: Add fontSize prop for our icons with custom fontSize

tbergq:
Some of our icons have a custom fontSize (not orbit sizes).

We need to override width, height and lineHeight also, better to do this in the Icon component than adding this style to very single component that overrides this style

### DIFF
--- a/packages/shared/src/icons/Icon.js
+++ b/packages/shared/src/icons/Icon.js
@@ -12,6 +12,7 @@ type Props = {|
   +size?: 'small' | 'medium' | 'large',
   +color?: string,
   +style?: StylePropType,
+  +fontSize?: number,
 |};
 
 export default class Icon extends React.Component<Props> {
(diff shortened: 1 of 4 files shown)


packages/shared/src/icons/Icon.js
```diff
@@ -12,14 +12,35 @@ type Props = {|
...
+  getCustomFontSizeStyle = () => {
+    // Universal-components adds these values based on size. We have some custom sizes in mobile
+    // So we have to override these values
+    const { fontSize } = this.props;
+
+    if (fontSize == null) {
+      return null;
+    }
+
+    return {
```
RobinCsl: Should we return a `StyleSheet.create` of that object instead? It would be better for performance, right?

tbergq: If there were re-renders in the Universal-components triggered without re-render in this component, it would be yes. But I don't expect that the universal-component will re-render without this component re-rendering.

RobinCsl: But if we had many Icons on the screen all with the same imposed font size, that would harm performance, right? And it does not hurt to add this Stylsheet.create, or does it?

tbergq: But still the cached value would only live inside this particular render function, so

Icon-1 would have cache-1
Icon-2 would have cache-2

now Icon-1 re-renders, so it has cache-3.

RobinCsl: Oh really? I thought `StyleSheet.create` was a bit more clever and stored these numbers in some central style cache.

tbergq: No, it returns it to where you do StyleSheet.create(..)

So if you do in a file, as we normally do, it is persisted through the application. Then is you do it inside a render, it is persisted untill render is called again, then it creates a new cache.

RobinCsl: Makes sense. Thanks for the explanations 🙂

PR #1444, opened 2019-01-17
Fix: smaller DropMarker image

RobinCsl:
Summary: Since the map is just taking in the picture, a 150x150 image was displayed as marker. The base image is now 50x50 so that should fix MOBILE-2259.


tbergq:
Thanks, 

🤞 for that it works also in the app 😄

PR #1445, opened 2019-01-17
Add search form to homepage

tbergq:
It is not that pretty, but it is nice to be able to search in
different cities, different dates and room config

### DIFF
--- a/app/App.js
+++ b/app/App.js
@@ -4,6 +4,7 @@ import React from 'react';
 import { type DimensionType, Dimensions } from '@kiwicom/mobile-shared';
 
 import Navigation from './core/src/navigation';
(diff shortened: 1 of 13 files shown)


mrtnzlml:
Now imagine you are sharing the same thingy here with Margarita. What a dream... :)

RobinCsl:
I agree, not so pretty 😄 but it's a cool addition for our test environment.
And yeah, if we could steal the screen from Margarita, that would be even more awesome!

tbergq:
Yeah, this is what I meant with my comment on another PR. It is wrapped in a SafeAreaView. Sometimes it works, mostly it doesn't. I don't think this is important, since it is just in our test environment

PR #1450, opened 2019-01-18
Delete MMB app

tbergq:
There is no need to keep tons of dead code in the project.
It makes refactoring harder. If we ever go back to working on MMB
we can revert this commit and fix the issues.

And there is no need to update code-push-target-version, since no code in the bundle is touching the native modules that has been removed.


tbergq:
> bundled in the deployed app which are not used

Probably yes. Maybe they are not even bundled in 🤷‍♂️ 

If you are in doubt, just do a build which targets code-push staging, and test it 🙂

PR #1459, opened 2019-01-24
Add strings.xml to gitignore

tbergq:


README.md
```diff
@@ -67,6 +67,17 @@ and set it as your environment variable:
 export RNKIWIMOBILE_DEPLOYMENT_TOKEN="YOUR_TOKEN_HERE"
 ```
 
+
+also you should add two files `android/app/src/main/res/values/strings.xml` and `android/playground/src/main/res/values/strings.xml`
+```
+<resources>
+    <string name="app_name">reactNativeApp</string>
```
RobinCsl: The one for the Playground has a different name. Should you make that clearer here?

tbergq:
@RobinCsl 🚀

PR #1463, opened 2019-01-24
Remove unused packages assets and pdf

tbergq:


RobinCsl:
All of this was only ever used in MMB, right?

RobinCsl:
All of this was only ever used in MMB, right?

tbergq:
> All of this was only ever used in MMB, right?

Yes

PR #1467, opened 2019-02-01
Replace react-native-tooltips with tooltip from universal-components

tbergq:


app/hotels/src/navigation/HotelsNavigationOptions.js
```diff
@@ -84,7 +84,7 @@ export default ({
         <MapHeaderButton
           onPress={goToAllHotelsMap}
           icon={icon}
-          text={text}
+          text={text.id}
```
RobinCsl: Is that change a typo?

tbergq: No: 
```
type Props = {|
  +cityName?: string,
  +checkin?: string,
  +checkout?: string,
  +icon: React.Element<typeof Icon>,
  +text: {| +id: TranslationKeysType |},
  +goToAllHotelsMap: () => void,
|};
```

but it might be better to change this.

tbergq: @RobinCsl 🚀

app/hotels/src/navigation/allHotels/SearchResultsScreen.js
```diff
@@ -52,8 +52,8 @@ class SearchResultsScreen extends React.Component<Props> {
 
     const text =
```
RobinCsl: Maybe that should be `translationId` instead of `text`. WDYT?

tbergq: I think I should change it to just pass the `Translation` component, since that is what we normally do

RobinCsl: Yes, that makes sense.

tbergq:
@RobinCsl 🚀

tbergq:
I think text is sufficient since all our text is translated

PR #1471, opened 2019-02-08
Setup environment for new FastTrackBanner app

tbergq:


app/fastTrack/src/appRegistry/FastTrackBanner.js
```diff
@@ -0,0 +1,20 @@
...
+  render() {
+    return <Translation passThrough="Put code here" />;
+  }
+}
+
+// With navigation solves problem with swipe gestures, back button
+// And navigation between native app and RN app.
+// I am not 100% sure you need it if you will just run as a fragment.
+// You can experiment without to to verify
+export default WithNativeNavigation(FastTrackBanner, 'FastTrackBanner');
```
vblazenka: This should be exported without `WithNativeNavigation` HOC , as you said it will be used as a fragment.

PR #1472, opened 2019-02-08
Basic fast track banner mock

vblazenka:
This is just a simple view that will serve for banner mock with integrated modal inside just for testing.
I will use this a fragment inside native app.

When I succeed integrating this banner and when we get new designs, I will refactor this and to it how it should be done.

### DIFF
--- a/app/fastTrack/src/appRegistry/FastTrackBanner.js
+++ b/app/fastTrack/src/appRegistry/FastTrackBanner.js
@@ -1,20 +1,98 @@
-// @flow strict
+// @noflow
 
 import * as React from 'react';
-import { WithNativeNavigation } from '@kiwicom/mobile-shared';
-import { Translation } from '@kiwicom/mobile-localization';
+// eslint-disable-next-line
+import { View } from 'react-native';
+
+import {
+  SimpleCard,
+  Text,
+  CenteredView,
+  TextButton,
+  Modal,
+} from '@kiwicom/mobile-shared';
 
 // This will be prop types passed from native app
-type Props = {||};
+type Props = {|
+  +height: number,
+|};
 
-class FastTrackBanner extends React.Component<Props> {
+type State = {|
+  +isModalVisible: boolean,
+|};
+
+class FastTrackBanner extends React.Component<Props, State> {
+  constructor(props: Props) {
+    super(props);
+
+    this.state = {
+      isModalVisible: false,
+    };
+  }
+
+  onOpenModal = () => {
+    this.setState({
+      isModalVisible: true,
+    });
+  };
+
+  onCloseModal = () => {
+    this.setState({
+      isModalVisible: false,
+    });
+  };
+
+  renderModal = () => (
+    <Modal
+      isVisible={this.state.isModalVisible}
+      onRequestClose={this.onCloseModal}
+      testId="fastTrackBanner"
+    >
+      <View>
+        <Text
+          style={{
+            marginBottom: 16,
+            textAlign: 'center',
+            fontWeight: 'bold',
+            color: '#006dc7',
+          }}
+        >
+          This is modal content.
+        </Text>
+        <TextButton
+          style={{ flexBasis: 100 }}
+          type="secondary"
+          onPress={this.onCloseModal}
+          title="Close modal"
+        />
+      </View>
+    </Modal>
+  );
+
   render() {
-    return <Translation passThrough="Put code here" />;
+    return (
+      <SimpleCard style={{ height: this.props.height }}>
+        <React.Fragment>
+          <Text
+            style={{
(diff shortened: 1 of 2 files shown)


app/fastTrack/src/appRegistry/FastTrackBanner.js
```diff
@@ -1,20 +1,98 @@
-// @flow strict
+// @noflow
```
RobinCsl: Could you explain the reason behind this?

app/fastTrack/src/appRegistry/FastTrackBanner.js
```diff
@@ -1,20 +1,98 @@
...
 
 import * as React from 'react';
-import { WithNativeNavigation } from '@kiwicom/mobile-shared';
-import { Translation } from '@kiwicom/mobile-localization';
+// eslint-disable-next-line
```
RobinCsl: That should not be necessary?

app/fastTrack/src/appRegistry/FastTrackBanner.js
```diff
@@ -1,20 +1,98 @@
...
+  renderModal = () => (
+    <Modal
+      isVisible={this.state.isModalVisible}
+      onRequestClose={this.onCloseModal}
+      testId="fastTrackBanner"
+    >
+      <View>
+        <Text
+          style={{
+            marginBottom: 16,
```
RobinCsl: Could you extract these styles using `StyleSheet.create`, where `StyleSheet` is also in `@kiwicom/mobile-shared`?

app/fastTrack/src/appRegistry/FastTrackBanner.js
```diff
@@ -1,20 +1,98 @@
...
+    return (
+      <SimpleCard style={{ height: this.props.height }}>
+        <React.Fragment>
+          <Text
+            style={{
+              marginBottom: 16,
+              textAlign: 'center',
+              fontWeight: 'bold',
+              color: '#00a991',
+            }}
```
tbergq: I suggest you import `StyleSheet` from '@kiwicom/mobile-shared` and create styles, it is more performant than inline styles

tbergq: Also, you should `import { defaultTokens} from '@kiwicom/mobile-oribt` and use then for color and such

app/fastTrack/src/appRegistry/FastTrackBanner.js
```diff
@@ -1,20 +1,98 @@
...
+      <View>
+        <Text
+          style={{
+            marginBottom: 16,
+            textAlign: 'center',
+            fontWeight: 'bold',
+            color: '#006dc7',
+          }}
+        >
+          This is modal content.
```
tbergq: I suggest you use <Translation /> component so you don't forget to translate your strings.

app/fastTrack/src/appRegistry/FastTrackBanner.js
```diff
@@ -1,20 +1,98 @@
...
+            }}
+          >
+            Fast Track Banner test
+          </Text>
+          <TextButton
+            type="primary"
+            onPress={this.onOpenModal}
+            title="Open Modal"
+          />
+          {this.renderModal()}
```
tbergq: Please create a separate component instead

app/fastTrack/src/appRegistry/FastTrackBanner.js
```diff
@@ -1,20 +1,98 @@
...
 import * as React from 'react';
-import { WithNativeNavigation } from '@kiwicom/mobile-shared';
-import { Translation } from '@kiwicom/mobile-localization';
+// eslint-disable-next-line
+import { View } from 'react-native';
+
+import {
+  SimpleCard,
+  Text,
+  CenteredView,
```
tbergq: This component is not used

app/fastTrack/src/appRegistry/FastTrackBanner.js
```diff
@@ -1,20 +1,98 @@
...
+      </View>
+    </Modal>
+  );
 
-class FastTrackBanner extends React.Component<Props> {
   render() {
-    return <Translation passThrough="Put code here" />;
+    return (
+      <SimpleCard style={{ height: this.props.height }}>
+        <React.Fragment>
```
tbergq: I don't think this fragment is necessary

app/fastTrack/src/appRegistry/FastTrackBanner.js
vblazenka: I will fix everything. I didn't want to spend a lot of time on this component since I will delete everything anyways.

PR #1485, opened 2019-03-01
Hotels: Show datepicker when clicking on date in header

tbergq:
MOBILE-2972

### DIFF
--- a/app/hotels/src/HotelsContext.js
+++ b/app/hotels/src/HotelsContext.js
@@ -2,6 +2,7 @@
 
 import * as React from 'react';
 import { withContext } from '@kiwicom/mobile-shared';
+import { DateUtils } from '@kiwicom/mobile-localization';
 
 import type { CurrentSearchStats } from './filter/CurrentSearchStatsType';
 
@@ -24,6 +25,8 @@ const defaultState = {
   },
   actions: {
     setCurrentSearchStats: () => {},
+    setCheckinDate: () => {},
+    setCheckoutDate: () => {},
   },
   getGuestCount: () => 0,
   setHotelId: () => {},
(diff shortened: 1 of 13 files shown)


app/hotels/src/HotelsContext.js
```diff
@@ -143,6 +150,47 @@ class Provider extends React.Component<Props, State> {
 
   setPaymentLink = (paymentLink: ?string) => this.setState({ paymentLink });
 
+  setCheckinDate = (checkin: Date) => {
```
RobinCsl: All this logic is because Booking.com only allows bookings for up to one month, correct?

app/hotels/src/navigation/allHotels/SearchDatepickers.js
```diff
@@ -0,0 +1,99 @@
...
+          maxDate={DateUtils().addDays(364)}
+        />
+        <DatePicker
+          isVisible={this.props.showCheckoutDatepicker}
+          labels={labels}
+          onDismiss={this.props.toggleShowCheckinDatepicker}
+          onConfirm={this.setCheckoutDate}
+          date={new Date(this.props.checkout)}
+          minDate={DateUtils().addDays(1)}
+          maxDate={DateUtils().addDays(365)}
```
RobinCsl: I guess that since it's possible to book so far ahead, it would be nice to see the year in the badges if it's different than the current year. WDYT?

app/hotels/src/HotelsContext.js
tbergq: Correct

app/hotels/src/navigation/allHotels/SearchDatepickers.js
tbergq: Yeah, this is according to design though. I could check with designer if we need to display year here. But, if you don't mind, I would like to do a new PR if the designer decides we should add year.

PR #1487, opened 2019-03-01
Upgrade kiwicom/relay and kiwicom/eslint-config

tbergq:


app/hotels/src/allHotels/AllHotelsSearchList.js
```diff
@@ -22,10 +22,10 @@ type HotelType = {|
 
 export class AllHotelsSearchList extends React.Component<Props> {
   componentDidMount() {
-    const hotelId = this.props.data[0]?.hotelId;
```
RobinCsl: What's the reason behind this refactor?

app/hotels/src/allHotels/AllHotelsSearchList.js
```diff
@@ -7,11 +7,11 @@ import { createFragmentContainer, graphql } from '@kiwicom/mobile-relay';
 import { Translation } from '@kiwicom/mobile-localization';
 
 import AllHotelsSearchRow from './AllHotelsSearchRow';
-import type { AllHotelsSearchList as AllHotelsSearchListProps } from './__generated__/AllHotelsSearchList.graphql';
+import type { AllHotelsSearchList_data as AllHotelsSearchListProps } from './__generated__/AllHotelsSearchList_data.graphql';
 import { withHotelsContext, type HotelsContextState } from '../HotelsContext';
 
 type Props = {|
-  +data: AllHotelsSearchListProps | Array<?empty>,
```
RobinCsl: Maybe because that changed 😊

app/hotels/src/allHotels/AllHotelsSearchList.js
tbergq: Because flow type definition of `data` was wrong

PR #1502, opened 2019-03-08
Relay: upgrade kiwicom/relay and kiwicom/fetch

tbergq:


packages/relay/index.js
```diff
@@ -1,8 +1,10 @@
 // @flow
 
 import * as React from 'react';
-import Relay from '@kiwicom/relay';
-import { type GraphQLTaggedNode as _GraphQLTaggedNode } from 'react-relay'; // eslint-disable-line
+import {
+  commitMutation as _commitMutation,
```
RobinCsl: Is it just to put emphasis on its _internal API_ nature?

app/core/src/screens/hotelsStack/placepicker/HotelCityQuery.js
```diff
@@ -1,18 +1,14 @@
...
-} from '@kiwicom/mobile-relay';
+import { PublicApiRenderer, graphql } from '@kiwicom/mobile-relay';
 
 import type { HotelCityQueryResponse as HotelCityType } from './__generated__/HotelCityQuery.graphql';
 import HotelCityList from './HotelCityList';
 
 type Props = {|
   +onPress: () => void,
-  +query: GraphQLTaggedNode,
+  +query: string,
```
RobinCsl: Why is this change necessary?

packages/relay/index.js
tbergq: No, it is because I do export a const named commitMutation, so there would be a naming collision

app/core/src/screens/hotelsStack/placepicker/HotelCityQuery.js
tbergq: Because the type of query is string, and not GraphQLTaggedNode.

tbergq: It is not like the query from graphql tagget template. It is a prop called query

packages/relay/index.js
RobinCsl: Ah right. 😄 I should have expanded the file view!

PR #1504, opened 2019-03-08
Localization: Create script to get untranslated strings

### LINKED ISSUE #1496: Create script to get all untranslated keys in phraseApp
Getting keys translated takes too long. We have 7 months old keys that has not yet been translated. It seems we need to provide them with a list of untranslated keys to get them to translate. 

I would rather not to this job manually each time. 

Output of the script should be written to some json | txt file.

tbergq:
closes #1496


packages/localization/src/scripts/UntranslatedStrings.js
```diff
@@ -0,0 +1,50 @@
...
+import { getVocabularies } from '../DefaultVocabulary';
+import {
+  _fetch,
+  _paginate,
+  findKeys,
+  _log,
+  _removePrefix,
+} from './PhraseAppHelpers';
+
+const localTranslations = getVocabularies([
```
RobinCsl: Maybe add the same comment as here https://github.com/kiwicom/mobile/pull/1504/files#diff-74c59ca94d3d71b23941003d7840e88eL21

or extract that to a common file too?

tbergq: Sure 👍

tbergq: 🚀

tbergq:
> You probably want to add a untranslatedStrings.txt in .gitignore?

good idea. 
🚀

PR #1506, opened 2019-03-11
Relay: Refactor QueryRenderer to use props from kiwi package

tbergq:
Replace footer with errorFooter.
Reconnect close modal components where they belong in component tree.

### DIFF
--- a/app/hotels/src/map/singleHotel/SingleMap.js
+++ b/app/hotels/src/map/singleHotel/SingleMap.js
@@ -5,17 +5,26 @@ import { View } from 'react-native';
 import { graphql, createFragmentContainer } from '@kiwicom/mobile-relay';
 import { StyleSheet, StretchedImage } from '@kiwicom/mobile-shared';
 import { defaultTokens } from '@kiwicom/mobile-orbit';
+import {
+  withNavigation,
+  type NavigationType,
+} from '@kiwicom/mobile-navigation';
 
 import MapView from './MapView';
 import AdditionalInfo from './AdditionalInfo';
 import gradient from '../gradient.png';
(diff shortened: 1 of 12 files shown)


app/hotels/src/map/singleHotel/SingleMap.js
```diff
@@ -5,17 +5,26 @@ import { View } from 'react-native';
...
 import type { SingleMap_hotel as HotelType } from './__generated__/SingleMap_hotel.graphql';
+import CloseModal from '../../components/CloseModal';
 
 type Props = {|
   +hotel: ?HotelType,
+  +navigation: NavigationType,
 |};
 
 const SingleMap = (props: Props) => {
+  function goBack() {
```
RobinCsl: As far as I understand, this is similar as class construction. Will that be recreated on every render?

tbergq: It probably will. But the future is with react-hooks, and hooks will be available from rn-0.59. This is how we have to create functions with hooks.

packages/relay/src/__tests__/__snapshots__/QueryRenderer.test.js.snap
```diff
@@ -2,8 +2,13 @@
 
 exports[`QueryRenderer renders footerComponent if passed 1`] = `
```
RobinCsl: Message does not seem to agree with the output?

tbergq: Thanks, I will update the test 😄

tbergq: 🚀

PR #1507, opened 2019-03-12
Hotels: Use local schema to keep track of selected rooms

tbergq:
Delete HotelDetailScreenContext


app/hotels/src/singleHotel/roomList/RoomRow.js
```diff
@@ -29,36 +30,49 @@ type Props = {|
...
-  },
-  +numberOfRooms: number,
   +getGuestCount: () => number,
+  +numberOfRooms: number,
+  +relay: RelayProp,
 |};
 
 export class RoomRow extends React.Component<Props> {
-  isDisabled = () => this.props.getGuestCount() <= this.props.numberOfRooms;
+  isDisabled = () => this.props.numberOfRooms >= this.props.getGuestCount();
```
RobinCsl: This change does not seem necessary?

RobinCsl: (But I'm fine with it 😉 )

tbergq: It is more readable in my opinion, that's why I changed it 😊

app/hotels/src/singleHotel/summary/RoomSummary.js
```diff
@@ -1,44 +1,98 @@
...
+    const positiveSelections = availableRooms.filter(room => {
+      const selectedCount = room?.selectedCount ?? 0;
+      return selectedCount > 0;
+    });
+
+    if (positiveSelections.length === 0) {
+      return null;
+    }
+    const currency =
+      positiveSelections[0]?.incrementalPriceWithExtraCharges?.[0]?.price
```
RobinCsl: So this will change in the near future, since `Price` is deprecated?

app/hotels/src/singleHotel/summary/RoomSummary.js
```diff
@@ -1,44 +1,98 @@
...
+        ?.currency;
+    const amountAndPersons = positiveSelections.reduce(
+      (acc, room) => {
+        const selectedCount = room?.selectedCount ?? 0;
+        const incrementalPrice =
+          room?.incrementalPriceWithExtraCharges?.[selectedCount - 1];
+        const amount = incrementalPrice?.price?.amount ?? 0;
```
RobinCsl: We might need to use something more robust for price calculation as float operations are not exact in JavaScript. cf Slack conversion in #graphql channel about Dinero.js

app/hotels/src/singleHotel/summary/RoomSummary.js
tbergq: Yes, I should apply new type on gql first

app/hotels/src/singleHotel/summary/RoomSummary.js
tbergq: Yes, but I will leave that to future PR after adding new money type.

app/hotels/src/singleHotel/summary/__tests__/RoomSummary.test.js
```diff
@@ -0,0 +1,73 @@
...
+};
+
+describe('calculateSelectedData', () => {
+  it('calculates data correctly', () => {
+    // $FlowExpectedError: passing just props needed to run test
+    const Component = new RoomSummary({ room });
+
+    const test = Component.calculateSelectedData();
+    expect(test).toEqual({
+      amount: 25,
```
RobinCsl: Is this amount correct?
I would have thought that two rooms with (5+20) and another with (5), should be
2 * 25 + 1*5 = 55.

tbergq: No need to multiplicate, the prices under incrementalPrice[selectedCount - 1] is already calculated.

PR #1509, opened 2019-03-14
Relay: Add sentry logging for failed graphql-requests

### LINKED ISSUE #1479: Setup sentry logging for graphql errors

tbergq:
closes #1479


packages/relay/package.json
```diff
@@ -7,16 +7,18 @@
...
+    "@sentry/browser": "^4.6.4"
   },
   "devDependencies": {
     "graphql": "^14.0.2"
   },
   "peerDependencies": {
     "mockdate": "^2",
+    "react": "^",
     "react-native": "^0.57",
+    "react-native-dotenv": "",
```
RobinCsl: No version number here or ^, is it a typo?

tbergq: It doesn't really matter what is there 😊

tbergq:
@RobinCsl finally got the CI happy
🚀

packages/relay/package.json
```diff
@@ -7,16 +7,18 @@
     "@kiwicom/fetch": "^2.2.1",
     "@kiwicom/mobile-localization": "^0",
     "@kiwicom/mobile-shared": "^0",
-    "@kiwicom/relay": "^0.27.0"
+    "@kiwicom/relay": "^0.27.0",
+    "@sentry/browser": "^4.6.4"
```
RobinCsl: browser?

RobinCsl: Okay, related to this https://github.com/kiwicom/mobile/pull/1509/files#diff-e0465f4baf58d3c87df7993cb549f65fR12

scripts/configureApplication.js
```diff
@@ -4,56 +4,15 @@ const fs = require('fs');
...
+const { config } = require('dotenv');
+
+config();
+
+const { SENTRY_DSN } = process.env;
 
 const log = message => console.log(`➤ ${message}`); // eslint-disable-line no-console
 const error = message => console.log(chalk.red(`➤ ${message}`)); // eslint-disable-line no-console
 
-// TODO: fetch .env variables from Vault (or maybe just store the .env file itself in the Vault - this is painful)
```
RobinCsl: Removed because this was dead code, right?

packages/relay/package.json
tbergq: Yes, the browser package works. The react-native-package requires setup on the rootView (even outstide of our framework) and we don't want to catch native errors. Just grapqhl errors

scripts/configureApplication.js
tbergq: Yes, I think it is better to be able to actually use the `.env` than waiting around for something that we have not been able to achieve in 1.5 years 🙈

app/core/config/application.js
```diff
@@ -1,35 +1,11 @@
...
-  SENTRY_DSN,
-} from 'react-native-dotenv';
+import { SENTRY_DSN } from 'react-native-dotenv';
 
 const application = {
   animations: {
     duration: 100, // ms
   },
-  apiKey: {
-    googleMaps: sanitize(API_KEY_GOOGLE_MAPS),
```
RobinCsl: This used to be used in MMB, as far as I know.
We don't use it at all in hotels?

app/core/config/application.js
```diff
@@ -1,35 +1,11 @@
...
-        clientId: sanitize(AUTH_ANDROID_GOOGLE_CLIENTID),
-      },
-    },
-    ios: {
-      google: {
-        clientId: sanitize(AUTH_IOS_GOOGLE_CLIENTID),
-      },
-    },
-    kiwi: {
-      backend: sanitize(AUTH_KIWI_BACKEND),
```
RobinCsl: What were these for?

app/core/config/application.js
tbergq: This is in `secrets.xml`

app/core/config/application.js
tbergq: I don't think they ever were used

tbergq: At least for as long as I have been in the project

packages/relay/package.json
tbergq: Also not sure if it is better to use `@sentry/node` but it should be possible with `rn-nodeify`

app/core/config/application.js
tbergq: > We don't use it at all in hotels?

Yes, but we get it from native app. If we need to use it in development, we add it to `secrets.xml`

RobinCsl: Okay, but these only work for Android. It's fine with me since MMD required to make API calls to Google's API if I remember well.

PR #1519, opened 2019-03-18
Fix amount type sent to logger from BookNow

tbergq:


rn-cli.config.js
```diff
@@ -15,7 +15,7 @@ const sharedBlacklist = [
   /website\/node_modules\/.*/,
   /heapCapture\/bundle\.js/,
   /\.build[/\\].*/,
-  /ios[/\\]build[/\\]Build[/\\].*/,
+  /ios[/\\]build[/\\].*/,
```
tbergq: This change is related to me starting upgrade to RN 0.59. It seems it has changes the ios build folder.

PR #1520, opened 2019-03-26
Upgrade react-native to 0.59.0

tbergq:
Working on ios
Working on android

Upgrade react-native-maps to v0.24.0-rc1

### DIFF
--- a/app/hotels/src/map/allHotels/MapView.js
+++ b/app/hotels/src/map/allHotels/MapView.js
@@ -6,7 +6,10 @@ import { orderByDistance, getBounds } from 'geolib';
 import { createFragmentContainer, graphql } from '@kiwicom/mobile-relay';
 import { StyleSheet } from '@kiwicom/mobile-shared';
 
-import PriceMarker from './PriceMarker';
+import MapViewMarker, {
+  type MarkerPressEvent,
+  type LatLng,
+} from './MapViewMarker';
 import type { MapView_data as MapViewData } from './__generated__/MapView_data.graphql';
 
 type Props = {|
@@ -22,20 +25,6 @@ type State = {|
     latitudeDelta: number,
     longitudeDelta: number,
   |},
-|};
-
-type MarkerPressEvent = {|
-  nativeEvent: {
-    action: 'marker-press',
-    coordinate: LatLng,
-    id: string, // id of the marker
-    target: string,
-  },
-|};
-
-type LatLng = {|
-  longitude: number,
-  latitude: number,
 |};
 
 const NO_OF_MARKERS_IN_REGION = 15;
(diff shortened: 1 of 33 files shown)


packages/shared/package.json
```diff
@@ -5,26 +5,26 @@
...
     "@kiwicom/mobile-navigation": "^0",
     "@kiwicom/mobile-orbit": "^0",
-    "@kiwicom/mobile-config": "^0",
     "@kiwicom/rnmodules": "^0",
     "@ptomasroos/react-native-multi-slider": "^1.0.0",
     "prop-types": "^15.6.2",
     "querystring": "^0.2.0",
     "react-native-image-progress": "^1.1.1",
-    "react-native-maps": "0.21.0",
+    "react-native-maps": "github:react-native-community/react-native-maps#6c225af612d4681a00d4849b60774d7c6426835e",
```
tbergq: This was the commit has marking the `0.0.24-rc1`

tbergq: And also, 0.23.0 does not work with rn 0.59

tbergq:
also note, I am merging into a feature branch, I might want to do some more native upgrades before we merge this into master.

app/hotels/src/map/allHotels/MapView.js
```diff
@@ -175,7 +165,15 @@ export class Map extends React.Component<Props, State> {
       return;
     }
 
-    this.map.animateToCoordinate(coordinate, duration);
+    this.map.animateCamera(
```
RobinCsl: changes in API, I see.

tbergq: Yes, animateToCoordinate has been deprecated

packages/localization/src/Alert.js
```diff
@@ -37,8 +37,6 @@ type AlertOptions = {
 
 type AlertButtonStyle = 'default' | 'cancel' | 'destructive';
 
-type AlertType = 'default' | 'plain-text' | 'secure-text' | 'login-password';
```
RobinCsl: Why are we removing this?

packages/shared/package.json
```diff
@@ -5,26 +5,26 @@
...
     "react-native-modal": "^7.0.0",
     "react-native-read-more-text": "^1.1.0"
   },
   "peerDependencies": {
+    "@kiwicom/universal-components": "^0",
```
RobinCsl: If we're using TextInput from universal, there were some changes in prop names in v0.0.16.

RobinCsl: We definitely need to get better at maintaining some kind of changelog with breaking changes, etc. I will create an issue for it.

packages/localization/src/Alert.js
tbergq: Using alert type was deprecated, and has been removed in v 0.59

packages/shared/package.json
tbergq: Ok, thanks for the heads up, I am still on 0.0.14, and we don't have any textInput I think

RobinCsl: https://github.com/kiwicom/margarita/issues/595

packages/localization/src/Alert.js
tbergq: https://github.com/facebook/react-native/commit/e2bd7db

PR #1523, opened 2019-03-26
Upgrade react-native to 0.59.2

tbergq:
Upgrade react-navigation
Upgrade react-native-gesture-handler
Upgrade react-native-code-push

### DIFF
--- a/.build/package.json
+++ b/.build/package.json
@@ -4,9 +4,9 @@
   "private": false,
   "main": "index.js",
   "dependencies": {
-    "react-native": "0.59.0",
+    "react-native": "0.59.2",
     "react-native-maps": "0.24-rc1",
-    "react-native-code-push": "^5.5.2",
-    "react-native-gesture-handler": "^1.0.15"
+    "react-native-code-push": "5.6.0",
+    "react-native-gesture-handler": "1.1.0"
   }
 }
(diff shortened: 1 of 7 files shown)


RobinCsl:
Could not run it on Android because of 2.10 SNAPSHOT missing.

As for iOS, I got two warnings:
- one Relay-related 
Simulator Screen Shot - iPhone X - 2019-03-27 at 17 15 29

- one WebView related being extracted out of `react-native` core
Simulator Screen Shot - iPhone X - 2019-03-27 at 17 14 59

And running `yarn e2e-iphone`, I got the following error screen on all tests: 
Simulator Screen Shot - iPhone 6 - 2019-03-27 at 17 18 37

tbergq:
I am not 100% sure about the relay error, but it might come from cached result not being up to date with latest query 🤔 

There are also other deprecated features than webview, also asyncstorage. I plan to extract them after the upgrade is done, but before we merge into master, since they will be new native dependencies. 

detox is broken with react-natvigation. There are some issues about it, I expect them to be fixed in the near future. We can upgrade detox later without bumping target-version, since it is not production code. 

As for android, remember the part in the readme about doing upgrades 😊 

I will merge this one into feature branch for now.

PR #1524, opened 2019-03-27
Android: Upgrade logging module to 0.2.15

tbergq:


android/app/src/main/java/com/reactnativeapp/MainActivity.java
```diff
@@ -43,6 +43,7 @@ private void setupEventSender() {
           new ArrayList<>(),
           BuildConfig.DEBUG,
           BuildConfig.DEBUG,
+          BuildConfig.DEBUG,
```
RobinCsl: I could not find info about this change in signature. Do you have any resource?

tbergq: Me neither, I had to search for it in mobile-android repo.

PR #1532, opened 2019-03-28
Remove patching of RCTUtils from configureApplication

tbergq:
The fix has been merged into react-native itself

### DIFF
--- a/.build/package.json
+++ b/.build/package.json
@@ -4,9 +4,9 @@
   "private": false,
   "main": "index.js",
   "dependencies": {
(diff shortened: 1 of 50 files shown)


RobinCsl:
I guess for this one, I would need to test on a real iOS device, right? And the Playground version, not the `yarn ios` default. Correct?

tbergq:
That is one option. The other one is to go to the source code of the file and see that the line we try to replace no longer exists. So this part of the script no longer does anything.

tbergq:
replaced by #1534

PR #1533, opened 2019-03-29
Add webview from react-native-community

tbergq:


tbergq:
> no mock was added because we're not testing the code with a webview anywhere, right?

Right, it would have crashed any test were it was rendered. So no need to mock it (yet)

PR #1560, opened 2019-04-09
Localization: Use text from universal-components in translations

### LINKED ISSUE #1558: Move datepicker labels from hotels/components to shared
Now it makes tests fail because of require cycles. We will need to solve these require cycles first.

tbergq:
Move datepicker labels to datepicker component

closes #1558


tbergq:
I will try a different approach to solve this

PR #1562, opened 2019-04-09
Hotels: Refactor RenderSearchResults to use hooks

tbergq:


app/hotels/src/allHotels/RenderSearchResults.js
```diff
@@ -30,97 +30,97 @@ export const topValue = Dimensions.get('window').height;
 export const lowValue = -100;
 const transitionDuration = 250;
 
-export class RenderSearchResults extends React.Component<Props> {
-  constructor(props: Props) {
-    super(props);
+export const RenderSearchResults = (props: Props) => {
+  const [mapAnimation] = React.useState(
```
RobinCsl: Question: why use `useState` when we don't need the setter?
Ah, I guess it's for it not to be recreated on a re-render, right? Since it's a bit subtle, could you add some comment to explain?

tbergq: I think this comment would be abundant. This is how it will work now. Also, if you see all examples from react-native, they always put the animated value in state:

```
https://facebook.github.io/react-native/docs/animated
Animated.timing(
  // Animate value over time
  this.state.fadeAnim, // The value to drive
  {
    toValue: 1, // Animate to final value of 1
  },
).start(); // Start the animation
```

app/hotels/src/allHotels/__tests__/RenderSearchResults.test.js
```diff
@@ -18,16 +18,77 @@ const defaultProps = (stats: Object = { maxPrice: 100, minPrice: 50 }) => ({
   },
 });
 
+jest.mock('../../map/allHotels/MapScreen');
+jest.mock('../AllHotelsSearchList');
+
 it('it initiales correctly when it should show list', () => {
```
RobinCsl:
```suggestion
it('initialises correctly when it should show list', () => {
```

app/hotels/src/allHotels/__tests__/RenderSearchResults.test.js
```diff
@@ -18,16 +18,77 @@ const defaultProps = (stats: Object = { maxPrice: 100, minPrice: 50 }) => ({
...
+      "transform": Array [
+        Object {
+          "translateY": 1334,
+        },
+      ],
+    }
+  `);
 });
 
 it('it initiales correctly when it should show map', () => {
```
RobinCsl:
```suggestion
it('initialises correctly when it should show map', () => {
```

app/hotels/src/allHotels/RenderSearchResults.js
RobinCsl: Ok, I'm convinced!

PR #1563, opened 2019-04-10
Navigation: Add useSetNavigationParamsHook for setParams in useEffect hook

tbergq:


packages/navigation/src/useSetNavigationParams.js
```diff
@@ -0,0 +1,23 @@
...
+ * the component to go into endless re-render.
+ * This custom hook is a workaround for this, ensuring that you watch all values passed to set params, but not react-navigation itself.
+ */
+export default function useSetNavigationParams<T: {}>(
+  setParams: SetParams,
+  config: T,
+) {
+  React.useEffect(() => {
+    setParams(config);
+  }, Object.entries(config).map(([key, value]) => value)); // eslint-disable-line react-hooks/exhaustive-deps, no-unused-vars
```
RobinCsl:
```suggestion
  }, Object.values(config)); // eslint-disable-line react-hooks/exhaustive-deps
```

app/hotels/src/navigation/allHotels/SearchResultsScreen.js
```diff
@@ -44,20 +46,15 @@ type Props = {|
 const noop = () => {};
 
 const SearchResultsScreen = (props: Props) => {
-  React.useEffect(() => {
-    props.navigation.setParams({
-      toggleMap: props.setResultType,
-      setCheckinDate: props.setCheckinDate,
-      setCheckoutDate: props.setCheckoutDate,
-    });
-  }, []);
```
RobinCsl: So with the new implementation, all changes to the values are going to trigger a new `useEffect`, when the previous one set `toggleMap`, `setCheckinDate` and `setCheckoutDate` at the initial render only, and not in subsequent re-renders.
Is that a "regression" in some sense?

tbergq: After doing some research on the eslint-rule, and specifically this [comment](https://github.com/facebook/react/issues/14920#issuecomment-470863122) I think we should have all props in the array. 

None of this props will change, but in theory they could.

RobinCsl: Okay, makes sense! Thanks for sharing your research!

PR #1565, opened 2019-04-10
Hotels: Refactor SearchResultsContext to use hooks

tbergq:


app/hotels/src/navigation/allHotels/SearchResultsContext.js
```diff
@@ -10,48 +9,33 @@ const defaultValue = {
...
 };
 
-const { Consumer, Provider: ContextProvider } = React.createContext<State>({
+export const SearchResultsContext = React.createContext<State>({
   ...defaultValue,
 });
 
 type Props = {|
   +children: React.Node,
+  +show?: ResultType,
```
RobinCsl: Is that just for the test?

app/hotels/src/allHotels/__tests__/RenderSearchResults.test.js
```diff
@@ -22,44 +23,50 @@ jest.mock('../../map/allHotels/MapScreen');
...
+      `);
 });
 
 it('initialises correctly when it should show map', () => {
   const wrapper = renderer.create(
-    // $FlowExpectedError: Passing only props needed for this test
-    <RenderSearchResults {...defaultProps({}, 'map')} />,
+    <SearchResultsContext.Provider show="map">
+      {/* $FlowExpectedError: Passing only props needed for this test */}
+      <RenderSearchResults {...defaultProps({}, 'map')} />,
```
RobinCsl: Is passing `map` here necessary since you have the provider?

app/hotels/src/navigation/allHotels/SearchResultsContext.js
tbergq: Yes

app/hotels/src/allHotels/__tests__/RenderSearchResults.test.js
tbergq: True, I should remove that

tbergq: 🚀

PR #1569, opened 2019-04-11
Hotels: Add onLayout event to RenderSearchResults

### LINKED ISSUE #1568: Map is visible when you open hotels in landscape, then rotate device
ezgif com-video-to-gif (1)

tbergq:
If the iPad was opened in landscape mode, then flipped
the map would show underneath the list view.
This fixes this bug.

closes #1568


tbergq:
@RobinCsl 🚀

PR #1573, opened 2019-04-12
Use Hotels context as hook in

tbergq:
AllHotelsSearchList
AllHotelsSearchRow
HotelsPaginationContainer

### DIFF
--- a/app/hotels/src/HotelsContext.js
+++ b/app/hotels/src/HotelsContext.js
@@ -34,9 +34,11 @@ const defaultState = {
   closeHotels: () => {},
 };
 
-const { Consumer, Provider: ContextProvider } = React.createContext<State>({
+export const HotelsContext = React.createContext<State>({
   ...defaultState,
 });
+
+const { Consumer, Provider: ContextProvider } = HotelsContext;
 
 export type ApiProvider = 'booking' | 'stay22';
 
@@ -205,6 +207,4 @@ export const withHotelsContext = (select: (state: State) => Object) =>
 
 export type HotelsContextState = State;
 
-const HotelsContext = { Consumer, Provider };
-
-export default HotelsContext;
+export default { Consumer, Provider };
--- a/app/hotels/src/allHotels/AllHotelsSearchList.js
+++ b/app/hotels/src/allHotels/AllHotelsSearchList.js
@@ -7,11 +7,10 @@ import { createFragmentContainer, graphql } from '@kiwicom/mobile-relay';
 
 import AllHotelsSearchRow from './AllHotelsSearchRow';
 import type { AllHotelsSearchList_data as AllHotelsSearchListProps } from './__generated__/AllHotelsSearchList_data.graphql';
-import { withHotelsContext, type HotelsContextState } from '../HotelsContext';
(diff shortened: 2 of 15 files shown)


app/hotels/src/HotelsContext.js
```diff
@@ -205,6 +207,4 @@ export const withHotelsContext = (select: (state: State) => Object) =>
 
 export type HotelsContextState = State;
 
-const HotelsContext = { Consumer, Provider };
-
-export default HotelsContext;
+export default { Consumer, Provider };
```
RobinCsl: In the other PR, you only exported Provider I believe. Not sure if it's important!

tbergq: Yes, I hope I will get there, but I don't want to refactor all components in one PR. There is around 30 files connecting to this context.

app/hotels/src/allHotels/AllHotelsSearchList.js
```diff
@@ -7,73 +7,77 @@ import { createFragmentContainer, graphql } from '@kiwicom/mobile-relay';
...
+);
 
-  renderItem = ({ item, index }: {| +item: ?HotelType, +index: number |}) => (
-    <AllHotelsSearchRow
-      key={item?.id}
-      data={item}
-      testID={index === 0 ? 'firstHotelResult' : ''}
-    />
-  );
+function getFirst<T>(list: $ReadOnlyArray<T>): ?T {
```
RobinCsl: I have two questions about this file:
1. is there a reason to use `function` instead of an arrow function?
2. why did you extract this functions from the body of the component? Does that mean they will "never" be garbage collected since they're not part of the component?

tbergq: 1. Not really
2. 

> why did you extract this functions from the body of the component
They don't use any props, so they might as well be outside the component. 

> Does that mean they will "never" be garbage collected since they're not part of the component?

🤷‍♂️

app/hotels/src/allHotels/__tests__/HotelsPaginationContainer.test.js
```diff
@@ -1,63 +1,96 @@
...
+    </HotelsContext.Provider>,
+  );
+
 it('calls Logger when mounted', () => {
   const spy = jest.spyOn(Logger, 'ancillaryDisplayed');
-  // $FlowExpectedError: Passing only props needed for this test
-  const Component = new HotelsPaginationContainer(defaultProps());
-  Component.componentDidMount();
+  const wrapper = renderComponent(defaultProps());
+  wrapper.update(); // useEffect is not called on intial render, but when browser is idle, https://github.com/facebook/react/issues/14050#issuecomment-447888631
```
RobinCsl: It's a shame that hooks make the tests slightly less readable and understandable: why do you need to call `update` twice?

RobinCsl: Ah okay, to verify the logger was called only at the initial render.

tbergq: I don't need to call update twice. I call update twice, in order to see that the Logger is called only once.

PR #1574, opened 2019-04-15
mobile-relay: Fixed props for PrivateApiRenderer

vblazenka:
**PrivateApiRendererWithContext** as default export should take **authHeaderKey** as a prop and forward it to **PrivateApiRenderer** and _accessToken_ is optional since it is taken from context and not provided via props.


vblazenka:
@tbergq Done! :)

PR #1576, opened 2019-04-16
Hotels: Refactor hotels context components to use hooks

tbergq:


app/hotels/src/allHotels/NewAllHotels.js
```diff
@@ -5,28 +5,25 @@ import { View } from 'react-native';
 import { StyleSheet } from '@kiwicom/mobile-shared';
```
RobinCsl: Really nice how the code is simplified here. It's so much easier to read.

app/hotels/src/filter/price/PriceFilter.js
```diff
@@ -1,169 +1,105 @@
...
-          min={priceMin}
-          max={priceMax}
-          start={start}
-          end={end}
-          currency={currency}
-          daysOfStay={daysOfStay}
-        />
-      </React.Fragment>
-    );
+    props.onChange(filter);
```
RobinCsl: If I read well, there is no `closePopup` anymore in the `handleSave` function. Is that intentional?

tbergq: Yes, we always got a warning saying you can not setState after component is unmounted for this filter. I finally figured out why. When we set the new value to context, query component re-renders and searches with new input. Causing this component to unmount. So, there is no need to call closePopup, the component unmounts.

PR #1577, opened 2019-04-16
ancillaries fast track 

vblazenka:
Integrated fast track ancillary using relay.


tbergq:
There seems to be some error on gitlab not reporting CI result back to github, but your CI is failing: 

https://gitlab.com/kiwicom/mobile/-/jobs/197610258

app/ancillaryFactory/src/appRegistry/AncillaryFactory.js
```diff
@@ -0,0 +1,50 @@
...
+  +service: string,
+  +bookingId: number,
+  +kwAuthToken: string,
+|};
+
+class AncillaryFactory extends React.Component<Props> {
+  getRequestedAncillary() {
+    const { service } = this.props;
+
+    const ancillary: AncillaryDefinition =
```
tbergq: This type is wrong, it should be `?AncillaryDefinition`,

app/ancillaryFactory/src/appRegistry/AncillaryFactory.js
```diff
@@ -0,0 +1,50 @@
...
+      ancillariesDefinitions[service.toUpperCase()];
+
+    if (ancillary && typeof ancillary.renderComponent === 'function') {
+      return ancillary.renderComponent;
+    }
```
tbergq: you don't need this if else `return ancillary?.renderComponent;` is doing the exact same thing

app/ancillaryFactory/src/appRegistry/AncillaryFactory.js
```diff
@@ -0,0 +1,50 @@
...
+
+  render() {
+    const { bookingId, kwAuthToken } = this.props;
+
+    const ancillaryProps = {
+      bookingId,
+      kwAuthToken,
+    };
+
+    const Component = this.getRequestedAncillary() || NoAncillaryMessage;
```
tbergq: you should use `??` instead of `||`.

app/ancillaryFactory/src/components/NoAncillaryMessage.js
```diff
@@ -0,0 +1,15 @@
+// @flow strict
+
+import * as React from 'react';
+import { View } from 'react-native';
+import { Text, Translation } from '@kiwicom/mobile-shared';
+
+const NoAncillary = () => (
+  <View>
+    <Text>
```
tbergq: You don't need to wrap `<Translation />` in text unless you want to add some styles.

app/ancillaryFactory/src/components/NoAncillaryMessage.js
```diff
@@ -0,0 +1,15 @@
...
+      <Translation passThrough="No ancillary view found for the provided parameters." />
```
tbergq: Shouldn't this also be translated?

app/fastTrack/src/appRegistry/__tests__/__snapshots__/FastTrackBanner.test.js.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`FastTrackBanner Should render correctly 1`] = `null`;
```
tbergq: Is this the expected output of this component?

app/fastTrack/src/components/FastTrackModal.js
```diff
@@ -1,51 +1,79 @@
 // @flow strict
 
 import * as React from 'react';
-import { View } from 'react-native';
+import { View, WebView } from 'react-native';
```
tbergq: Please use `webview` from '@kiwicom/mobile-shared` instead. WebView from `react-native` is deprecated.

app/fastTrack/src/components/FastTrackModal.js
```diff
@@ -1,51 +1,79 @@
...
+    return (
+      <Modal
+        isVisible={props.isVisible}
+        onRequestClose={props.onCloseModal}
+        style={styles.modal}
+      >
+        <View style={styles.container}>
+          <WebView
+            source={{
+              uri: this.props.data?.url || '',
```
tbergq: replace `||` with `??`

app/ancillaryFactory/src/components/NoAncillaryMessage.js
vblazenka: I think no, I added this component as a helper for developers if they mistype ancillary service name.

app/fastTrack/src/appRegistry/__tests__/__snapshots__/FastTrackBanner.test.js.snap
vblazenka: Oh, definitely not. That component renders `PrivateApiRenderer` which uses `QueryRenderer`. I can't find any test that uses `QueryRenderer` or `PublicApiRenderer`. Can you give me some pointers, maybe I don't need to test this?

tbergq: So, actually, the test is not wrong, since 

```
// PrivateApiRenderer.js
 if (props.accessToken === null) {
    // TODO: Return login
    return null;
  }
```

You could do this: 

```
const component = renderer.create(
      <AuthContext.Provider accessToken="token">
        <FastTrackBanner bookingId={123} kwAuthToken="x.y.z" />,
      </AuthContext.Provider>,
    );
```

But please mock the `QueryRenderer` component, unless you will actually call the server.

Not sure what good that test would do though. I would rather consider just shallow rendering.

PR #1578, opened 2019-04-16
Hotels: Use hotels context as hook

tbergq:
add react-native-testing-library

### DIFF
--- a/app/hotels/src/map/allHotels/MapView.js
+++ b/app/hotels/src/map/allHotels/MapView.js
@@ -46,18 +46,9 @@ export class Map extends React.Component<Props, State> {
     };
   }
 
-  componentDidMount() {
-    const { selectedIndex } = this.props;
-    // it's necessary to call showCallout to bring selected marker to foreground
-    if (this.markers[selectedIndex]) {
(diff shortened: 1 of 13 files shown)


app/hotels/src/map/allHotels/MapView.js
```diff
@@ -46,18 +46,9 @@ export class Map extends React.Component<Props, State> {
...
-    }
-  }
-
-  componentDidUpdate() {
-    this.animateToCoordinate(this.props.selectedIndex);
-    if (this.markers[this.props.selectedIndex]) {
-      this.markers[this.props.selectedIndex].showCallout();
+  componentDidUpdate(prevProps: Props) {
+    if (prevProps.selectedIndex !== this.props.selectedIndex) {
+      this.animateToCoordinate(this.props.selectedIndex);
```
RobinCsl: Are you sure this is not necessary anymore?

tbergq: I cant see any difference with or without it. Marker is not brought forward anyway.

RobinCsl: Ah okay :D Then I'm fine with it!

PR #1583, opened 2019-04-23
Ancillaries: Added AncillaryViewController for integrating ancillaries

vblazenka:
This is remaining code for fast track ancillary.
iOS team wanted to have a clear view controller for ancillaries integration, that's why it doesn't have `RNKiwi` prefix.


vblazenka:
@tbergq I think it should work now. I had to add `AncillaryViewController` to target Headers in Build Phases.

PR #1591, opened 2019-04-26
Shared: Refactor sliderlabels

tbergq:
It was currently running in an endless loop
instead of recalculation on some prop change.
This took up too many resources. I also suspect it to
have caused MOBILE-3932 in JIRA


packages/shared/src/forms/__tests__/SliderLabels.test.js
```diff
@@ -0,0 +1,179 @@
...
+import * as React from 'react';
+import renderer from 'react-test-renderer';
+import ShallowRenderer from 'react-test-renderer/shallow';
+
+import Translation from '../../Translation';
+import SliderLabels from '../SliderLabels';
+
+const shallowRenderer = new ShallowRenderer();
+
+it('one label', () => {
```
RobinCsl:
```suggestion
test('one label', () => {
```

tbergq: I like `it`, but I will change it to `it('renders with one label)`

packages/shared/src/forms/useCalculateSliderPositions.js
```diff
@@ -0,0 +1,140 @@
...
+      let val;
+
+      if (input > max) {
+        val = max;
+      } else if (input < min) {
+        val = min;
+      } else {
+        val = input;
+      }
+      return Math.round((val / (max - min)) * width);
```
RobinCsl: I believe this could be larger than the width:
max=10
min=1
input = 10 => val = 10

10/9*width > 1*width

Is that the intended behaviour?

RobinCsl: Maybe it should be `(val - min) / (max - min) * width`, in which case:
* val = min => 0 (no offset)
* val = max => `width` (full offset)
* min < val < max => % of width

tbergq: This gives me a headache 🙈  I mostly copy pasted this from what it was, and it works as intended

Issue #1598, opened 2019-05-02
Update hotel result list design


tbergq:
[figma](https://www.figma.com/file/ayF92epBKcFcwdfE8IWIGMNa/Hotels-Results-%26-Detail?node-id=815%3A11052)

- [x] Header date button
- [x] Filter buttons
- [x] Price range popup
- [ ] Rating popup

PR #1599, opened 2019-05-02
Hotels: Apply new styles to date header button

tbergq:
relates to #1598


tbergq:
> It just seems in Figma that the text is slightly bolder than the result in the simulator

Maybe, but they didn't add any fontWeight to figma, so my best guess is to keep it to normal

PR #1601, opened 2019-05-02
Hotels: Update PricePopup with new design

tbergq:
Shared: Add type info and circled prop to AdaptableBadge
Shared: Update slider to match with new design

### DIFF
--- a/app/hotels/src/filter/price/PricePopup.js
+++ b/app/hotels/src/filter/price/PricePopup.js
@@ -7,9 +7,10 @@ import {
   ButtonPopup,
   Slider,
   StyleSheet,
-  SliderLabels,
   Price,
   Translation,
+  AdaptableBadge,
+  TranslationFragment,
 } from '@kiwicom/mobile-shared';
 import { defaultTokens } from '@kiwicom/mobile-orbit';
 import { SafeAreaView } from 'react-navigation';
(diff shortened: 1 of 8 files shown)


app/hotels/src/filter/price/PricePopup.js
```diff
@@ -77,30 +78,35 @@ export default class PricePopup extends React.Component<Props, State> {
...
-              <Price amount={start * daysOfStay} currency={currency} />
-            }
-            startValue={start}
-            endLabel={<Price amount={end * daysOfStay} currency={currency} />}
-            endValue={end}
-          />
+            <AdaptableBadge
+              translation={
+                <TranslationFragment>
+                  <Price amount={start * daysOfStay} currency={currency} />
```
RobinCsl: Should we use the number library to compute the price?

packages/shared/src/AdaptableBadge.js
```diff
@@ -1,4 +1,4 @@
-// @flow strict
+// @flow
```
RobinCsl: Why this change?

packages/shared/src/AdaptableBadge.js
```diff
@@ -51,4 +60,10 @@ const style = StyleSheet.create({
   text: {
     color: defaultTokens.paletteWhite,
   },
+  info: {
+    backgroundColor: defaultTokens.paletteBlueNormal,
+  },
+  circled: {
+    borderRadius: 11,
```
RobinCsl: The designer is not using Orbit values?

packages/shared/src/forms/SliderMarker.js
```diff
@@ -0,0 +1,45 @@
+// @flow
+
+import * as React from 'react';
+import { View } from 'react-native';
+import { defaultTokens } from '@kiwicom/mobile-orbit';
+
+import StyleSheet from '../PlatformStyleSheet';
+
+export default function SliderMarker() {
```
RobinCsl: There is no "pressed" state?

app/hotels/src/filter/price/PricePopup.js
tbergq: You mean decimal? No, since this particular value still comes as Float from the server.

packages/shared/src/AdaptableBadge.js
tbergq: It had it's reason, then I improved it, so yes, I will add it back tomorrow. Thanks 😊

packages/shared/src/AdaptableBadge.js
tbergq: I think you already know the answer to that 😞

packages/shared/src/forms/SliderMarker.js
tbergq: Should it be?

tbergq:
@RobinCsl 🚀

packages/shared/src/forms/SliderMarker.js
RobinCsl: As a user, I would like some visual feedback that I pressed the handle. But I don't know if the designer has this in mind? I could not find anything in Figma so 🤷‍♂

tbergq: When you have your finger over the marker, you don't see it

RobinCsl: My finger is not so big :D I guess that's because I'm using the mouse with the emulator.

tbergq: > I guess that's because I'm using the mouse with the emulator

🙈 

I can't see the marker when I put my finger on top of it, but when dragging, it becomes partially visible, but by then, I already know that I have activated it

PR #1602, opened 2019-05-03
Hotels: Update ScorePopup with new design

### LINKED ISSUE #1598: Update hotel result list design

tbergq:
Shared: Delete SliderLabels
closes #1598


app/hotels/src/filter/score/ScorePopup.js
```diff
@@ -29,6 +29,14 @@ const convertScoreToSliderValue = (minScore: number | null) => {
   return minScore ? minScore - SLIDER_SHIFT : 0;
 };
 
+const labels = {
```
RobinCsl: Why not use an array here? Is it to put emphasis on the importance of the order?

tbergq: Because we will only use one at a time. Putting it in an array, flow will ask for a `key` prop, which we really don't need in this case.

PR #1603, opened 2019-05-03
Hotels: Add input validation to search parameters

tbergq:
We have tons of errors in sentry, coming from checkin = checkout, checkin before today
checkout > checkin + 30, checkout more than one year in the future
cityId = -1. I registered a bug with the native team in November
and they still have not solved it. Let's do some validation on our end


packages/localization/src/vocabularies/Hotels.js
```diff
@@ -66,6 +66,16 @@ const HotelsVocabulary = {
     '__distance__km from center',
   'hotels_search.all_hotels_search_list.show_map': 'Show map',
   'hotels_search.all_hotels_search_list.show_list': 'Show list',
+  'hotels_search.new_all_hotels.interval_error':
+    'Time of stay must be between 1 and 30 days. Got __interval__',
```
RobinCsl:
```suggestion
    'Time of stay must be between 1 and 30 days. You chose __interval__ days.',
```

Issue #1607, opened 2019-05-06
RoomSummary is not given the right width prop in Tablet mode


RobinCsl:
The RoomSummary card computes the allowed space given for the price tag for SummaryRow component based on the width of the device, and it should be half the width in Tablet view.

PR #1608, opened 2019-05-06
Hotels: RoomSummary and HotelCard on SingleHotelMap are also swipeabl…

RobinCsl:
…e on Android

This fixes MOBILE-4113.

It seems that VerticalSwipeResponder does not like too much to be nested in `View` components for the swiping to work on Android. :man-shrugging:

These changes can be seen in Room Summary after selecting rooms and on the map view for a given hotel; it's now possible to swipe the card up on Android (was already working on iOS).

Another issue was spotted during this fix: the RoomSummary card computes the allowed space given for the price tag for SummaryRow component based on the width of the device, and it should be half the width in Tablet view. Issue created: https://github.com/kiwicom/mobile/issues/1607

### DIFF
--- a/app/hotels/src/map/singleHotel/AdditionalInfo.js
+++ b/app/hotels/src/map/singleHotel/AdditionalInfo.js
@@ -9,6 +9,7 @@ import {
   Device,
   BottomSheetHandle,
 } from '@kiwicom/mobile-shared';
+import { defaultTokens } from '@kiwicom/mobile-orbit';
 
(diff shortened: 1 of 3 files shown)


RobinCsl:
> (I am not trying to run these PRs since I believe you tried it and everything is OK.)

Sure, thanks for letting me know. I did try on Android tablet and phone, and iOS iPhone and iPad simulators. I believe QA will pay extra attention anyway.

PR #1609, opened 2019-05-06
Hotels: SummaryRow should use dynamic width to truncate text

RobinCsl:
This fixes https://github.com/kiwicom/mobile/issues/1607

The issue was that using the `withDimension` HOC meant using the dimensions of the device. On Tablet view, the RoomSummary component is displayed only on half the screen width while SummaryRow assumed there is the full width of the device to render the text. This caused the row to go out of the container if the name of the hotel room was too long on Tablet.

The SummaryRow component was refactored to use two onLayout event listeners so that both the width of the container of the row and the width of the price tag are computed, and the max width for text can be deduced from that.

### DIFF
--- a/app/hotels/src/singleHotel/summary/SummaryRow.js
+++ b/app/hotels/src/singleHotel/summary/SummaryRow.js
@@ -6,7 +6,6 @@ import {
   Price,
   Text,
   type OnLayout,
-  withDimensions,
   Translation,
 } from '@kiwicom/mobile-shared';
 import { View } from 'react-native';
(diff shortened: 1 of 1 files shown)


app/hotels/src/singleHotel/summary/SummaryRow.js
```diff
@@ -15,49 +14,38 @@ import { defaultTokens } from '@kiwicom/mobile-orbit';
...
-    );
+  function onPriceLayout(event: OnLayout) {
+    setPriceWidth(event.nativeEvent.layout.width);
   }
-}
+
+  const maxWidth =
+    containerWidth != null && priceWidth != null
+      ? containerWidth - priceWidth - SPACING
+      : 0;
```
mrtnzlml: Does it behave correctly when the max width is zero?

RobinCsl: Setting the max width to zero just displays the price on the right hand-side, and the text is obviously not showing. That's an acceptable behaviour for me (especially when it should not happen after the `View`s rendered and executed their `onLayout`).

Screenshot 2019-05-07 at 09 59 24
Screenshot 2019-05-07 at 10 01 00

PR #1612, opened 2019-05-24
Hotels: Set price stats to HotelsContext for stay22

tbergq:
Add complete flow types for decimal-light.

After this PR ->  #1611


tbergq:
> (though it's not currently displayed on Stay22 since there's no filters there, right?)

Yes, that is true, I forgot to mention that in the description. This is the first step to enable filtering for stay22. I also have a pending PR for price-filtering on the graphql-server.

Next step will be to implement this when it is merged on the server.

PR #1614, opened 2019-05-24
Hotels: Use decimal for priceStats

### LINKED ISSUE #1611: Refactor HotelsContext priceStats to use Decimal

tbergq:
closes #1611

### DIFF
--- a/app/hotels/src/filter/price/PricePopup.js
+++ b/app/hotels/src/filter/price/PricePopup.js
@@ -14,105 +14,107 @@ import {
 } from '@kiwicom/mobile-shared';
 import { defaultTokens } from '@kiwicom/mobile-orbit';
 import { SafeAreaView } from 'react-navigation';
+import { Decimal } from 'decimal.js-light';
 
 type Props = {|
   onClose: () => void,
   onSave: ({ minPrice: number, maxPrice: number }) => void,
   isVisible: boolean,
-  min: number,
-  max: number,
-  start: number,
-  end: number,
+  min: Decimal,
+  max: Decimal,
+  start: Decimal,
+  end: Decimal,
   currency: string,
   daysOfStay: number,
 |};
 
-type State = {|
-  price: {
-    start: number,
-    end: number,
-  },
-|};
+export default function PricePopup(props: Props) {
+  const { onSave: propsOnSave } = props;
+  const [price, setPrice] = React.useState({
+    start: new Decimal(0),
+    end: new Decimal(0),
+  });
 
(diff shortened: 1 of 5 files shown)


app/hotels/src/filter/price/PricePopup.js
```diff
@@ -14,105 +14,109 @@ import {
...
-            start,
-            end,
-          },
-        };
-  };
-
-  handlePriceChanged = ([start, end]: number[]) =>
-    this.setState({
-      price: { start, end },
+  React.useEffect(() => {
```
RobinCsl: So every time `end` and `start` change, this is fired. Isn't it too much with the `handleChange` `useCallback` below?

app/hotels/src/filter/price/PricePopup.js
```diff
@@ -14,105 +14,109 @@ import {
...
+        start: new Decimal(start),
+        end: new Decimal(end),
+      }),
+    [],
+  );
 
-  render() {
-    const { start, end } = this.state.price;
-    const { min, max, currency, daysOfStay } = this.props;
+  const savePrice = React.useCallback(() => {
```
RobinCsl: Can you explain why `useCallback` is used here? And why not use `useMemo` or nothing?

app/hotels/src/filter/price/PricePopup.js
tbergq: Yeah, this is the problem with destructuring. It is harder to understand what is actually happening. 

the `start` and `end` in the effect is from props. while the `handlePriceChange` below sets the values to state. 

I should refactor this to be more readable.

app/hotels/src/filter/price/PricePopup.js
tbergq: > The useCallback Hook lets you keep the same callback reference between re-renders so that shouldComponentUpdate continues to work

> The useMemo Hook makes it easier to control when individual children update, reducing the need for pure components.

>  useCallback(fn, deps) is equivalent to useMemo(() => fn, deps)

I guess both useCallback and useMemo could be used here.

app/hotels/src/filter/price/PricePopup.js
RobinCsl: > I should refactor this to be more readable.

So you mean in a future PR or in this PR?

tbergq: I did the refactor already

RobinCsl: Ah right, forgot to refresh 🙈

PR #1617, opened 2019-05-27
Hotels: Add filter stripe for stay22 hotels

tbergq:


app/hotels/src/allHotels/Stay22HotelsSearch.js
```diff
@@ -34,6 +36,7 @@ export function Stay22HotelsSearch() {
     longitude,
     closeHotels: onClose,
   }: HotelsContextState = React.useContext(HotelsContext);
+  const hotelsFilter = useHotelsFilter();
```
RobinCsl: As the app grows, it's going to be difficult to track how things relate to each other if we encapsulate logic so much. Like it's not clear that this is connected to the context, so it should be updated when the context changes, and then this should be re-rendered too.

I guess this is more a comment/question than a real review to modify it. 

At the same time, it does make code-sharing easier and makes the code of this file more readable. I'm just afraid if everything is hidden away that it will become difficult to debug at some point.

tbergq: I guess in pre-hook days, if you wanted to do something similar, you created a HOC, the HOC could connect to context and inject the hotelsFilter as a prop. The logic would still be hidden away, but re-usable. 

The `use...` prefix here should let you know it is a hook, and give you a hint to look inside, if you want to debug 🙂.

Hooks are still new, and we might do mistakes and learn from them. But this is one of the reasons hooks were created, to make sharing of logic between components easier.

RobinCsl: 💯% agree to all you said.

PR #1619, opened 2019-05-27
Upgrade dependencies

tbergq:


package.json
```diff
@@ -51,23 +51,23 @@
     "@babel/plugin-transform-runtime": "^7.4.4",
     "@babel/runtime": "^7.4.4",
     "@kiwicom/eslint-config": "^4.0.1",
-    "@kiwicom/monorepo": "^0.19.0",
+    "@kiwicom/monorepo": "^0.21.0",
```
mrtnzlml:
```diff
-    "@kiwicom/monorepo": "^0.21.0",
+    "@kiwicom/monorepo-utils": "^0.21.0",
```

tbergq: done in #1623 

Same package with new name, or are the also any differences @mrtnzlml

mrtnzlml: There shouldn't be any significant changes, only new name.

tbergq: Ok, thanks, New package has already been added 😊

PR #1624, opened 2019-06-04
Hotels: Add max and min date to datepicker

tbergq:
Adjust for timezone offset in hotels context. When max or min date limit is hit
datepicker will return date which will be actual day +,- one day
depending on the timezone.


app/hotels/src/navigation/HotelsNavigationOptions.js
```diff
@@ -68,6 +68,7 @@ const hotelsNavigationOptions = ({
             }
             date={checkin}
             onDateChange={setCheckinDate}
+            minDate={DateUtils.getToday()}
```
RobinCsl: Don't know if that makes a huge difference, but here you're creating 3 instances of `Date` by calling `DateUtils.getToday()`. Since `addDays` should be immutable, you could probably get away with creating just one Date instance and then call addDays to it below? WDYT?

tbergq: > Don't know if that makes a huge difference

It should not

RobinCsl: What I mean is: not in this one instance, but if we did this everywhere, then we would be consuming more memory than necessary.

tbergq: 🚀 

I think we saved memory consumption of all users of kiwi.com app 🎉

PR #1630, opened 2019-06-11
Shared: Add useApi hook

tbergq:
Hotels: Refactor HotelsContext to be functional component.
Use new useApi with dateFactory to separate checkin, checkout
logic from the render of the provider


app/hotels/src/HotelsContext.js
```diff
@@ -155,170 +156,57 @@ const validateInput = ({ checkin, checkout, cityId }: SearchInput) => {
...
+function Provider(props: Props) {
+  const cityId = props.cityId ?? null;
+
+  const [hotelId, setHotelId] = React.useState(props.hotelId ?? '');
+  const [paymentLink, setPaymentLink] = React.useState(null);
+  const [currentSearchStats, setCurrentSearchStats] = React.useState(
+    defaultState.currentSearchStats,
+  );
+
+  const dateFactoryState = useApi(dateFactory, {
```
RobinCsl: I don't know if `dateFactoryState` really helps me to understand better what's going on.
Perhaps, extracting `checkin, setCheckin, checkout, setCheckout` from `useApi(...)` would make it slightly more understandable that's it's a black box spitting out this information. 
Again, this is just a suggestion; I don't know what would be a better practice.

app/hotels/src/__tests__/HotelsContext.test.js
```diff
@@ -16,177 +14,78 @@ describe('getAsUtcDate', () => {
...
-    });
-
-    it('gives error for checkin date before today', () => {
-      const wrapper = getInstance({
-        checkin: formatForMachine(DateUtils().addDays(-1)),
-        checkout: formatForMachine(DateUtils().addDays(1)),
-      });
-      expect(wrapper.state.errors).toEqual({ beforeToday: true });
-    });
+const getValidateProps = (cityId, checkin, checkout) => ({
```
RobinCsl:
```suggestion
const getValidateProps = ({cityId, checkin, checkout}) => ({
```
Maybe as an object, it would be clearer in the tests below what you're setting when calling `getValidateProps(-1)`?

app/hotels/src/allHotels/HotelsPaginationContainer.js
```diff
@@ -35,16 +38,13 @@ export function HotelsPaginationContainer(props: Props) {
...
     if (priceMax != null && priceMin != null) {
       setCurrentSearchStats({
-        priceMax,
-        priceMin,
+        priceMax: new Decimal(priceMax),
+        priceMin: new Decimal(priceMin),
       });
     }
-  }, [props.data, setCurrentSearchStats]);
+  }, [priceMax, priceMin, setCurrentSearchStats]);
```
RobinCsl: That looks a lot clearer, I like that!

packages/shared/src/useApi/useApi.js
```diff
@@ -0,0 +1,28 @@
+// @flow strict
```
RobinCsl: Could you add a test or some README that would give an example on how to use it, and why you'd want to use it?
I know the article is great, but a quick example on how to use it could be beneficial.

app/hotels/src/HotelsContext.js
```diff
@@ -85,22 +83,39 @@ type State = {|
   +checkout: Date | null,
   +roomsConfiguration: RoomConfigurationType | null,
   +currency: string,
-  currentSearchStats: CurrentSearchStats,
+  +currentSearchStats: {|
```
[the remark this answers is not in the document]
tbergq: Yes, look like 😊

app/hotels/src/HotelsContext.js
tbergq: > it's a black box spitting out this information

I see a function. Functions are either void or have a return type. We have the source code, and IMHO, it should be clear what is going on. 

I will however use destructuring here, since I did it on the line below anyway 🙂

app/hotels/src/__tests__/HotelsContext.test.js
tbergq: 👍

app/hotels/src/allHotels/HotelsPaginationContainer.js
tbergq: It is also a performance improvement, since it is no longer diffing the entire `props.data` 😄

tbergq:
@RobinCsl 🚀

PR #1632, opened 2019-06-12
Hotels: Add info text when dates autochange

### LINKED ISSUE #1627: Add info text when adjusting dates because user selected invalid dates
User selects checkout > checkin + 30
User selects checkin > checkout - 30
User selects checkin > checkout
User selects checkout < checkin

tbergq:
User might select a date out of scope, and the other date will change.
This message has been added to make the user aware of the change.

Add toast component in shared package
closes #1627


packages/shared/src/toast/Toast.js
```diff
@@ -0,0 +1,96 @@
...
+import { Animated, View } from 'react-native';
+
+import type { TranslationType } from '../../types/Translation';
+import AdaptableBadge from '../AdaptableBadge';
+import StyleSheet from '../PlatformStyleSheet';
+import type { StylePropType } from '../../types/Styles';
+
+type Props = {|
+  +text: TranslationType,
+  +duration: number,
```
jvaclavik: There should be `+duration?` when you use default props

tbergq: `+duration?` would indicate that `this.props.duration` might be undefined, which it never will be. In case it is not passed as a prop, it will use the default value.

PR #1645, opened 2019-06-21
Shared: Add swipe-away functionality to toast

tbergq:


packages/shared/src/toast/Toast.js
```diff
@@ -20,58 +25,85 @@ type State = {|
   +isVisible: boolean,
 |};
 
+type HandlerChangeEvent = {
+  +nativeEvent: {
+    +state: number,
+    ...
```
jvaclavik: This is cool! I didn't know that we can use it this way in Flow

tbergq: I see you are not up to date with the `@kiwicom/eslint-config` in margarita then.

It will give you an error if you try to use inexact object without `...`

packages/shared/src/toast/Toast.js
```diff
@@ -20,58 +25,85 @@ type State = {|
...
+        this.timeout = setTimeout(this.hide, this.props.duration);
       });
     });
   };
 
   hide = () => {
-    Animated.timing(this.opacityValue, {
-      toValue: 0,
+    Animated.timing(this.translateY, {
+      toValue: -250,
```
jvaclavik: Btw what if height of the toast will be bigger than 250px?

tbergq: Then I could always increase it. But it will be hidden when the animation ends anyways.

PR #1646, opened 2019-06-21
Hotels: Fix performance issue on list page

tbergq:
Rendering is slow because list view renders, and map view renders
even if only list view is visible. Changing behaviour to hide
map view, then show spinner until it is loaded entirely

### DIFF
--- a/app/hotels/src/allHotels/RenderSearchResults.js
+++ b/app/hotels/src/allHotels/RenderSearchResults.js
@@ -1,9 +1,15 @@
 // @flow strict
 
 import * as React from 'react';
-import { Animated, Dimensions, View } from 'react-native';
+import { Animated, Dimensions, View, InteractionManager } from 'react-native';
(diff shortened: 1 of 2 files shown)


app/hotels/src/allHotels/RenderSearchResults.js
```diff
@@ -1,9 +1,15 @@
 // @flow strict
 
 import * as React from 'react';
-import { Animated, Dimensions, View } from 'react-native';
+import { Animated, Dimensions, View, InteractionManager } from 'react-native';
 import { graphql, createFragmentContainer } from '@kiwicom/mobile-relay';
-import { StyleSheet, Device, type OnLayout } from '@kiwicom/mobile-shared';
+import {
+  StyleSheet,
+  Device,
```
jvaclavik: Why do you need `Device`? It looks like it's not used here

PR #1647, opened 2019-06-21
Shared: Fix android toast swipe

tbergq:
The toast became visible above the header.
Changed marginTop to paddingTop to prevent this behaviour.
React-native-gesture-handler was not set up correctly, fix that also(in dev)
Note it will not work in production until we fix it in RNHotelsActivity.kt
and deploy a new framework

### DIFF
--- a/app/hotels/src/navigation/allHotels/navigationOptions/NavigationOptions.js
+++ b/app/hotels/src/navigation/allHotels/navigationOptions/NavigationOptions.js
@@ -8,13 +8,19 @@ import HeaderLeft from './HeaderLeft';
 import HeaderRight from './HeaderRight';
 
 export default function getNavigationOptions() {
+  const statusBarHeight = Platform.select({
+    ios: 0,
+    android: StatusBar.currentHeight,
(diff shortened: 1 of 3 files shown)


app/hotels/src/navigation/allHotels/navigationOptions/NavigationOptions.js
```diff
@@ -10,11 +10,18 @@ import HeaderRight from './HeaderRight';
 export default function getNavigationOptions() {
   return {
     headerStyle: {
-      height: 78,
-      marginTop: Platform.select({
+      height:
+        78 +
+        Platform.select({
```
jvaclavik: This could be in some const, because it's repetitive

tbergq: It could, I can fix 😊

tbergq: @jvaclavik 🚀

PR #1651, opened 2019-06-25
Upgrade react-native to latest version

tbergq:


tbergq:
Just a reminder if you want to test it

https://github.com/kiwicom/mobile#upgrade-react-native

android/build.gradle
```diff
@@ -31,14 +31,12 @@ subprojects {
         }
 
         dependencies {
-            classpath "com.android.tools.build:gradle:3.4.0-alpha04"
+            classpath "com.android.tools.build:gradle:$gradlePluginVersion"
```
RobinCsl: I don't know how significant this is, but the `gradle-wrapper.jar` has the SHA256 checksum of the 2.4 version release (https://gradle.org/release-checksums/)

You might want to also upgrade it (`cd android; ./gradlew wrapper --gradle-version 5.4.1`) if you want the latest version, and you can check the checksum hash by using
```
shasum --algorithm=256 gradle/wrapper/gradle-wrapper.jar
```
and then check the hash with the ones available on https://gradle.org/release-checksums/

package.json
```diff
@@ -36,7 +36,7 @@
     "detox": "^12.11.1",
     "dotenv": "^7.0.0",
     "react": "16.8.3",
-    "react-native": "0.59.2",
+    "react-native": "0.59.9",
```
RobinCsl: Would you also like to update the other dependencies, like in https://react-native-community.github.io/upgrade-helper/?from=0.59.2&to=0.59.9?

android/dependencies.gradle
```diff
@@ -22,6 +22,7 @@ ext {
     playServicesMapsVersion = "16.1.0"
     playServicesMeasurementBaseVersion = "16.0.2"
     playServicesAnalyticsVersion = "16.0.4"
+    facebookVersion = "4.39.0"
```
RobinCsl: Why do we need this new dependency?

android/build.gradle
tbergq: Thanks, I was unaware of this. It added some extra changes to `gradlew` file, but I was able to run, so I guess it is fine 🤷‍♂

package.json
tbergq: They are already up to date. You can run `yarn upgrade-interactive --latest` and see that none of these dependencies show up. 

You can also verify by checking the yarn.lock file.

android/dependencies.gradle
tbergq: Logging module, we are logging stuff to facebook.

PR #1652, opened 2019-06-26
Upgrade dependencies

tbergq:


.flowconfig
```diff
@@ -99,5 +99,4 @@ unsafe-getters-setters
 ;untyped-type-import
 sketchy-null
 
-[version]
```
RobinCsl: You think this is useless? It's true that I always need to change this file when I upgrade flow... and the rest of the config is left untouched anyway.

packages/shared/index.js
```diff
@@ -110,7 +110,6 @@ export {
 /**
  * Non-UI components (components without render output):
  */
-export { default as AppStateChange } from './src/AppStateChange';
```
RobinCsl: That file would have been relevant in a full-on RN app, right?

.flowconfig
tbergq: Yes, I think it is, I noticed it is also removed from `incubator/universe`. 
Well actually the tag is there, but no version specified. Which is basically the same.

packages/shared/index.js
tbergq: It could be relevant in a brownfield app as well, but we created this long time ago, we had a use-case for it. We no longer use it, so we don't need dead code.

RobinCsl: Absolutely

PR #1653, opened 2019-06-26
Hotels: Show correct max number of guests

tbergq:
We showed maxPerson avaiable for room-type. Room type could be double room.
But there would be different block for 1 person living in it vs 2 persons.
This is the reason we had rooms that looked the exact same
with different prices.

### DIFF
--- a/app/hotels/src/singleHotel/roomList/BeddingInfo.js
+++ b/app/hotels/src/singleHotel/roomList/BeddingInfo.js
@@ -14,7 +14,7 @@ type Props = {|
 
 export class BeddingInfo extends React.Component<Props> {
   formatBeddingInfo = (): $ReadOnlyArray<React.Element<typeof Translation>> => {
-    const { room } = this.props;
+    const room = this.props.room?.room;
     const beddingOptions = room?.bedding ?? [];
     const beddingTranslation = [];
     beddingOptions.forEach((beddingOption, index) => {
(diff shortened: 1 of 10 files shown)


tbergq:
We need to wait for MR on BE to be merged before we can merge this.

tbergq:
> We need to wait for MR on BE to be merged before we can merge this.

This has now been merged on the server, and we are good to go.

app/hotels/src/singleHotel/roomList/BeddingInfo.js
```diff
@@ -73,11 +73,13 @@ export class BeddingInfo extends React.Component<Props> {
...
 export default createFragmentContainer(BeddingInfo, {
   room: graphql`
-    fragment BeddingInfo_room on HotelRoomInterface {
-      maxPersons
-      bedding {
-        type
-        amount
+    fragment BeddingInfo_room on HotelRoomAvailabilityInterface {
+      maxOccupancy
+      room {
```
RobinCsl: It's unfortunate `room` is an attribute of `room` object. It makes reasoning about what's happening slightly confusing. But then for `maxPersons` variable a few lines before, it makes sense. Could you rename the fragment variables to be something else? Though I'm not sure...

tbergq: The parent is `availableRoom` which in the end is a room. There has been some unfortunate naming from the start of this project. Should we follow logic from the provider, there should have been Room and Block. But it makes no sense to start doing this on the client now. 

Maybe our graphql is ready for some refactoring. However this is a major job. Not something to be done anytime soon I guess.

RobinCsl: I agree, will approve it as it is.

PR #1658, opened 2019-07-08
Feature/target version 11

tbergq:


tbergq:
I would also have liked to upgrade react-native to `0.60.0`, but `react-native-gesture-handler` is not ready for that yet, so it will have to be in `target-version-12`

PR #1661, opened 2019-07-11
Bundle index.android.bundle into /src/main/assets of rnkiwimobile

tbergq:
This is a quick fix to solve app crashing due to missing bundle.
The bundle is now probably add to the framework twice.
Someone with gradle/android knowledge should fix it in the gradle script

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "reactNativeApp",
+  "name": "reactnativeapp",
   "version": "0.0.1",
   "private": true,
   "workspaces": [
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "reactNativeApp",
+  "name": "reactnativeapp",
```
RobinCsl: Why do you change the name of the package?

tbergq: It was giving me an ugly yellow warning on my IDE 🙈

tbergq:
> Would it be a matter of fixing the gradle script instead so that the initial bundle is found correctly?

Yes, it would. I don't know why this broke, since we did not change anything. Though I have no idea how to fix this and make it work. 

If you know how to, please fix it 😄 

I was hoping to ask someone from android team to have a look, because they will have smaller app size if they complete this task for us.

PR #1667, opened 2019-08-02
Upgrade react-native to 0.60.4

tbergq:


android/app/build.gradle
```diff
@@ -73,7 +73,8 @@ import com.android.build.OutputFile
  */
 
 project.ext.react = [
-    entryFile: "index.js"
+    entryFile: "index.js",
+    enableHermes: true
```
RobinCsl: nice!

android/app/src/main/java/com/reactnativeapp/MainApplication.java
```diff
@@ -31,21 +32,22 @@ public boolean getUseDeveloperSupport() {
...
-                    new RNDeviceInfoPackage(),
-                    new RNCardManagerPackage(),
-                    new RNGestureHandlerPackage()
-            );
-        }
+    protected List<ReactPackage> getPackages() {
+        @SuppressWarnings("UnnecessaryLocalVariable")
+        List<ReactPackage> packages = new PackageList(this).getPackages();
+        // Packages that cannot be autolinked yet can be added manually here, for example:
+        packages.add(new RNCWebViewPackage());
```
RobinCsl: Did you try to remove some of them here? I would think that most of them can be autolinked.

RobinCsl: Except our custom ones, of course.

index.js
```diff
@@ -10,6 +10,8 @@ YellowBox.ignoreWarnings([
   // TODO this a warning from react-navigation, see: https://reactnavigation.org/docs/en/common-mistakes.html
   // This happens in our app because the package-stack architecture we have for native apps
   'You should only render one navigator',
+  'Warning: componentWillReceiveProps is deprecated',
```
RobinCsl: Is that necessary? Maybe getting a reminder could be useful in case we forget that these warnings were suppressed and some weird bug happens in the future. WDYT?

tbergq: We don’t have this in our codebase, this comes from 3rd party libraries, and is just noise 😊

package.json
```diff
@@ -19,7 +19,7 @@
     "graphql": "yarn babel-node ./scripts/getSchema.js",
     "ios": "yarn react-native run-ios",
     "lint": "yarn jest --config=.jest-eslint.config.js",
-    "postinstall": "node ./scripts/configureApplication.js",
+    "postinstall": "node ./scripts/configureApplication.js && jetify",
```
RobinCsl: I might be wrong, but I believe `yarn android` jetifies by default.

react-native.config.js
```diff
@@ -0,0 +1,31 @@
+// @flow
+
+const deactivatedAutolinkingModules = [
```
RobinCsl: Ah okay, that explains previous occurrence in `MainApplication.java`.

android/app/src/main/java/com/reactnativeapp/MainApplication.java
tbergq: I tried, and it did not work. Feel free to give it a try. Also, we cannot autolink for Android in rnkiwimobile framework, since we fetch dependencies from maven, and not node modules

react-native.config.js
RobinCsl: Could you remind me the reason why this is necessary?

package.json
RobinCsl:
```
$ yarn android
...
info Running jetifier to migrate libraries to AndroidX. You can disable it using "--no-jetifier" flag.
...
```

react-native.config.js
tbergq: If I don't do this, I get an error telling me to unlink the native modules, when I try to unlink the native modules, It says the module is not linked. When I remove it from build, and wait for it to be autolinked, it is not.

package.json
tbergq: I am not sure it is run when we upload the android framework. I got androidx build errors when trying to run the release script.

RobinCsl: Ah right... I forgot about this part...

PR #1670, opened 2019-08-06
Fix map animation

tbergq:
Hide map and show map after animation is done

### DIFF
--- a/app/hotels/src/allHotels/RenderSearchResults.js
+++ b/app/hotels/src/allHotels/RenderSearchResults.js
@@ -1,7 +1,7 @@
 // @flow strict
 
 import * as React from 'react';
-import { Animated, Dimensions, View, InteractionManager } from 'react-native';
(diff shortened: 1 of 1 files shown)


app/hotels/src/allHotels/RenderSearchResults.js
```diff
@@ -91,18 +93,14 @@ export const RenderSearchResults = (props: Props) => {
           duration: transitionDuration,
           useNativeDriver: true,
         }),
-      ]).start();
+      ]).start(() => {
+        setShouldRenderMap(true);
```
RobinCsl: You also want it to be true here? Is that on purpose?

tbergq: No, I was testing something, and I forgot to flip it back, thanks.

tbergq:
@RobinCsl 🚀

PR #1672, opened 2019-08-07
Android: compile hermes for staginRelease variant

tbergq:


android/rnkiwimobile/build.gradle
```diff
@@ -82,6 +82,7 @@ dependencies {
     if (enableHermes) {
       def hermesPath = "../../node_modules/hermesvm/android/";
       debugImplementation files(hermesPath + "hermes-debug.aar")
+      stagingReleaseImplementation files(hermesPath + "hermes-debug.aar")
```
RobinCsl: Could you explain your choice? Since it's "stagingRelease", I would have thought `hermes-release.aar` should be used.

tbergq: Yeah, this should be a release build I guess, so I will change it to release, thank you 😊

tbergq: @RobinCsl 🚀

PR #1677, opened 2019-08-16
playground-ios: Add account to playground app

tbergq:
Note that this is ok, but we should not yet tell native that
account is available.
If they try to access it in production it could crash.
We need to publish new code-push-target-version before it is safe to use
in production

### DIFF
--- a/ios/reactNativeApp.xcodeproj/project.pbxproj
+++ b/ios/reactNativeApp.xcodeproj/project.pbxproj
@@ -34,6 +34,8 @@
 		27E931E420ADA0630053AE19 /* main.m in Sources */ = {isa = PBXBuildFile; fileRef = 27E931E320ADA0630053AE19 /* main.m */; };
 		51A8F6BE673D4671A3545CD8 /* CircularPro-Medium.otf in Resources */ = {isa = PBXBuildFile; fileRef = 50EB547AC35C4D25B73F0376 /* CircularPro-Medium.otf */; };
 		60854A62F6F281394EE676B5 /* libPods-reactNativeApp.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 0520FD3BC77D396F74CF8285 /* libPods-reactNativeApp.a */; };
+		63497BE72306A93D0006DC8B /* Roboto.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 63497BE62306A93D0006DC8B /* Roboto.ttf */; };
+		63497BE82306A93D0006DC8B /* Roboto.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 63497BE62306A93D0006DC8B /* Roboto.ttf */; };
 		63738759220D81E600665942 /* JavaScriptCore.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 63738758220D81E600665942 /* JavaScriptCore.framework */; };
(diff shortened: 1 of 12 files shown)


tbergq:
Seems I have to wait for new version of `account-native` first ⌛️

ios/reactNativeApp.xcodeproj/project.pbxproj
```diff
@@ -468,6 +472,7 @@
 				898C440300784E69B91D0ED6 /* Roboto-Medium.ttf in Resources */,
 				21C5EFC22189F86200F3C098 /* package.json in Resources */,
 				FA705002A8FD4F5BBDB4B4F4 /* Roboto-Regular.ttf in Resources */,
```
RobinCsl: Should that be somehow removed?

RobinCsl: Actually, why is Roboto even linked for iOS? It's not used, right?

tbergq: I think at some historical point we probably linked it, and autolinking does not remove it. 

This does not `bubble` up to the kiwi.com app anyways.

tbergq: But sure, we could add an issue to remove it.

yarn.lock
```diff
@@ -1328,20 +1328,23 @@
...
+    "@kiwicom/js" "^0.13.0"
+    "@kiwicom/orbit-components" "^0.53.0"
     "@kiwicom/orbit-design-tokens" "^0.6.5"
     "@kiwicom/relay" "^3.2.0"
     buffer "^5.2.1"
     dataloader "^1.4.0"
     date-fns "^1.30.1"
     graphql "^14.4.2"
     graphql-iso-date "^3.6.1"
+    npm-registry-client "^8.6.0"
```
RobinCsl: Not related completely to this PR, but why is `npm-regristry-client` a dependency of this package?

yarn.lock
```diff
@@ -1328,20 +1328,23 @@
...
     react-native-safe-area-view "^0.14.6"
     rimraf "^2.6.3"
```
RobinCsl: Same question for `rimraf`?

yarn.lock
tbergq: I made a note of it, and I will improve it in the next release of account-native package.

yarn.lock
tbergq: Same as above

yarn.lock
```diff
@@ -1328,20 +1328,23 @@
...
+  integrity sha512-N03UKZR6+CLQTZX9s8/JE3Six1vtxqR07TtUO4wSRxrp/Ehq84sZ3te4FF2bEdglpJnmCXfgX+jhFZXy21uNoA==
   dependencies:
     "@kiwicom/graphql-global-id" "^0.10.0"
     "@kiwicom/graphql-utils" "^0.8.0"
+    "@kiwicom/js" "^0.13.0"
+    "@kiwicom/orbit-components" "^0.53.0"
     "@kiwicom/orbit-design-tokens" "^0.6.5"
     "@kiwicom/relay" "^3.2.0"
     buffer "^5.2.1"
     dataloader "^1.4.0"
```
RobinCsl: Why do you need `dataloader`? That should be useful for the GraphQL server part, not in the client, right?

tbergq: Account don't have a backend yet, so it uses an in-memory-grahpql-mock, it is needed for now

tbergq: Actually, I think it is in account just because someone copy-pasted it from another package. I will remove it for next release as well.

PR #1681, opened 2019-08-19
Account: Add paddingTop on android root view

tbergq:


tbergq:
@RobinCsl this one also please 🙏

PR #1683, opened 2019-08-20
Upgrade detox

tbergq:
Major upgrade, but seems the only BC is dropping support
for iOS 9.X simulators


tbergq:
Also note, fails to run on android, but what else is new 🤷‍♀

PR #1686, opened 2019-08-22
accounts: Prepare price alerts screen

tbergq:
Note: I want to prepare Also price alerts. This way, the next time we increase the framework version, it should be possible for native to access and test the accounts screens and we can keep pushing new code with code-push. 

We should release yet a new version before going to production, since there is a risk that the user will see the `under development screen` if he visits the screen before the app has downloaded the new bundle from code-push

### DIFF
--- a/android/rnkiwimobile/src/main/java/com/kiwi/rnkiwimobile/account/RNAccountSettingsActivity.kt
+++ b/android/rnkiwimobile/src/main/java/com/kiwi/rnkiwimobile/account/RNAccountSettingsActivity.kt
@@ -1,0 +1,9 @@
+package com.kiwi.rnkiwimobile.account
+
+import com.kiwi.rnkiwimobile.RNKiwiActivity
+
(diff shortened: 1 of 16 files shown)


android/rnkiwimobile/src/main/java/com/kiwi/rnkiwimobile/account/RNAccountSettingsActivity.kt
```diff
@@ -0,0 +1,9 @@
+package com.kiwi.rnkiwimobile.account
+
+import com.kiwi.rnkiwimobile.RNKiwiActivity
+
+abstract class RNAccountSettingsActivity(initialProperties: RNAccountInitialProperties) :RNKiwiActivity(RNAccountModule.getInitialProperties(initialProperties)) {
```
Blackhex: I don't know how much do you watch code formatting in this project but the semicolon between new class and base class is not formatted as in the rest of the project.

tbergq: Now that you mentioned it, enough to trigger my OCD  🙈 

Updated now 😊

PR #1689, opened 2019-08-23
Feature/target version 13

tbergq:


docs/common/changelog.md
```diff
@@ -1,6 +1,21 @@
 # Code-push version changelog
 
-## Target versoin 12.0.0
+## Target version 13.0.0
+
+- upgrade rnkiwimobile to version `0.0.49`
+
+### unreleased
```
RobinCsl: Does 
```
- Upgrade react-native to 0.60.5
- Upgrade native dependencies
```

come with upgrading rnkiwimobile to version `0.0.49`? Or it's released only through CodePush (though I doubt that since it's native deps?)?

tbergq: This is just referencing the maven aar framework version,it is released once the CI passes. To be fair, so are js changes because they are bundled in with the framework. We just have no version to tie them too before we run yarn code-push

RobinCsl:
Do you want to squash the commits, @tbergq?

tbergq:
Yes, I will handle the squash

PR #1690, opened 2019-08-23
Shared: Use orbit-react-native in CloseButton

tbergq:


packages/shared/src/buttons/CloseButton.js
```diff
@@ -13,7 +13,7 @@ type Props = {|
 
 export default function CloseButton({ onPress }: Props) {
   return (
-    <Button onPress={onPress} type="secondary">
```
RobinCsl: I don't like web APIs in React Native 😄 `onPress` feels more platform agnostic

yarn.lock
```diff
@@ -1518,6 +1518,13 @@
   dependencies:
     ramda "^0.26.1"
 
+"@kiwicom/orbit-react-native@^0.3.2":
+  version "0.3.2"
+  resolved "https://registry.yarnpkg.com/@kiwicom/orbit-react-native/-/orbit-react-native-0.3.2.tgz#4dd99ad7986c04b5ab31bda34ea6a51a871653a9"
+  integrity sha512-VFf+3KsS/FeQ2A71qhIQiEwtyem5g/O3fRRANYZADRrmqXr6x8RVekASZvnYWKHudc+w0OsuhW46d4kFC1dHsA==
+  dependencies:
+    "@kiwicom/orbit-design-tokens" "^0.6.6"
```
RobinCsl: No `styled-components`, halleluja!

packages/shared/src/buttons/CloseButton.js
tbergq: It feels weird to me to add onClick here but, seems to be requirement from orbit that the API is the same as for the web components 🙈

tbergq:
Let me have a second look at it. Definitely the buttons should act the same 🤔

tbergq:
The opacity is matching the buttons in the native app, so I would say it is an improvement, we should just need to use the orbit button everywhere 😄 

The activeOpacity however is coming way to late yes. I want to investigate it a bit.

tbergq:
Ok, definitely not ready yet, there is no ripple effect on android either. I will see if there is some other component that qualifies. 

Closing this MR.

PR #1694, opened 2019-08-29
rnkiwimobile: publish jsc-android to maven

shmoula:


tbergq:
@Blackhex please have a look

PR #1698, opened 2019-10-24
Dispatch gesture events on main queue

sjavora:
Fixes [crashes](https://jira.kiwi.com/browse/MOBILE-7970) when building app with Xcode 11.


tbergq:
Since this PR contains changes to native code, you would need to upgrade `code-push-target-binary-version` in `package.json` -> `14.0.0`, and upgrade version in `.build/package.json`.

PR #1705, opened 2019-11-18
Upgrade flow to 0.112.0

tbergq:
Move all of node_modules under declarations
Add @adeira/flow-bin

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -28,25 +28,9 @@ node_modules/warning/.*
 [include]
 
 [declarations]
-<PROJECT_ROOT>/node_modules/@kiwicom/account-native
+.+/node_modules/.+
 
 [untyped]
-; Be careful - do not add 'StyleSheet' here
-.*/node_modules/@react-native-community/cli/.*/.*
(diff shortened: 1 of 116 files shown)


RobinCsl:
> Move all of node_modules under declarations

 Considering the large number of FlowFixMe annotations added in the PR, what's the reasoning behind this? Have there been changes in Flow that motivate it?

> Add @adeira/flow-bin

What's the benefit of adding this as a dependency?

tbergq:
[declarations] up until recent version (0.111 or 0.110) was broken, so modules was added to untyped instead to silence warnings from within react-native itself. This has led to us also typing our code without types for these modules. Now that declarations is fixed, we should not make future errors here, but rather silence the ones we have and fix them in the future. Fixing these errors within one PR would make this PR way too big. 

`Add @adeira/flow-bin` is for the moment the same as `@kiwicom/flow-bin@0.6.0`, which is being used in `universe`, kiwicom/flow-bin is no longer maintained, while adeira/flow-bin is. 

It does some caching, making flow faster.

PR #1707, opened 2019-11-18
FlowFixMe clean up

tbergq:


app/hotels/src/allHotels/useHotelsFilter.js
```diff
@@ -11,9 +11,8 @@ export default function useHotelsFilter() {
   return React.useMemo(() => {
     const { hotelAmenities, maxPrice, minPrice, ...restFilterParams } = filterParams;
     return {
-      ...restFilterParams,
-      // $FlowFixMe Errors after moving rn modules from untyped to declarations
       ...(maxPrice != null ? { maximumPrice: maxPrice.toFixed(2) } : {}),
+      ...restFilterParams,
```
tbergq: Hmm, this might not work as expect 🤔

tbergq: Yes, it should, nvm

app/hotels/src/map/allHotels/HotelSwipeList.js
```diff
@@ -8,10 +8,10 @@ import {
   BottomSheet,
   Device,
   StyleSheet,
-  type OnLayout,
   BottomSheetHandle,
   Translation,
 } from '@kiwicom/mobile-shared';
+import type { LayoutEvent } from 'react-native/Libraries/Types/CoreEventTypes';
```
RobinCsl: Would you rather have a layer from a common package instead of importing directly from `react-native`?

tbergq: Probably a good idea, it will be easier to remember.

tbergq: @RobinCsl 🚀

PR #1709, opened 2019-11-19
Flow fix me cleanup 3

tbergq:


packages/relay/index.js
```diff
@@ -7,14 +7,15 @@ import {
...
+  type Environment,
 } from '@kiwicom/relay';
 import { Alert } from '@kiwicom/mobile-localization';
 
-import PublicEnvironment from './src/PublicEnvironment';
-import PrivateEnvironment from './src/PrivateEnvironment';
 import ConnectionManager from './src/ConnectionManager';
 import type { QueryRendererProps as _QueryRendererProps } from './types';
 
+export { default as PublicEnvironment } from './src/PublicEnvironment';
```
RobinCsl: So we don't really need the PrivateEnvironment?

tbergq: We don't need to export at this point at least

tbergq:
> Could you update

Yes, I will do it later today. I didn't even remember that we had these docs anymore 😂

tbergq:
@RobinCsl 🚀

PR #1710, opened 2019-11-19
accountDeleted method for iOS

sjavora:
Add RNKiwiAccountUpdater to expose `accountDeleted` method to JS. Add `didDeleteAccount` property to react (ha) to account deletion in iOS app.

### DIFF
--- a/.build/package.json
+++ b/.build/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@kiwicom/react-native-app-hotels",
-  "version": "0.0.51",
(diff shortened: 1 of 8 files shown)


tbergq:
I changed the base branch of this PR to `code-push-target-binary-version-16` so we can merge it together with android.

tbergq:
Code looks good to me, once pipeline passes, we can merge it.

tbergq:
Hmm, well, `yarn ios` crashes, but I think I can handle it ⌛️

PR #1713, opened 2019-11-20
Flow fix me cleanup 4

tbergq:


ios/Podfile.lock
```diff
@@ -248,4 +248,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 0d3a33be0c26b65700ef25b55314f395972202ba
 
-COCOAPODS: 1.7.5
```
RobinCsl: This does not seem like it should be part of this PR?

packages/shared/src/forms/TextInput.js
```diff
@@ -5,22 +5,22 @@ import { TextInput as OriginalTextInput, View } from 'react-native'; // eslint-d
...
-import { type StylePropType } from '../../types/Styles';
+import type { ViewStyleProp, TextStyleProp } from '../../types/Styles';
 import StyleSheet from '../PlatformStyleSheet';
 import Text from '../Text';
 
 type Props = {|
   +value?: string,
   +placeholder?: React.Element<typeof Translation>,
-  +placeholderStyle?: StylePropType,
+  +placeholderStyle?: $FlowFixMe, // TODO: TextInput styles should be available from next release
```
RobinCsl: 👍

ios/Podfile.lock
tbergq: I could make a separate PR for it, but it might as well be a part of this one

PR #1714, opened 2019-11-20
Upgrade detox

tbergq:
note: You may have to reinstall applesimutils
if applesimutils --list returns empty array, then you have to


tbergq:
@RobinCsl please check again.

> On another note, to make the tests more resilient, we should consider disabling YellowBox warnings in e2e tests.

Actually, I think the right approach is to run e2e-tests in production build. I will open a task for it.

RobinCsl:
> Actually, I think the right approach is to run e2e-tests in production build.

That's the best approach, yes 😊

PR #1716, opened 2019-11-20
Upgrade dependencies

tbergq:
replace @kiwicom/relay with @adeira/relay

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "ios": "yarn react-native run-ios",
     "lint": "yarn jest --config=.jest-eslint.config.js",
     "postinstall": "monorepo-babel-node ./scripts/configureApplication.js && jetify",
-    "relay": "kiwicom-relay-compiler",
+    "relay": "adeira-relay-compiler",
     "release-android": "yarn babel-node ./scripts/buildAndroidSnapshots.js",
     "release-ios": "yarn babel-node ./scripts/releaseFramework.js",
     "start": "yarn react-native start",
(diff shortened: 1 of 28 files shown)


scripts/configureApplication.js
```diff
@@ -66,25 +66,4 @@ if (!fs.existsSync(dotEnvPath)) {
   fs.writeFileSync(dotEnvPath, `SENTRY_DSN=${sentryDsn}`);
 }
 
-log('Patch node crawler');
-
-const files = globSync('/**/jest-haste-map/build/crawlers/node.js', {
```
tbergq: I think this should be fixed for all sub-dependencies as well now 🤞

packages/localization/package.json
```diff
@@ -8,11 +8,11 @@
     "@kiwicom/rnmodules": "^0",
     "date-fns": "^1.30.1",
     "form-data": "^2.5.1",
+    "@adeira/fetch": "^1.0.0",
```
RobinCsl: Do you want it as a dep and not as a peerDep like previously?

package.json
```diff
@@ -43,10 +43,10 @@
     "react": "16.8.6",
     "react-native": "0.60.5",
     "react-native-app-registry-components-to-constants": "0.0.1",
+    "@adeira/fetch": "^1.0.0",
```
RobinCsl: Same question here.

packages/localization/package.json
tbergq: Yes, it is a dependency, and should be just that instead of peerDependency

PR #1717, opened 2019-11-21
Add deleteAccount() callback.

Blackhex:
Not sure this is enough.


tbergq:
> Not sure this is enough

I will test it in the feature branch before merging into master. I will ping you if something is not working.

PR #1719, opened 2019-11-22
Upgrade @kiwicom/account and pass bridge function

tbergq:
Cleanup code on ios since it was broken
And move it to RNModules

### DIFF
--- a/ios/RNModules/RNAccountManager/RNAccountManager/RNAccountManager.h
+++ b/ios/RNModules/RNAccountManager/RNAccountManager/RNAccountManager.h
@@ -11,6 +11,6 @@
 
 RCT_EXTERN const NSNotificationName _Nonnull RNKiwiAccountDeletedNotification;
 
-@interface RNKiwiAccountUpdater : NSObject <RCTBridgeModule>
+@interface RNAccountManager : NSObject <RCTBridgeModule>
 
 @end
--- a/ios/RNModules/RNAccountManager/RNAccountManager/RNAccountManager.m
+++ b/ios/RNModules/RNAccountManager/RNAccountManager/RNAccountManager.m
@@ -6,11 +6,15 @@
 //  Copyright © 2019 Kiwi.com. All rights reserved.
 //
 
+#import "RNAccountManager.h"
+
 const NSNotificationName RNKiwiAccountDeletedNotification = @"RNKiwiAccountDeletedNotification";
 
-#import "RNKiwiAccountUpdater.h"
 
-@implementation RNKiwiAccountUpdater
+
+@implementation RNAccountManager
+
+RCT_EXPORT_MODULE();
 
 RCT_EXPORT_METHOD(accountDeleted) {
   [[NSNotificationCenter defaultCenter] postNotificationName:RNKiwiAccountDeletedNotification object:nil userInfo:@{}];
--- a/app/account/package.json
+++ b/app/account/package.json
@@ -6,7 +6,7 @@
   "dependencies": {
     "@kiwicom/mobile-shared": "*",
     "@kiwicom/mobile-localization": "*",
-    "@kiwicom/account-native": "0.0.10"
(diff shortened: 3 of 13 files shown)


tbergq:
@RobinCsl, if you want to test that the method is there, the easiest way is to just add `console.warn(NativeModules.RNAccountManager.accountDeleted);` to `WithAccountStack.js`

@sjavora , there were some syntax errors, so I had to fix them. I also moved to Bridge module to `RNModules` where the rest of the bridge modules lives. Please verify that I did not break anything.

tbergq:
also, @RobinCsl , I didn't test on android yet. I want to do that after this one has been merged.

ios/RNModules/RNAccountManager/RNAccountManager/RNAccountManager.m
```diff
@@ -6,12 +6,19 @@
...
 
-#import "RNKiwiAccountUpdater.h"
 
-@implementation RNKiwiAccountUpdater
 
+@implementation RNAccountManager
+
+RCT_EXPORT_MODULE();
+
++ (BOOL)requiresMainQueueSetup {
```
sjavora: On the native side, this is doing nothing here that requires the main queue - why was this added?

ios/RNModules/RNAccountManager/RNAccountManager/RNAccountManager.h
```diff
@@ -9,8 +9,9 @@
 @import Foundation;
 #import "React/RCTBridgeModule.h"
 
+@protocol RCTBridgeModule;
 RCT_EXTERN const NSNotificationName _Nonnull RNKiwiAccountDeletedNotification;
 
-@interface RNKiwiAccountUpdater : NSObject <RCTBridgeModule>
+@interface RNAccountManager : NSObject <RCTBridgeModule>
```
sjavora: I specifically named this class so that it wouldn't be a "manager". Why would you change this?

ios/RNModules/RNAccountManager/RNAccountManager/RNAccountManager.m
tbergq: I can try running without it, I really just saw that most of our modules have this

tbergq:
> What exactly was not building? I tried it before submitting and had no errors.

It was missing an `;`  in `RNKiwiViewController.m` line 240, and `NSNotificationName ` was placed before the import of the `.h` file, so it was not defined. 

And it did not build for me 😊

ios/RNModules/RNAccountManager/RNAccountManager/RNAccountManager.h
tbergq: So that we have the same name for android and ios

ios/RNModules/RNAccountManager/RNAccountManager/RNAccountManager.m
tbergq: Seems it was not necessary, so I removed it again. Thanks.

sjavora:
Oh yeah, I was building a different target - that explains why those build errors didin't show up /facepalm

ios/RNModules/RNAccountManager/RNAccountManager/RNAccountManager.h
```diff
@@ -9,8 +9,9 @@
 @import Foundation;
 #import "React/RCTBridgeModule.h"
 
+@protocol RCTBridgeModule;
```
sjavora: This shouldn't be necessary - it is defined in the import above

tbergq: Yes, looks like you are right, I will remove it

tbergq: @sjavora Removed

PR #1720, opened 2019-11-22
Playground add permissions to load bundle from bundler

tbergq:


android/playground/src/main/AndroidManifest.xml
```diff
@@ -12,7 +13,8 @@
         android:label="@string/app_name"
         android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"
-        android:theme="@style/AppTheme">
+        android:theme="@style/AppTheme" 
+        android:usesCleartextTraffic="true" >
```
RobinCsl: Since that's just the playground, this would never be the case in production, right?

tbergq: Right, it is only `rnkiwimobile` which is exposed to native application.

RobinCsl:
I can load the `Accout Setting` portion, but the `Hotels` and `Hotels Fragment` crash.

`Hotels` just loads the view, loads the hotel list, and once it's loaded, it crashes.

`Hotels Fragment` returns an error screen:
Screenshot_1575304411

I don't think it's totally related to this PR though. 🤔

tbergq:
I am getting 401's from our maven today 😞

tbergq:
Well, I don't have any of the issues you are describing, and I need this one to finish accounts, so I am going to merge it. 

If the problem you describe arises, we can either fix it, or revert this commit. Anyway, this is not affecting any production code.

PR #1723, opened 2019-11-25
Fix DeviceLocale for android

tbergq:
Android native module only return Locale like `en-GB`
So we have to parse this in JS for it to work as it does on iOS

### DIFF
--- a/packages/localization/src/GetDeviceLocale.android.js
+++ b/packages/localization/src/GetDeviceLocale.android.js
@@ -1,0 +1,20 @@
+// @flow strict
+
+import { NativeModules } from 'react-native';
+
+const locale = NativeModules.RNDeviceInfo.Locale; // e.g. en-US
+const [language, territory] = locale.split('-');
+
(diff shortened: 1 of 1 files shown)


tbergq:
This code lives in our repo, the android code is [here](https://github.com/kiwicom/mobile/blob/master/android/react-native-native-modules/src/main/java/com/skypicker/reactnative/nativemodules/device/RNDeviceInfoManager.kt). One could surely do the same in native code, but I am for one, more comfortable doing it javascript, and I would not have to increase bundle version yet again.

If you still feel there are strong advantages of doing this in native code, I would refactor it with next RN-upgrade instead 🙂

RobinCsl:
Nah, it's fine. I was just confused with the module being called RNDeviceInfo and googling this gave me `react-native-device-info` and there was no locale in it. Of course, it came from `NativeModules`, so I should have had a hint that this wasn't from an NPM package...

Looks fine to me!

PR #1727, opened 2019-12-03
Upgrade yarn to 1.21.0

tbergq:


.yarnrc
```diff
@@ -2,7 +2,7 @@
 # yarn lockfile v1
 
 
-lastUpdateCheck 1572004604492
+lastUpdateCheck 1575384140796
```
RobinCsl: Is there a command you execute to update that file?

tbergq: https://github.com/tbergq/monorepo/blob/master/apps/notes/yarn/yarn.md#yarn-policies-set-version

PR #1736, opened 2020-01-06
fix(localization): Fix typo in Hotels' translations

RobinCsl:
This commit also silences ESLint warnings in `e2e/` folder for the rule `adeira/valid-test-folder` as that folder only contains tests: it would be redundant to create a `__tests__/` folder for each subfolder in `e2e/`.


tbergq:
We could also have configured it, but I guess, yes, it is fine to deactivate it for this folder.

PR #1739, opened 2020-01-07
Upgrade flow-bin to 0.115.0

tbergq:
Delete test.sh script, move commands to package.json
Run tests with runInBand(-i) option -> Should execute faster


RobinCsl:
What do you think about creating a specific task for each test in `.gitlab-ci.yml`? That would make them run in parallel and would even be faster overall in CI.

tbergq:
I tested it once, and it actually took longer time. Because the install, upload cache, download cache just makes it all slower.

PR #1743, opened 2020-01-08
Move universal-datepicker to shared

tbergq:


packages/shared/src/forms/datePicker/universalDatepicker/DatePicker.ios.js
```diff
@@ -0,0 +1,115 @@
+// @flow
+
+import * as React from 'react';
+import { View, DatePickerIOS } from 'react-native';
```
RobinCsl, line 4: Consider adding the same TODO as for the Android date picker note to make sure we don't forget about this one too?

tbergquist-godaddy: Not sure it matters anymore, I think we are shutting down RN for good.

PR #1744, opened 2020-01-08
Move UniversalIcons to shared

tbergq:


packages/shared/src/icons/UniversalIcon.js
```diff
@@ -0,0 +1,55 @@
...
+import iconsMap from './icons.json';
+
+const sanitizeIconCharacter = (iconCharacter: string) => {
+  if (/^E(?:.{3})$/.test(iconCharacter)) {
+    return String.fromCharCode(parseInt(iconCharacter, 16));
+  }
+  return iconCharacter;
+};
+
+export default function Icon({ name, color = '#46515e', size = 'medium', style }: $FlowFixMe) {
```
RobinCsl, line 18: Consider changing the default color's value to match one of the design token colors?

